Issue #10, opened 2018-02-28
Questasim uvm_dpi.so


lfernand4:
HI,

I have build the TB and module with questasim 10.6c_3.  I start the simulation : make simc (without GUI)
![image](https://user-images.githubusercontent.com/36438648/36794481-c2f6b404-1ca0-11e8-9c56-ea86be71bbb5.png)

 I have the following warning : 
![image](https://user-images.githubusercontent.com/36438648/36794423-94b35dae-1ca0-11e8-90ce-217319f1c5ef.png)

Now I run (run -a) : 
![image](https://user-images.githubusercontent.com/36438648/36794540-ea8dbda0-1ca0-11e8-92f5-525d45bc6dad.png)

How can I solve this error ? where does it come from ?
I have only add -64 in the scipt.

Thanks in advance.

raulbehl:
Even I am seeing the same issue. I do see in the Makefile that the C-code is compiled using the vlog command only:
```
# build TBs
$(library)/.build-tb: $(dpi) $(tbs)
	# Compile top level with DPI headers
	vlog$(questa_version) -sv $(tbs) -work $(library) $(filter %.c %.cc, $(dpi)) -ccflags "-g -std=c++11 " -dpiheader tb/dpi/elfdpi.h
	touch $(library)/.build-tb
```

Questasim should have been able to load the compiled .so files automatically but for some reason it is not doing that. 
As a work around to get things working, I did the following:
- Compiled the tb/dpi/*.c files separately:
```
g++ -c -fPIC -m64 -std=c++0x -I$QUESTASIM_HOME/include -shared -o elf_dpi.o -c elfdpi.cc
g++ -shared -m64 -o elf_dpi.so elf_dpi.o
```
- Updated the simc target of the Makefile
```
vsim${questa_version} -64 -c -lib ${library} ${top_level}_optimized +max-cycles=$(max_cycles) +UVM_TESTNAME=${test_case} \
	 +BASEDIR=$(riscv-test-dir) $(uvm-flags) +ASMTEST=$(riscv-test) -coverage -classdebug -sv_lib ariane_core/ariane/tb/dpi/elf_dpi -do "do tb/wave/wave_core.do"
```
With the above hack I was able to get it working. The correct fix should enable questasim to included those libraries without doing a separate g++ compile.

lfernand4:
Thanks for explanations. Now I want to run dhrystone, do you know how to do it ? 
[...]

PR #14, opened 2018-03-04
Fix for #10

raulbehl:
- Added build-dpi target in Make to compile DPI C-code using gcc
  - Updated the vsim commands in the Makefile to include the newly
    generated DPI library
  - Updated the vsim command with the "-64" switch as we are always
    using this switch with vlog


zarubaf:
Thanks for the contribution! Works smooth on my side.

PR #27, opened 2018-03-19
:bug: Fixes #26

### LINKED ISSUE #26: Flush multiplication unit
Currently the multiplication/division unit is not flushed when the controller changes control flow. This needs a fix asap.

raulbehl:
- Added flush_i port to mult/div module
    - Both result and valid are de-asserted when flush_i is high

### DIFF
--- a/src/mult.sv
+++ b/src/mult.sv
@@ -40,8 +40,8 @@ module mult (
     logic                     div_valid_op;
     logic                     mul_valid_op;
     // Input Arbitration
-    assign mul_valid_op = mult_valid_i && (operator_i inside { MUL, MULH, MULHU, MULHSU, MULW });
-    assign div_valid_op = mult_valid_i && (operator_i inside { DIV, DIVU, DIVW, DIVUW, REM, REMU, REMW, REMUW });
+    assign mul_valid_op = ~flush_i && mult_valid_i && (operator_i inside { MUL, MULH, MULHU, MULHSU, MULW });
+    assign div_valid_op = ~flush_i && mult_valid_i && (operator_i inside { DIV, DIVU, DIVW, DIVUW, REM, REMU, REMW, REMUW });
 
     // ---------------------
     // Output Arbitration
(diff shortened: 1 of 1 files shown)


zarubaf:
Hi @raulbehl,

sorry for the long delay. Unfortunately I think we need some more logic to fully mitigate this potential bug:

The problem is that the controller can decide that he wants to flush (say if the committing instruction is writing a CSR with side effects e.g.: address translation) and so we need to kill all in-flight instructions. The problem are not the instruction which are valid in that cycle where commit flushes but potentially long lasting instructions (like divisions) which can commit tens of cycles later. For the cycle where `valid_o` is high the scoreboard will take care of not saving it (flushing) it. But what the divider unit needs to do is:

1. Clear the pipelined valid signal of the multiplication unit.
2. Reset the divider FSM to idle

If you want to PR those changes I'd be happy to merge, but if you don't want to I can also take care of it. For the moment I will open a request for change.

raulbehl:
Hi Florian, 

Apologies for the delayed response. Sure. I’ll try to make the requested changes. Thank you for the review

zarubaf:
Hey @raulbehl,

sorry for the long radio silence (I have been on holidays the last week). Unfortunately we are not quite there. I've requested another round of changes.

Thanks for your patience. Let me know if something is not clear.

Cheers,
Florian

raulbehl:
Hi Florian, 

I think I am probably missing the exact issue here. 
Here is my understanding:
In the last commit, I am driving the `mul_valid_op` to 1'b0 which would ultimately drive the `mult_valid_q` signal to 1'b0 if flush is asserted. I think this should have the same effect as changing the `mult_valid_q` flop itself with the `flush_i` condition? 

For the divider unit, I had updated the serial_divider module FSM to go to IDLE if we see flush_i asserted in any of the FSM states. 

Please let me know if there is something I am missing.

zarubaf:
Hey @raulbehl,

have been looking at it again and you are absolutely right. I think what confused me was the explicit looking at the flush signal in each state. I will slighlty refactor and merge.

Thanks!

raulbehl:
Thank you for the merge :)

PR #29, opened 2018-03-21
Add placeholders for PMP CSRs for memory protection.

jrrk:
Make ISA test path relative to $(RISCV). Add placeholder CSRs to prevent divergence when ISA tests initialise PMPs. Function needs fleshing out later. Path to fesvr/htif.h needs adding to verilator flags.


jrrk:
Some changes backed out to avoid travis conflicts. Shell script added to simulate what Travis does.

zarubaf:
Thanks for the changes!

PR #30, opened 2018-03-22
WIP: Improve Verilator version of isa test to redirect commit log to a different file for each test

jrrk:
The upstream version of the Ariane Verilator support always outputs the commit log to trace_core_00_0.dasm. This is unsuitable for the run-asm-tests-verilator target because the file gets overwritten every time. This patch redirects the output stream in the Makefile before Verilator is launched. It also upgrades the output produced to a convenient machine readable format for subsequent processing in a formal model. The cycle count is reduced to 32 bits for more convenient viewing in a terminal.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -145,7 +145,7 @@ run-asm-tests: build
 		$(library).$(test_top_level)_optimized;)
 
 run-asm-tests-verilator: verilate
-	$(foreach test, $(riscv-tests), obj_dir/Variane_wrapped $(riscv-test-dir)/$(test);)
+	$(foreach test, $(riscv-tests), obj_dir/Variane_wrapped $(riscv-test-dir)/$(test) > $(riscv-test-dir)/$(test).log ;)
 
 run-failed-tests: build
 	# make the tests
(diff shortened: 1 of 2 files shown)


zarubaf:
Can you elaborate a bit on the formal model? Is it publicly available? Currently I see a little bit of a drawback that we can't disassemble the instruction stream anymore with `spike-dasm` (the instruction word needs to be wrapped in `DASM()` although I completely agree with you regarding the filename and the much needed additional information. I think what I currently would need is a little bit more information on the formal tool to make this useful for everybody.

Thanks,

Florian

jrrk:
The formal tool is available here:

https://github.com/jrrk/l3riscv

This is a fork of a model written by SRI International's Computer 
Science Laboratory. The L3 itself was written at Cambridge with a 
permissive license.

It needs to be integrated into the regression suite. The stream is 
filtered by a C program so it's no problem to add DASM() at some point

I see two usage scenarios, post processing the log file and checking the 
execution live using DPI. In this mode it can check for the processor 
hanging.

In view of the above you might want to defer merging until it all works 
together.

PR #51, opened 2018-07-03
Fix FP forwarding and missing FP loads/stores

### LINKED ISSUE #48: Forwarding of FP registers
There is a potential problem with the forwarding logic of floating point registers.

https://github.com/pulp-platform/ariane/blob/8620f41ea1db93c7883844011092147a23cabab3/src/scoreboard.sv#L217-L220

In the excerpt above, it is the destination register of `mem_q[i]` that is being forwarded as source register of `issue_instr_o`. Because of this, I think the correct check should be `is_rd_fpr(mem_q[i].sbe.op) == is_rs1_fpr(issue_instr_o.op)` instead. There are more cases of this throughout the scoreboard.

---

Also, in `is_rd_fpr` (and `is_rs2_fpr`), shouldn't the test below go up to `[FLD:FLB]` instead of stopping at `FLW`?

https://github.com/pulp-platform/ariane/blob/8620f41ea1db93c7883844011092147a23cabab3/include/ariane_pkg.sv#L253-L256

stmach:
Fixes #48

### DIFF
--- a/include/ariane_pkg.sv
+++ b/include/ariane_pkg.sv
@@ -227,7 +227,7 @@ package ariane_pkg;
     function automatic logic is_rs2_fpr (input fu_op op);
         if (FP_PRESENT) begin // makes function static for non-fp case
             unique case (op) inside
-                [FSD:FSW],                       // FP Stores
+                [FSD:FSB],                       // FP Stores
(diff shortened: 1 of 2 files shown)


zarubaf:
LGTM. Thanks

PR #58, opened 2018-07-14
WIP: Fix #54

### LINKED ISSUE #54: Potential L1 dcache issues
I have a memory test program that goes wrong as soon as the range is increased to 64K bytes (this is in a large behavioural memory). I have racked my brains and cannot think of a reason unless it is to do with L1-cache misbehaviour.

```
cd tmp/riscv-tests
git remote remove origin
git remote add origin https://github.com/jrrk/riscv-tests.git
git fetch
git checkout -b ariane_memtester
cd build
make
cd ../../..

```
```
xxx@xxx:/local/scratch/jrrk2/ariane-isatest$ obj_dir/Variane_wrapped tmp/riscv-tests/build/benchmarks/memtester.riscv

Bare metal DRAM test
memtester version 4.3.0 (64-bit)
Copyright (C) 2001-2012 Charles Cazabon.
Licensed under the GNU General Public License version 2 (only).

test range is 0x800134f0 to 0x800234ef
Loop 1/1:
ok         3 3 2 2 1 1 0 0
  Random Value: FAILURE: 0x7ffecaaf != 0xfff99a4c at offset 0x60.
p1 = 0x80013550, p2 = 0x8001b550
xxx@xxx:/local/scratch/jrrk2/ariane-isatest$ git status
On branch master
nothing to commit, working directory clean
xxx@xxx:/local/scratch/jrrk2/ariane-isatest$
```

(This output takes about 10 minutes to appear on my workstation)

I hope you can reproduce the above. If you succeed it in fixing it some strange things will happen perhaps when byte writing is tested because your ram model does not support this mode.

suehtamacv:
I discussed with Zaruba the reasons behind #54. 

The bug was triggered by a store followed by a load that triggered the eviction of the corresponding line (without saving its contents to memory).  To solve this, the store FSM checks whether its index matches the load index. If they match, the store is treated as if it has missed on the cache. 

I am unsure of the performance implications, or if the regression tests still pass.

### DIFF
--- a/src/cache_ctrl.sv
+++ b/src/cache_ctrl.sv
@@ -69,7 +69,8 @@ module cache_ctrl #(
         input  logic [63:0]                                        bypass_data_i,
         // check MSHR for aliasing
         output logic [55:0]                                        mshr_addr_o,
-        input  logic                                               mshr_addr_matches_i
+        input  logic                                               mshr_addr_matches_i,
+        input  logic                                               mshr_index_matches_i
 );
 
     // 0 IDLE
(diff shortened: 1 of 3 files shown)


zarubaf:
Thanks that seems to pass regressions.

PR #73, opened 2018-08-13
Small Bugfixes

Moschn:
Some small problems I came across while trying to synthesize ariane on an FPGA. 

I currently do not have access to a Qustasim/Modelsim installation and cannot fully test the patches.


src/store_buffer.sv
```diff
@@ -81,11 +81,11 @@ module store_buffer (
     // Speculative Queue - Core Interface
     // ----------------------------------------
     always_comb begin : core_if
-        automatic logic [DEPTH_SPEC:0] speculative_status_cnt;
+        automatic logic [$clog2(DEPTH_SPEC):0] speculative_status_cnt;
         speculative_status_cnt = speculative_status_cnt_q;
 
         // we are ready if the speculative and the commit queue have a space left
-        ready_o = (speculative_status_cnt_q < (DEPTH_SPEC - 1)) || commit_i;
+        ready_o = (int'(speculative_status_cnt_q) < (DEPTH_SPEC - 1)) || commit_i;
```
zarubaf: We should probably make the localparam the right size instead of casting every use.

src/store_buffer.sv
```diff
@@ -145,10 +145,10 @@ module store_buffer (
     assign data_we_o  = 1'b1; // we will always write in the store queue
 
     always_comb begin : store_if
-        automatic logic [DEPTH_COMMIT:0] commit_status_cnt;
+        automatic logic [$clog2(DEPTH_COMMIT):0] commit_status_cnt;
         commit_status_cnt = commit_status_cnt_q;
 
-        commit_ready_o = (commit_status_cnt_q < DEPTH_COMMIT);
+        commit_ready_o = (int'(commit_status_cnt_q) < DEPTH_COMMIT);
```
zarubaf: Here as well.

Moschn:
I changed the localparams to the correct size

PR #75, opened 2018-08-17
cache hierarchy changes and LSU load path optimizations

msfschaffner:
the uvm-components submodule needs to be fixed to point to 6bb81367a35bde8952217685dba9457183bc63ea
in order to account for the interface changes (see separate pull request)

### DIFF
--- a/Bender.yml
+++ b/Bender.yml
@@ -13,7 +13,7 @@ dependencies:
 
 sources:
   - include/ariane_pkg.sv
-  - include/nbdcache_pkg.sv
+  - include/std_cache_pkg.sv
   - target: not(synthesis)
     files:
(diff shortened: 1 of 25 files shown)


msfschaffner:
sorry, forgot to adapt the ariane wrapper for verilator (the uvm-components submodule already pointed to the correct commit). its passing now.

zarubaf:
I've overlooked a change in:

https://github.com/pulp-platform/ariane/blob/d9f37012a4dd782cc6efd922b37bc18b3c06e074/Makefile#L15

We need to remove the version string for our external users. We can't impose the strange ETH versioning we have to the rest of the world.

zarubaf:
Thanks!

PR #86, opened 2018-08-24
Ariane next

msfschaffner:
remove flat byte enables
switch icache to inferrable blockrams
exchange sram with regfile
switched sram to sram_wrapper in testharness
replace dirty/valid sram with regfile
replace behav_sram with fpga inferrable ram
remove flat byte enables
fix in makefile
add reset to valid regs


zarubaf:
LGTM in principle and thanks for the long needed port to common cells... Just one question: What are you planning to achieve with the wrapped SRAMs? Have you talked to Fabian how we are doing it in Kosmodrom? That will likely break with it. Maybe I don't see the the benefit of your approach at the moment :(

msfschaffner:
hm... let me have a chat with him. the idea here was to get rid of the different behavioural descriptions (`xilinx_sram` and `behav_sram`), and provide one common wrapper that - per default - instantiates a behavioral model that is inferable by different FPGA toolchains (for rams with byte-enable's the coding style is different for Altera and Xilinx). the sram wrapper could in principle also be adapted to include SRAM macros for ASIC flows....

PR #90, opened 2018-08-30
Small changes into the Ariane

lvcargnini:
I spread some modules from the frontend into multiple files, just for the organization of the design, instead of a single file containing multiple modules I place each module in its file.

Also, I added four modules, one it is a generic that IC flows and FPGA flows translate into an asynchronous DFF. Also added a register model based on it. Added two generic SRAM modules, that on FPGA will be translated into its counterpart on BRAM automatically.  Needs to be improved to support both IC and FPGA flows, for it works on FPGA,  on ASIC it creates a sea of DFFs.


zarubaf:
Thanks.

I am okay with pulling out the modules as separate files. At a certain point in the future we might need to add some sub-folders. Furthermore I think we need to adapt the `src_files.yml` and the `Bender.yml` to also list the new sources. (If you want you can make a new PR with only those changes + listing them in the two files).

For the FPGA part I am adding @msfschaffner as a reviewer he actually having similar efforts right now and we do already have BRAM genrating SRAMs in the ariane_next branch (might want to take a look at the `fpga_support` subrepo). Lets try to reduce duplicate files and efforts as much as possible.

msfschaffner:
Concerning the DFF issue... which tool version are you using? When I synthesize the design with a simple clock constraint for a VU9P and Vivado 2018.1, I get the following summary in the place utilization report:

```
1.1 Summary of Registers by Type
--------------------------------

+-------+--------------+-------------+--------------+
| Total | Clock Enable | Synchronous | Asynchronous |
+-------+--------------+-------------+--------------+
| 51    |            _ |           - |            - |
| 0     |            _ |           - |          Set |
| 0     |            _ |           - |        Reset |
| 0     |            _ |         Set |            - |
| 0     |            _ |       Reset |            - |
| 0     |          Yes |           - |            - |
| 9     |          Yes |           - |          Set |
| 20200 |          Yes |           - |        Reset |
| 0     |          Yes |         Set |            - |
| 18    |          Yes |       Reset |            - |
+-------+--------------+-------------+--------------+
```

and under `9. Primitives`, the tool reports that these have been indeed mapped to `FDCE` register primitives. Hence I do not think that explicit wrapping of registers is needed. 

I suggest you wait for the next commit on ariane_next from my side, and then try again. There will be some general fixes related to FPGA synthesis that might fix some of your problems.

lvcargnini:
OK,

- I'll look into the other yml files and add the changes there
- add the copyright headers into additional files
- and look into the new branch ariane_next

@msfschaffner  regarding the register/dff wrapping I'm working in a few pipeline changes, and from past experiences on IC flow or FPGA flow, I prefer to be on the safe side, so I'm using such description on my side project. It doesn't need to be added on the official Ariane I agree on that.

Please, @zarubaf, deny this PR, I'll clone a fresh repo and roll over the changes, then I'll make the changes in ariane_next branch and then a new PR.

Now one question where do I learn how you guys are using this yml files ? which tools and so on?

zarubaf:
Thanks. There is some slim usage how-to on the other projects we have (`pulpino`, `pulpissimo`, ...). But basically its nothing fancy: All files which are needed to run simulation and synthesis should be listed in those files.

PR #95, opened 2018-09-10
Modifications for Ariane Next

lvcargnini:
I moved the modules from frontend.sv into their own files, also added the files into src_files.yml.


zarubaf:
Thanks

PR #96, opened 2018-09-12
Ariane next

msfschaffner:


zarubaf:
It seems that there is not enough time to build verilator as well...

zarubaf:
As you seem to have re-enabled the torture test, can you maybe add a paragraph to README.md how to run it? There is already a section allocated. Would be super awesome. Thanks

PR #112, opened 2018-10-13
FPU Integration

zarubaf:


zarubaf:
Thanks! Oh yes, absolutely. Thanks for the pointer!

msfschaffner:
Thanks @asb ! Note that we plan to release a SV version at some point (probably within the next 2 months).

Issue #119, opened 2018-10-16
[Enhancement] Optimize LSU return path into the scoreboard


msfschaffner:
The return path of the LSU and the LSU arbiter are on a critical path. Provisioning separate write ports into the scoreboard would allow to remove the LSU arbiter. This should be feasible with moderate overhead if we share the writeback ports between ALU and Mult/Div (see  #118).

Issue #118, opened 2018-10-16
[Enhancement] Remove separate scoreboard write port for Mult/Div


msfschaffner:
Share the Mult/Div scoreboard writeback port with the ALU port. This frees up an additional writeback port for the load unit.

PR #130, opened 2018-10-17
Re-structure write back path into scoreboard

zarubaf:
This PR addresses issues #118 and #119. Synthesis needs to be checked separately.


zarubaf:
Don't squash and merge on this one as it might be necessary to revert a single commit.

PR #133, opened 2018-11-08
First attempt to bring debug_rom inside Ariane, to start to make it independent of user's own memory map

jrrk:
Hi Florian,
  This pull request is incomplete and just for review. I have merged the executable code from dm_mem into a dummy icache line. If you think a tidied up version of this is a goer, I can proceed and merge the data operations from dm_mem into a dummy dcache line.

### DIFF
--- a/tb/dpi/remote_bitbang.cc
+++ b/tb/dpi/remote_bitbang.cc
@@ -16,7 +16,7 @@
 
 /////////// remote_bitbang_t
 
-remote_bitbang_t::remote_bitbang_t(uint16_t port) :
+remote_bitbang_t::remote_bitbang_t() :
   socket_fd(0),
   client_fd(0),
   recv_start(0),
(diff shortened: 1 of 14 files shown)


tb/dpi/remote_bitbang.cc
```diff
@@ -43,7 +43,7 @@ remote_bitbang_t::remote_bitbang_t(uint16_t port) :
   memset(&addr, 0, sizeof(addr));
   addr.sin_family = AF_INET;
   addr.sin_addr.s_addr = INADDR_ANY;
-  addr.sin_port = htons(port);
+  addr.sin_port = htons(2222);
```
msfschaffner: We should keep this parameterizable...

jrrk: By default the port is random, this is probably not intentional, no doubt there is a simpler fix.

PR #139, opened 2018-11-23
Minimal set of changes to add vcsrun target

jrrk:
Minimal set of changes to add vcsrun target (using Synopsys vcs), whilst still allowing verilator to work (needs review for Questa compatibility)

Some simulation features are difficult to make compatible with vcs, so conditionally disabled for now. Current version comes up in openocd and recognises debug behaviour correctly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,7 +65,7 @@ src :=  $(filter-out src/ariane_regfile.sv, $(wildcard src/*.sv))      \
         $(wildcard src/cache_subsystem/*.sv)                           \
         $(wildcard bootrom/*.sv)                                       \
         $(wildcard src/clint/*.sv)                                     \
-        $(wildcard src/axi_node/src/*.sv)                              \
+        $(filter-out src/axi_node/src/axi_node_wrap_with_slices.sv, $(wildcard src/axi_node/src/*.sv))      \
         $(wildcard src/axi_mem_if/src/*.sv)                            \
         $(filter-out src/debug/dm_pkg.sv, $(wildcard src/debug/*.sv))  \
         $(wildcard src/debug/debug_rom/*.sv)                           \
@@ -73,10 +73,7 @@ src :=  $(filter-out src/ariane_regfile.sv, $(wildcard src/*.sv))      \
         src/fpu/src/fpnew_top.vhd                                      \
         src/common_cells/src/deprecated/generic_fifo.sv                \
         src/common_cells/src/deprecated/pulp_sync.sv                   \
-        src/common_cells/src/deprecated/find_first_one.sv              \
         src/common_cells/src/rstgen_bypass.sv                          \
-        src/axi/src/axi_cut.sv                                         \
-        src/axi/src/axi_join.sv                                        \
         src/fpga-support/rtl/SyncSpRamBeNx64.sv                        \
         src/common_cells/src/sync.sv                                   \
         src/common_cells/src/cdc_2phase.sv                             \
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -65,18 +65,15 @@ src :=  $(filter-out src/ariane_regfile.sv, $(wildcard src/*.sv))      \
         $(wildcard src/cache_subsystem/*.sv)                           \
         $(wildcard bootrom/*.sv)                                       \
         $(wildcard src/clint/*.sv)                                     \
-        $(wildcard src/axi_node/src/*.sv)                              \
+        $(filter-out src/axi_node/src/axi_node_wrap_with_slices.sv, $(wildcard src/axi_node/src/*.sv))      \
```
zarubaf: What is the problem with the AXI slices?

Makefile
```diff
@@ -305,6 +302,32 @@ run-torture-verilator: verilate
...
+                    $(filter-out %.vhd, $(ariane_pkg))                                     \
+                    $(filter-out src/fpu_wrap.sv, $(filter-out %.vhd, $(src)))             \
+                    +define+$(defines)                                                     \
+                    src/util/sram.sv                                                       \
+                    +incdir+src/axi_node                                                   \
+                    -CFLAGS "-std=c++11 -I../tb/dpi"                                       \
+                    $(list_incdir)                                                         \
+                    tb/ariane_tb.sv tb/dpi/SimJTAG.cc tb/dpi/remote_bitbang.cc	           \
+
+#		    /opt/Xilinx/Vivado/2018.1/data/verilog/src/glbl.v                      \
```
zarubaf: Thanks, that looks really good. Note that I would suggest that you base your stuff on the `ariane_next` branch. We already merged some fpga changes in there and you are likely better off using this branch (as we are going to merge all the fpga stuff in the upcoming week).

tb/ariane_tb.sv
```diff
@@ -41,6 +43,9 @@ module ariane_tb;
 
     // Clock process
     initial begin
+`ifdef VCS
+        $vcdpluson;
```
zarubaf: What is the purpose of `$vcdpluson`?

tb/common/SimDTM.sv
```diff
@@ -1,6 +1,7 @@
 // See LICENSE.SiFive for license details.
 //VCS coverage exclude_file
 
+`ifndef VCS
```
zarubaf: Can we add a TODO here so that we switch this on later again? I don't see a reason why it should not work on VCS

zarubaf:
Thanks for the PR, VCS is a target which was very long on the list!

Makefile
jrrk: These modules are not needed by default. However vcs treats them as top-level modules, but is too dumb to allow top-level modules with unconnected interfaces. Solution: move to an optional list when needed, automatically put -v flag (as needed) in front of them.

tb/ariane_tb.sv
jrrk: This turns on top-level waveform tracing in Synopsys proprietary format. This may be converted later to compacted gtkwave format using:

vpd2vcd vcdplus.vpd | vcd2fst - test.fst

vpd2vcd is part of Synopsys vcs suite, vcd2fst is part of gtkwave install.

zarubaf: Is there a command-line argument to turn this one? I would prefer to move the vendor specific stuff to the Makefile. If not then this hack is fine for me

tb/common/SimDTM.sv
jrrk: This particular syntax is not a problem. It is the implicit introduction of the dependence on the htif library which is the tricky issue (perhaps to do with the thread model). Here be Dragons.

tb/ariane_tb.sv
jrrk: I haven't found one that does that exact thing. The -debug_pp does this but also enables the rather clunky GUI at the same time.

msfschaffner:
Hi Jonathan,

Indeed, thanks for coming up with a patch for this. 

However, as Florian mentioned, it would be good if you could direct future patches towards `ariane_next`, since our main development is carried out on that branch.

We are currently in the process of pulling several dev-branches (FPGA, OpenPiton...) into `ariane_next` and will likely upstream that to master in the following days. May I ask you to wait until the end of the week and rebase your changes against our upcoming release? That would simplify things quite a bit for us.

jrrk:
No problem, I understand the reason for the delay.

silenceluo:
Thanks for the great work!
Anyone knows how to add 'target-options' into VCS command to execute binary files? Thanks!

jrrk:
The vcs -R command executes the resulting binary immediately. But there 
are reasons against doing this by default,

for example you might want to capture output from compilation and 
execution in separate files, or run in background

for convenient use of the -gui option. Personally I prefer to debug with 
the waveform capture/postprocess options rather than using the gui.

PR #140, opened 2018-11-23
Add new ZSBL

Moschn:
This PR adds the new zero stage bootloader to the fpga bootrom.


fpga/src/bootrom/linker.lds
```diff
@@ -0,0 +1,38 @@
+ENTRY(main)
+
+SECTIONS
+{
+    ROM_BASE = 0x83000000; /* ... but actually position independent */
```
zarubaf: I think this `ROM_BASE` will make problems if the boot-vector actually point to `0x10000`. Right?

Moschn: Ups! Fixed

PR #150, opened 2018-11-28
Update the SD card howto

Moschn:
Use up-to-date make targets on the ariane-sdk


Moschn:
Maybe we should add a link to this file from the README. The README needs to be updated anyway to reflect the new additions

zarubaf:
Good idea

PR #156, opened 2018-12-08
:bug: Fix HINT instructions in compressed decoder

zarubaf:


src/compressed_decoder.sv
```diff
@@ -130,8 +128,6 @@ module compressed_decoder
                                 // 00: c.srli -> srli rd, rd, shamt
                                 // 01: c.srai -> srai rd, rd, shamt
                                 instr_o = {1'b0, instr_i[10], 4'b0, instr_i[12], instr_i[6:2], 2'b01, instr_i[9:7], 3'b101, 2'b01, instr_i[9:7], riscv::OpcodeOpImm};
-                                // shamt field must be non-zero
-                                if ({instr_i[12], instr_i[6:2]} == 6'b0) illegal_instr_o = 1'b1;
```
msfschaffner: I think bit 12 still must be zero according to the spec (even for HINT instructions).

src/compressed_decoder.sv
```diff
@@ -201,8 +197,6 @@ module compressed_decoder
                     riscv::OpcodeC2Slli: begin
                         // c.slli -> slli rd, rd, shamt
                         instr_o = {6'b0, instr_i[12], instr_i[6:2], instr_i[11:7], 3'b001, instr_i[11:7], riscv::OpcodeOpImm};
-                        if (instr_i[11:7] == 5'b0)  illegal_instr_o = 1'b1; // register not x0
-                        if ({instr_i[12], instr_i[6:2]}  == 6'b0)  illegal_instr_o = 1'b1; // shift amount must be non zero
```
msfschaffner: again bit 12 must be zero

src/compressed_decoder.sv
```diff
@@ -231,8 +225,6 @@ module compressed_decoder
                             if (instr_i[6:2] == 5'b0) begin
                                 // c.jr -> jalr x0, rd/rs1, 0
                                 instr_o = {12'b0, instr_i[11:7], 3'b0, 5'b0, riscv::OpcodeJalr};
-                                // rs1 != 0
-                                illegal_instr_o = (instr_i[11:7] != '0) ? 1'b0 : 1'b1;
```
msfschaffner: isn't this RESERVED instead of HINT?

src/compressed_decoder.sv
zarubaf: https://github.com/riscv/riscv-opcodes/blob/1dfd1df36278f89aa3896410dc50b5604087eebd/opcodes-rvc#L17-L18

src/compressed_decoder.sv
zarubaf: https://github.com/riscv/riscv-opcodes/blob/1dfd1df36278f89aa3896410dc50b5604087eebd/opcodes-rvc#L31

src/compressed_decoder.sv
zarubaf: https://github.com/riscv/riscv-opcodes/blob/1dfd1df36278f89aa3896410dc50b5604087eebd/opcodes-rvc#L35

msfschaffner:
I found these mismatches when I cross-checked with the tables under 16.8 in the [most recent spec](https://github.com/riscv/riscv-isa-manual/releases/download/draft-20181227-c6741cb/riscv-spec.pdf) on pages 110-111.

But could also be that I overlooked something (so auto-generating would certainly help here :D). Let's discuss F2F.

src/compressed_decoder.sv
asb: I think the spec should be taken as the canonical source rather than riscv-opcodes. c.jr with rs1==0 is indeed RESERVED rather than a HINT. It's legal for an implementation to handle this as a NOP, but treating it as an illegal instruction is probably more future-proof.

src/compressed_decoder.sv
zarubaf: Decoded as Hint according to the spec.

src/compressed_decoder.sv
zarubaf: Thanks

zarubaf:
Thanks, I've addressed the issues. I'd appreciate if you could take another look.

PR #161, opened 2019-01-14
This patch makes the dm relocatable to an arbitrary base address.

msfschaffner:
The last 12bits need to be zero however.

See also https://github.com/msfschaffner/riscv-fesvr/commit/c3a6e8b67c7d6e9c0246186f91c30e5a6f921592 for the required mods in FESVR.

The changes in FESVR and the DM are both backwards compatible in case the DM address base is set to 0x0.


zarubaf:
Why did we opt for implementing the debug base register in the Core CSRs? I honestly doubt that we will ever get this through in the official spec.

PR #167, opened 2019-01-25
Switch JTAG from PMOD to FTDI chip, update OpenPiton file list and OpenOCD cfg files.

msfschaffner:


zarubaf:
That is awesome. Thanks. Could you just add a CHANGELOG entry mentioning this awesome fix? Thanks

PR #175, opened 2019-02-06
Correct mdio_oe naming, streamline to allow 1GHz capability

jrrk:
The correction to the 1000BaseT mode is ready for review and merge. The one remaining issue is why we do not see interrupts. A very similar arrangement in LowRISC with Rocket does show the expected interrupts. I am seeing a slow down with all the new kernel options being enabled. This needs reviewing as well to see if they are all necessary.


jrrk:
Perhaps I missed something, but the point is that the DDR interface to the Ethernet PHY consists of fixed functions built into the pads, so adding constraints does not alter the behaviour (apart from generating a number of timing errors). Obviously if constraints were needed for an ASIC, that would be a different matter.

jrrk:
The Travis check failed but this would appear to be unrelated to the change made. Seems to be a timeout or similar in resolving a library to be downloaded.

jrrk:
@zarubaf ping are you ready to merge?

zarubaf:
Yes thanks. Was waiting for the Travis Check yesterday and forgot about it :-)

PR #177, opened 2019-02-08
Increase DDR Memory Size to 1 GiB

jrrk:
It is a curious fact that although the GenesysII Ariane board is fitted with a gigabyte of memory, the RTL prevents any more than 128M of that from being used. Since the IOMMU used 64M by default, that leaves very little for tmpfs and kernel. This patch addresses the issue in Verilog, and attempts to fix the device tree blob, but for some reason the bootrom still reports 128M. If BBL is locally patched with a local DTB, then Linux sees and can use the whole gigabyte. This quirk of behaviour needs reviewing. This patch also corrects the local MAC address to a valid Digilent value, but the driver does not make use of this entry at the present time. The proper way to obtain the MAC address is from the protected area of the SPI flash, as explained rather briefly in the Digilent documentation.


zarubaf:
Hm actually your changes should work as they are. Are you sure that you synthesized the latest version of the bootrom?

janhoogerbrugge:
Is the source code of the bootrom missing in this commit. I only see generated code.

Where is the bootrom reporting the memory size?

jrrk:
The bootrom source has not changed as such, only the device tree source, and image/Verilog file outputs. These are not checked, just built as needed. The bootrom does not report memory size, since it requires a device tree blob parser. BBL can report this number if you want it as follows:

`diff --git a/machine/fdt.c b/machine/fdt.c
index e8a504f..550b1c2 100644
--- a/machine/fdt.c
+++ b/machine/fdt.c
@@ -196,6 +196,7 @@ void query_mem(uintptr_t fdt)
 
   mem_size = 0;
   fdt_scan(fdt, &cb);
+  printm("Memory size reported by query_mem = %ld MBytes\n", mem_size >> 20);
   assert (mem_size > 0);
 }
 
`

zarubaf:
Let's merge this as it does not look wrong. We'll find the reporting issue. Thanks Jonathan.

PR #187, opened 2019-02-19
Add Exclusive Adapter

zarubaf:
The submodules are currently in the process of being upstreamed but that should be a working foundation.


zarubaf:
Fixes #184

PR #190, opened 2019-02-21
Refactor serpent AXI adapter, bump dbg and atomics submodules, add separate bootrom for linux on OpenPiton

msfschaffner:


zarubaf:
Looks good to me.

PR #193, opened 2019-03-14
Clean-up naming to distinguish OP from GP Ariane

zarubaf:
Mostly re-names Piton to Write-through caches.


.gitlab-ci.yml
```diff
@@ -29,7 +29,7 @@ variables:
 stages:
   - build
   - standard
-  - serpent
+  - wb
```
msfschaffner: shouldn't this be `wt`? I would just name these targets something like `write-through` and `write-back` to make them more readable...

.gitlab-ci.yml
```diff
@@ -149,96 +149,96 @@ serdiv-quest:
...
+    - make -j${NUM_JOBS} run-benchmarks-verilator defines=WT_DCACHE
   dependencies:
     - build
 
 s-icache-quest:
-  stage: serpent
+  stage: wb
   script:
-    - cd tb/tb_serpent_icache/
+    - cd tb/tb_wb_icache/
```
msfschaffner: should be `tb_wt_icache`, no?

msfschaffner: or we can just leave the wb/wt identifier away in case of the icache, since wt and wb are irrelevant

.gitlab-ci.yml
zarubaf: Absolutely, thanks for that.

.gitlab-ci.yml
zarubaf: Yeah, I was thinking about similar things but then decided to go for `wt_icache` but the `wt` flavor of icache is used in the `wt` cache subsystem. In a next step we could align just to use one icache e.g.: the `wt` but I'd prefer to defer that decision a bit.

msfschaffner: Ok SGTM.

PR #201, opened 2019-03-18
Ariane 4.1.1 release

msfschaffner:
This is a hotfix for the OpenPiton cache system.


msfschaffner:
I think it just points to the sdk release overview, and I would leave it that way. As you say, we would have to bump it everytime we do a new release otherwise.

PR #206, opened 2019-03-26
Fix wt_axi_adapter bug that only appeared when using dcache lines that are wider than icache lines.

msfschaffner:


zarubaf:
Please target towards `ariane-dev` and add a CHANGELOG entry. Thanks

msfschaffner:
Yeah my editorconfig was wrong...

PR #213, opened 2019-04-07
Re-work `irq_subsystem` and `debug_req` to make PLIC behavior more stable

zarubaf:
Patches to make the interrupt subsystem work properly.


zarubaf:
Its still WIP, I'll ping you when the review is really due. I just want to get the CI checks to run.

PR #215, opened 2019-04-09
rv_plic: Add lowrisc PLIC

zarubaf:
Use a PLIC which has been developed as part of the lowrisc project. It
has been integrated into the Ariane SoC as a submodule pointing to a
fork which has some (temporary) custom patches on top.


zarubaf:
I've factored out the PLIC which is orthogonal to the rest of the changes coming.

zarubaf:
@msfschaffner Can we bring in this PR?

PR #221, opened 2019-04-22
Execute PMA

zarubaf:
Check for execute permission and do not blindly issue the read to the SoC. This should fix the hang-ups we've experienced in OpenPiton before.

### DIFF
--- a/include/ariane_pkg.sv
+++ b/include/ariane_pkg.sv
@@ -34,20 +34,20 @@ package ariane_pkg;
     typedef struct packed {
       int                     NrNonIdempotentRules;  // Number of non idempotent rules
       logic [15:0][63:0]      NonIdempotentAddrBase; // base which needs to match
-      logic [15:0][63:0]      NonIdempotentAddrMaks; // bit mask which bits to consider when matching the rule
+      logic [15:0][63:0]      NonIdempotentLength; // bit mask which bits to consider when matching the rule
       int                     NrExecuteRegionRules;  // Number of regions which have execute property
       logic [15:0][63:0]      ExecuteRegionAddrBase; // base which needs to match
-      logic [15:0][63:0]      ExecuteRegionAddrMaks; // bit mask which bits to consider when matching the rule
+      logic [15:0][63:0]      ExecuteRegionLength; // bit mask which bits to consider when matching the rule
     } ariane_cfg_t;
 
     localparam ariane_cfg_t ArianeDefaultConfig = '{
       NrNonIdempotentRules: 2,
       NonIdempotentAddrBase: {64'b0, 64'b0},
-      NonIdempotentAddrMaks: {64'b0, 64'b0},
+      NonIdempotentLength:   {64'b0, 64'b0},
       NrExecuteRegionRules: 3,
-      //                      DRAM,               Boot ROM,        Debug Module
-      ExecuteRegionAddrBase: {64'h8000_0000,      64'h1_0000,      64'h0},
-      ExecuteRegionAddrMaks: {mask(64'h40000000), mask(64'h10000), mask(64'h1000)}
+      //                      DRAM,          Boot ROM,   Debug Module
+      ExecuteRegionAddrBase: {64'h8000_0000, 64'h1_0000, 64'h0},
+      ExecuteRegionLength:   {64'h40000000,  64'h10000,  64'h1000}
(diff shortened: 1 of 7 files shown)


include/ariane_pkg.sv
```diff
@@ -33,25 +33,44 @@ package ariane_pkg;
...
+      int                     NrExecuteRegionRules;  // Number of regions which have execute property
+      logic [15:0][63:0]      ExecuteRegionAddrBase; // base which needs to match
+      logic [15:0][63:0]      ExecuteRegionAddrMaks; // bit mask which bits to consider when matching the rule
     } ariane_cfg_t;
 
     localparam ariane_cfg_t ArianeDefaultConfig = '{
       NrNonIdempotentRules: 2,
       NonIdempotentAddrBase: {64'b0, 64'b0},
-      NonIdempotentAddrMaks: {64'b0, 64'b0}
+      NonIdempotentAddrMaks: {64'b0, 64'b0},
```
msfschaffner: Typo `Maks` -> `Mask`

include/ariane_pkg.sv
```diff
@@ -33,25 +33,44 @@ package ariane_pkg;
...
       // pragma translate_off
       `ifndef VERILATOR
         assert(Cfg.NrNonIdempotentRules <= 16);
+        assert(Cfg.NrExecuteRegionRules <= 16);
       `endif
       // pragma translate_on
     endfunction
 
+    // Generate a mask for a given power of two length
+    function logic [63:0] mask (input logic [63:0] len);
```
msfschaffner: This mask function essentially just checks whether len is a power of two - otherwise the mask is equal to the input. Is this intended? How about a Base + Length encoding/terminology instead of Base + Mask?

src/mmu.sv
```diff
@@ -235,8 +237,21 @@ module mmu #(
...
     end
 
+    always_comb begin : execute_addr_check
+        // if we don't specify any region we assume everything is accessible
+        match_any_execute_region = (Cfg.NrExecuteRegionRules == 0);
+        // check for execute flag on memory
+        for (int i = 0; i < Cfg.NrExecuteRegionRules; i++) begin
+            match_any_execute_region |= (Cfg.ExecuteRegionAddrBase[i] & Cfg.ExecuteRegionAddrMaks[i])
+                                    == (icache_areq_o.fetch_paddr & Cfg.ExecuteRegionAddrMaks[i]);
+        end
```
msfschaffner: Are you sure this masking is behaving correctly? I would rather go for a base + len comparison here (as we do with the cacheable range in the L1 cache)...

zarubaf: The thinking was that in terms of logic you want to keep this check simple. The only restriction you will have with this `base + mask` feature is that the regions must be a power of two, I think this is a reasonable assumption.

include/ariane_pkg.sv
zarubaf: It returns something as well: `return {64{1'b1}} << $clog2(len);` This basically translates the `length` to a bitmask. What I can do is to ask for a length in the config struct and use this function to calculate the mask. Might be simpler?

msfschaffner: Yes, but the mask that you calculate is basically equal to `len` (with the added power-of-two check). You apply that mask to your address afterwards using a vector `&`, and I think this is not behaving as you intended...

E.g., suppose your `len` is `32`, then your mask will be `0x20`. Then you `&` this with your segment base and the address in question and compare the two values to see whether you hit the segment... I think this is only correct if you make sure that all MSBs are set to `1` as well.

src/mmu.sv
msfschaffner: Sure I am completely fine with this. I just think there is something wrong with the way you generate your mask...

include/ariane_pkg.sv
zarubaf: Well, you shift `len` zeros in a vector full of `1`. This ensures that all the MSBs are set.

include/ariane_pkg.sv
zarubaf: Thanks, changed to `Length` now

include/ariane_pkg.sv
msfschaffner: Ah sorry my bad. Sure this is correct.

PR #222, opened 2019-04-22
Fix Performance Counter encoding

zarubaf:
The performance counters were accessible in user mode although
`mcounteren` or `scounteren` are not implemented. This commit moves the
writeable performance counters to machine mode. Access from
user/supervisor-space will trap as this is not implemented in Ariane.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Changed
 
+- Fix privilege bug in performance counters. The counters have always been accessible in user mode.
 - Re-work interrupt and debug subsystem to associate requests during decode. This improves stability on for non-idempotent loads.
 - Fix RISC-V PK simulation bug caused due to insufficient time to init the `a0` and `a1` registers via the bootrom
 - Fix bug in wt_axi_adapter (only appeared when dcache lines were wider than icache lines)
--- a/include/riscv_pkg.sv
+++ b/include/riscv_pkg.sv
@@ -354,26 +354,55 @@ package riscv;
         CSR_MHARTID        = 12'hF14,
         CSR_MCYCLE         = 12'hB00,
         CSR_MINSTRET       = 12'hB02,
+        // Performance counters (Machine Mode)
+        CSR_ML1_ICACHE_MISS = 12'hB03,  // L1 Instr Cache Miss
+        CSR_ML1_DCACHE_MISS = 12'hB04,  // L1 Data Cache Miss
+        CSR_MITLB_MISS      = 12'hB05,  // ITLB Miss
+        CSR_MDTLB_MISS      = 12'hB06,  // DTLB Miss
+        CSR_MLOAD           = 12'hB07,  // Loads
+        CSR_MSTORE          = 12'hB08,  // Stores
+        CSR_MEXCEPTION      = 12'hB09,  // Taken exceptions
+        CSR_MEXCEPTION_RET  = 12'hB0A,  // Exception return
+        CSR_MBRANCH_JUMP    = 12'hB0B,  // Software change of PC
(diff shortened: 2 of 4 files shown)


src/issue_read_operands.sv
```diff
@@ -230,53 +230,65 @@ module issue_read_operands #(
...
-        lsu_valid_n    = 1'b0;
-        mult_valid_n   = 1'b0;
-        fpu_valid_n    = 1'b0;
-        fpu_fmt_n      = 2'b0;
-        fpu_rm_n       = 3'b0;
-        csr_valid_n    = 1'b0;
-        branch_valid_n = 1'b0;
+    // This needs to be like this to make verilator happy. I know its ugly.
+    always_ff @(posedge clk_i or negedge rst_ni) begin
+      if (~rst_ni) begin
```
msfschaffner: `~` -> `!`

src/issue_read_operands.sv
```diff
@@ -230,53 +230,65 @@ module issue_read_operands #(
...
+        mult_valid_q   <= 1'b0;
+        fpu_valid_q    <= 1'b0;
+        fpu_fmt_q      <= 2'b0;
+        fpu_rm_q       <= 3'b0;
+        csr_valid_q    <= 1'b0;
+        branch_valid_q <= 1'b0;
         // Exception pass through:
         // If an exception has occurred simply pass it through
         // we do not want to issue this instruction
         if (~issue_instr_i.ex.valid && issue_instr_valid_i && issue_ack_o) begin
```
msfschaffner: same here

src/issue_read_operands.sv
```diff
@@ -401,14 +413,6 @@ module issue_read_operands #(
             operand_a_q           <= '{default: 0};
             operand_b_q           <= '{default: 0};
             imm_q                 <= 64'b0;
```
msfschaffner: you could just use `'0` for multibit resets

src/load_store_unit.sv
```diff
@@ -15,7 +15,8 @@
 import ariane_pkg::*;
 
 module load_store_unit #(
-    parameter int unsigned ASID_WIDTH = 1
+    parameter int unsigned ASID_WIDTH = 1,
```
msfschaffner: Maybe we should start to clean up the parameter style (all caps -> pascal case)

src/mmu.sv
```diff
@@ -19,7 +19,8 @@ import ariane_pkg::*;
 module mmu #(
       parameter int unsigned INSTR_TLB_ENTRIES = 4,
       parameter int unsigned DATA_TLB_ENTRIES  = 4,
-      parameter int unsigned ASID_WIDTH        = 1
+      parameter int unsigned ASID_WIDTH        = 1,
+      parameter ariane_pkg::ariane_cfg_t Cfg = ariane_pkg::ArianeDefaultConfig
 )(
```
msfschaffner: some more nitpicking ;) `)(` -> `) (`

src/ariane.sv
```diff
@@ -234,7 +234,7 @@ module ariane #(
     .DmBaseAddress       ( DmBaseAddress )
   ) i_frontend (
     .flush_i             ( flush_ctrl_if                 ), // not entirely correct
```
msfschaffner: what is `not entirely correct` here?

src/issue_read_operands.sv
zarubaf: Thanks

src/ariane.sv
zarubaf: Haha, I have no idea, I think the naming. For sure not the logic. But I'd prefer not to touch it in this commit. Otherwise I am opening the box for merge conflicts again.

src/load_store_unit.sv
zarubaf: I think we should merge this into the new config struct. Possibly not with this commit though as it is unrelated to the cause: `Fix Performance Counter encoding`.

PR #228, opened 2019-04-26
Feature/clean prediction

zarubaf:
I've Linux successfully booting with the cleaned up version on the FPGA. So I don't see much reason to not merge it in this release window.


include/ariane_pkg.sv
```diff
@@ -281,7 +287,7 @@ package ariane_pkg;
     // ---------------
 
     // leave as is (fails with >8 entries and wider fetch width)
-    localparam int unsigned FETCH_FIFO_DEPTH  = 8;
+    localparam int unsigned FETCH_FIFO_DEPTH  = 4;
```
msfschaffner: Would it enhance performance to use 8 or not?

zarubaf: Internally it’s using 2 FIFOs of this size. So total size is 8.

zarubaf:
Can you let me know once you have confirmation of the OpenPiton build?

PR #238, opened 2019-05-13
 fp_dirty: Dirty on fp register write

### LINKED ISSUE #233: FPU single precision add inaccuracy
I am testing the improved, pipelined FPU under ariane-dev branch. I am trying to run TestFloat and in the main it works well, but I am having problems with certain particular operands, as per the attached testcase:

  TEST_FP_OP2_S_( 2, fadd.s,             33354290,             33357822,   -3531
.952392578125 ); #   actually returns  33357822
  TEST_FP_OP2_S_( 3, fadd.s, -1.701411733192644E+38, -1.701411733192644E+38, -2.
350970765024232E-38 ); #  actually returns NAN

[rv64uf-p-custom.zip](https://github.com/pulp-platform/ariane/files/3137711/rv64uf-p-custom.zip)

I can add more test vectors later if needed

zarubaf:
That should fix loads to floating point regs not dirtying the FP state (should fix #233). 

@jrrk I'd greatly appreciate if you could test the fix with your setup.

### DIFF
--- a/src/commit_stage.sv
+++ b/src/commit_stage.sv
@@ -77,7 +77,7 @@ module commit_stage #(
     always_comb begin : dirty_fp_state
       dirty_fp_state_o = 1'b0;
       for (int i = 0; i < NR_COMMIT_PORTS; i++) begin
-        dirty_fp_state_o |= commit_ack_o[i] & commit_instr_i[i].fu inside {FPU, FPU_VEC};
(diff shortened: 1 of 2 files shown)


jrrk:
I’m at the SiFive symposium today but I can certainly look at it tomorrow.

jrrk:
I notice you only test commit_instr_i[0].op. What if commit_instr_i[1].op is set, is that possible ?

zarubaf:
That is indeed stupid and a mistake. Instead of the `0` it should be `i`.

PR #240, opened 2019-05-14
mstatus SD bit incorrect

jrrk:
The mstatus SD bit doesn't reflect the FS dirty status as it should.

### DIFF
--- a/include/riscv_pkg.sv
+++ b/include/riscv_pkg.sv
@@ -40,7 +40,7 @@ package riscv;
     } xs_t;
 
     typedef struct packed {
-        logic         sd;     // signal dirty - read-only - hardwired zero
+        logic         sd;     // signal dirty - read-only - high if fs field is dirty
         logic [62:36] wpri4;  // writes preserved reads ignored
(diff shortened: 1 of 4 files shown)


zarubaf:
I am a bit suprised (or confused) that this solves the issue (and happy at the same time). Considering that `Dirty   = 2'b11` an and reduction should make sure that both are `1`, shouldn't it? Anyway, I am fine if it works but was just wondering if we can re-formulate everything to use comparisons instead of the and reductions?

zarubaf:
Thanks for doing all the debug work for me! I've proposed a fix which does not require reading from the `_d` which I don't like too much. The fix should have the same effect.

jrrk:
OK

jrrk:
Since you brought style as an issue, I don't think converting xxx.fs == riscv::Dirty to (&xxx.fs), is a good style, because the maintainer has no way of knowing (in the package) that an enumerated type has been converted to a boolean expression (in the module body). I don't know if it prevents one-hot encoding for one thing.

zarubaf:
I absolutely agree, thanks for pointing this out! I've pushed a fix!

PR #241, opened 2019-05-14
Set `sd` bit

zarubaf:
@jrrk May I propose the following fix?

### DIFF
--- a/include/riscv_pkg.sv
+++ b/include/riscv_pkg.sv
@@ -40,7 +40,7 @@ package riscv;
     } xs_t;
 
     typedef struct packed {
-        logic         sd;     // signal dirty - read-only - hardwired zero
+        logic         sd;     // signal dirty state - read-only
         logic [62:36] wpri4;  // writes preserved reads ignored
         xlen_t        sxl;    // variable supervisor mode xlen - hardwired to zero
         xlen_t        uxl;    // variable user mode xlen - hardwired to zero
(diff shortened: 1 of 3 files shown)


jrrk:
The comment on the sd field at line 43 of riscv_pkg.sv is now invalid (assuming it ever meant anything)

zarubaf:
Agree and changed!

jrrk:
I approve this change. There is still work to do on sub-normalised divides but this is repeatable and unrelated. For example 2.350988561514729E-38 / 2 => 0. (it should give 1.175494350822288E-38, a subnormal value)

zarubaf:
Thanks a lot for all the help! 
I think those are "known" bugs. @stmach can probably comment more on this.

stmach:
Hello, thanks for your inputs.
Indeed we are aware of bugs in the FP divider and square-root unit, however this one I haven't seen yet (see end of post for more thoughts on the bug).
The div/sqrt unit we currently have is not very configurable and the code is quite convoluted, so we are hoping to replace it with something more flexible (and IEEE compliant) soon. Hence a bugfix for the current div/sqrt is not a high priority at this time.

*Long-winded addendum*
The result value you list is correct (under default rounding), however it's actually encoded as the smallest normal FP32 value `0x00800000`. The infinitely precise result lies exactly between this result and the largest subnormal FP32 value.
The div/sqrt bugs we've seen so far concern off-by-one-ulp because of wrong rounding. What's possibly happening here seems to be that the mantissa is correctly rounded from all-ones (from the infinitely precise, subnormal result) up to all-zeroes, however the overflow produced from rounding doesn't seem to be added onto the (subnormal = 0) exponent. This leads to the zero result.
I don't know whether this bug is contained to this subnormal/normal border case, but it might be possible that mantissa overflows from rounding in general are buggy.

PR #243, opened 2019-05-15
Remove interface name aliasing to keep VCS happy

jrrk:
At the moment in an alias for Slave and out an alias for Master. This causes VCS to complain and refuse to compile when the names are mixed. This PR remove the in and out aliases. Requires [this PR](https://github.com/pulp-platform/axi/pull/9)

### DIFF
--- a/fpga/src/ariane_peripherals_xilinx.sv
+++ b/fpga/src/ariane_peripherals_xilinx.sv
@@ -23,17 +23,17 @@ module ariane_peripherals #(
     input  logic       clk_i           , // Clock
     input  logic       clk_200MHz_i    ,
     input  logic       rst_ni          , // Asynchronous reset active low
-    AXI_BUS.in         plic            ,
-    AXI_BUS.in         uart            ,
-    AXI_BUS.in         spi             ,
-    AXI_BUS.in         gpio            ,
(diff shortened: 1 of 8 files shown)


zarubaf:
Okay I am okay with this. Lets wait for the depending PR before we proceed.

zarubaf:
@jrrk The PR is merged, can you please update here so that CI passes and we can merge?

jrrk:
It seems like quite a lot has changed in the meantime, I have rebased manually as best I can, but which may or may not be correct. Perhaps somebody who knows GitHub better than me can cherry-pick the stuff that matters, if necessary.

jrrk:
I've completed everything I can think of, the Travis is still failing due to a timeout. Perhaps this should be manually reviewed to approve for merge because I see no reason why it would affect simulation performance.

msfschaffner:
Hi Jonathan, you might want to pull in #248 . We split up the Travis targets and disabled Torture for the time being to work around the timeout issue. In the long run we should rework the tests, e.g. bundle some of the redundant compilation stages, and switch to a different loading mechanism than via the FESVR (this is unfortunately quite wasteful in terms of simulation cycles).

zarubaf:
Closing in favor of #254

PR #245, opened 2019-05-21
Improve timing of scoreboard and FPU

msfschaffner:


msfschaffner:
Note: this also (slightly) improves the performance since some forwarding paths were not correctly activated before.

PR #248, opened 2019-05-27
Split into smaller travis tests

zachzzc:


msfschaffner:
Can you try to increase the timeout for the tests that fail?

https://docs.travis-ci.com/user/common-build-problems/

zachzzc:
It seems like the 25min wait time is not still enough. I will further increase it to 60min

msfschaffner:
So for some reason torture gets stuck with the writeback cache. This cache has a couple of issues anyway, and we currently do not consider it as `production ready`. Hence I would say we disable the torture test for the writeback cache, and only run it on the write-through cache (there it actually completes after 10mins or so).

zachzzc:
Torture test is unstable for both WT and WB cache. I don’t known if we want to disable both of them.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

msfschaffner:
For the sake of productivity, let's disable them for the time being. I've opened an issue to track this, since we should track down what the problems are.

msfschaffner:
Thanks!

PR #253, opened 2019-05-30
Generalize the realign module to enable arbitrary fetch width

zachzzc:
Test the code with 32 bit and 64 bit fetch width.

### DIFF
--- a/include/ariane_pkg.sv
+++ b/include/ariane_pkg.sv
@@ -275,9 +275,10 @@ package ariane_pkg;
     // Fetch Stage
     // ---------------
 
-    // leave as is (fails with >8 entries and wider fetch width)
     localparam int unsigned FETCH_FIFO_DEPTH  = 4;
     localparam int unsigned FETCH_WIDTH       = 32;
+    // the number of bits that the address of fetch block should align to
+    localparam NR_ALIGN_BITS = $clog2(FETCH_WIDTH/8);
     // maximum instructions we can fetch on one request (we support compressed instructions)
     localparam int unsigned INSTR_PER_FETCH = FETCH_WIDTH / 16;
 
--- a/src/frontend/frontend.sv
+++ b/src/frontend/frontend.sv
@@ -67,9 +67,9 @@ module frontend #(
     logic [63:0]   replay_addr;
 
     // shift amount
-    logic [$clog2(ariane_pkg::INSTR_PER_FETCH)-1:0] shamt;
     // address will always be 16 bit aligned, make this explicit here
-    assign shamt = icache_dreq_i.vaddr[$clog2(ariane_pkg::INSTR_PER_FETCH):1];
+    logic shamt;
+    assign shamt = icache_dreq_i.vaddr[1];
 
     // -----------------------
     // Ctrl Flow Speculation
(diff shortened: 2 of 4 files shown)


include/ariane_pkg.sv
```diff
@@ -278,6 +278,8 @@ package ariane_pkg;
     // leave as is (fails with >8 entries and wider fetch width)
```
msfschaffner: @zarubaf is this still the case? or can we remove this comment?

src/instr_realign.sv
```diff
@@ -50,290 +61,102 @@ module instr_realign (
...
-                        instr_o[2] = data_i[63:48];
-                        valid_o[2] = valid_i;
-                    end else begin
-                        unaligned_d = 1'b1;
-                        unaligned_instr_d = data_i[63:48];
-                        unaligned_address_d = addr_o[2];
-                    end
-                end
+        // iteratively realign the rest instructions
+        for (int unsigned i = 0; i < INSTR_PER_FETCH; i++) begin : gen_instr_realign
```
msfschaffner: Can you run an FPGA synthesis of the old variant and the new version and compare the timing reports? I would like to make sure that this does not impact timing.

zarubaf:
Awesome, thanks a lot! I agree with Michael on running a synthesis regression just to be sure. I can also try to boot Linux with the changes to see whether everything is working out.

src/frontend/frontend.sv
```diff
@@ -67,9 +67,9 @@ module frontend #(
     logic [63:0]   replay_addr;
 
     // shift amount
-    logic [$clog2(ariane_pkg::INSTR_PER_FETCH)-1:0] shamt;
-    // address will always be 16 bit aligned, make this explicit here
-    assign shamt = icache_dreq_i.vaddr[$clog2(ariane_pkg::INSTR_PER_FETCH):1];
+    // // address will always be 16 bit aligned, make this explicit here
```
zarubaf: Please remove one of the `//`

include/ariane_pkg.sv
zarubaf: No, that mystery has been fixed. @zachzzc It would be nice if you could delete this comment :)

zachzzc:
Here is the comparison between the old version and the new one. 
image

Basically the 32 bit fetch does not impact the timing but the 64 fetch width will increase the critical path. I am not sure if the increment is tolerable here.

src/frontend/instr_queue.sv
```diff
@@ -177,8 +177,8 @@ module instr_queue (
   // shift the inputs
   for (genvar i = 0; i < ariane_pkg::INSTR_PER_FETCH; i++) begin : gen_fifo_input_select
     /* verilator lint_off WIDTH */
-    assign instr_data_in[i].instr = instr[i + idx_is_q];
-    assign instr_data_in[i].cf = cf[i + idx_is_q];
+    assign instr_data_in[i].instr = instr[i + INSTR_PER_FETCH - idx_is_q];
```
zarubaf: They need to be pre-fixed with `ariane_pkg` e.g. `ariane_pkg::INSTR_PER_FETCH`

src/frontend/instr_queue.sv
```diff
@@ -177,8 +177,8 @@ module instr_queue (
...
+    assign instr_data_in[i].cf = cf[i + INSTR_PER_FETCH - idx_is_q];
```
zarubaf: Same here.

PR #262, opened 2019-06-18
Add support for Xilinx Kintex-7 FPGA KC705 Evaluation Kit

Steinegger:
This PR adds support for the [Xilinx Kintex-7 FPGA KC705 Evaluation Kit](https://www.xilinx.com/products/boards-and-kits/ek-k7-kc705-g.html).

The main changes are the additional constraints file, the configuration for the KC705's DDR3 memory and the differently sized interface with a reduced number of LEDs and switches to accommodate the KC705's peripheral.

The fpga/src/kc705.svh and fpga/src/genesysii.svh have an additional define KINTEX7 to be used for parts of the code used for both the Genesys II and the KC705 and other boards of the family. This define is then also used in the ariane-ethernet repo (separate PR).

The fpga/xilinx/xlnx_mig_7_ddr3/tcl/run.tcl is using the short version of the board name to determine the config file used for the DDR memory, allowing for more boards to be added in the future using the same method.


Steinegger:
Pull request for the ariane-ethernet use of the KINTEX7 define can be found here: lowRISC/ariane-ethernet#2.

Synthesis won't complete without this define as the KC705 requires the same RAMB16 define as the Genesys 2 board.

Makefile
```diff
@@ -497,7 +497,8 @@ fpga: $(ariane_pkg) $(util) $(src) $(fpga_src) $(uart_src)
 	@echo read_verilog -sv {$(filter-out $(fpga_filter), $(src))} 	   >> fpga/scripts/add_sources.tcl
 	@echo read_verilog -sv {$(fpga_src)}   >> fpga/scripts/add_sources.tcl
 	@echo "[FPGA] Generate Bitstream"
-	cd fpga && make BOARD="genesys2" XILINX_PART="xc7k325tffg900-2" XILINX_BOARD="digilentinc.com:genesys2:part0:1.1" CLK_PERIOD_NS="20"
+	cd fpga && make BOARD="kc705" XILINX_PART="xc7k325tffg900-2" XILINX_BOARD="xilinx.com:kc705:part0:1.5" CLK_PERIOD_NS="20"
```
zarubaf: We probably need some more graceful handling here. Maybe we can pass a `board` parameter on which we select the appropriate part and board.

fpga/src/ariane_xilinx.sv
```diff
@@ -43,6 +44,39 @@ module ariane_xilinx (
   output logic [ 7:0]  led         ,
   input  logic [ 7:0]  sw          ,
   output logic         fan_pwm     ,
+`elsif KC705
```
zarubaf: ~I'd suggest you define a new common define e.g. `XILINX_7SERIES` which you set if either `GENESYSII` or `KC705` is set. Something like this.~
```
`ifdef GENESYSII
  `define XILINX_7SERIES
`elsif KC705
  `define XILINX_7SERIES
`endif
 ```

~This should allow you to group most of the stuff together (and keep replication a bit lower). Yeah I know, lot of ifdef party.~

NVM, just saw that you have the `KINTEX` define. I think we are better off if we use this for the code shared between `GENESYSII` and `KC705`

Makefile
Steinegger: Yeah, I agree (un-/)commenting the respective lines isn't pretty. I couldn't figure out the way you'd want this to be handled by looking at the code for the VCU118.
An export with the board name, maybe defaulting to the genesys2, seems like a good approach though.

jrrk: kc705 and GenesysII share the same Kintex FPGA, but not the same Ethernet. So perhaps the Ethernet needs to be packaged differently. I suggest a toplevel which instances Ethernet and SRAM interface, along with a fairly generic Ariane top level which mops up all the unix platform stuff like the PLIC and DM_TOP. You also need to make build directories depend on the architecture to prevent confusing the Vivado tool.

Steinegger: > We probably need some more graceful handling here. Maybe we can pass a `board` parameter on which we select the appropriate part and board.

I updated the Makefile to use the BOARD environmental variable to set the `BOARD`, `XILINX_PART`, `XILINX_BOARD` and `CLK_PERIOD_NS` parameters. It still defaults to the genesys2 board.

PR #277, opened 2019-07-10
:arrow_up: [fpu] Bump to fix divsqrt freezing bug

stmach:
Updates the FPU to v0.6.1 to fix a bug where certain issue patterns into the FP div/sqrt unit could lead to loss of in-flight operations.


zarubaf:
Thanks. The CI seems to error out because of some unsupported Verilator constructs in the FPU. Can you reproduce this on your machine by running `make verilate`?

stmach:
Oof yes I see it, didn't check Verilator compatibility in the last FPU update.
The issue is that the shift registers in the pipe are using a single signal (an array where index `i` goes into a FF to produce `i+1`). Index `0` is driven using `assign`, hence in a blocking way while the other indexes are driven by FFs (in a non-blocking way). Verilator doesn't support this (as it handles only entire signals by name).
Gonna rewrite the pipe regs a bit to please Verilator...

stmach:
Never mind, I added this particular warning to the (long) list of ignored warnings for Verilator. Tests are still passing for me (Verilator v4.014) so it should be good 🤞

PR #278, opened 2019-07-10
:wrench: Add mul and fp tests to CI

stmach:
Adds CI runs for mul (`rv64um*`) and fp (`rv64uf*` / `rv64ud*`) asm tests for both cache configurations.


stmach:
Sorry for pressing the 'update branch' button, the CI would have failed on the initial PR without #277 also in master, but by the time the runner finally got to checking out the repo it had already been merged 🙈

PR #288, opened 2019-07-19
First stab at RVFI interface

jrrk:
@zarubaf This is a first stab at implementing RVFI for Ariane. Some placeholders have not been filled in yet. Review welcome. Moreover RVFI port hierarchical references will need flattening for compatibility with older synthesis tools. Also included is the direct instruction injection facility (bypassing the icache) needed by Cambridge University's TestRIG tool which provides formal (random rather than symbolic) comparison against spike or sail models. Feedback on the veracity of this implementation is also eagerly anticipated. This methodology is based on the ibex RVFI verilator support.


zarubaf:
Thanks. Do you think you can have a look why the Verilator build is not succeeding anymore?

jrrk:
The Verilator target works for my purpose. But I realise what should happen is 
to keep the existing verilate

target and introduce a new target as the RVFI target. At the moment I am just 
interested in your

feedback as to whether the selection of the signals for the RVFI interface is a 
sensible one or not. I realise

this is incomplete and not ready to merge.

zarubaf:
Sorry, I seem to have had a (lossy) context switch between reading your request and answering. For the integer part the `RVFI` seems to be more or less covered. Personally I would find it nice if we could package up the `RVFI` into a packed struct and assemble one holistic interface on top. That also makes it easier to add signals in the future. It would be probably good to get started with basic integer instructions and then gradually move towards more CSRs, virtual memory and atomics.

jrrk:
I am intrigued by the comment in ariane.sv about the connection to flush_i of the frontend module being not entirely correct. Do you think flush_ctrl_ex would be a better signal to use given that fence and any other instructions that do not cause an exception would be more accurate to the spike behaviour if they did not flush the frontend ?

zarubaf:
I am not sure why the comment says "not entirely correct" - I can't remember what exactly I was thinking. But the flush signal which goes to the frontend also contains flushes due to branch miss-speculation, so I am pretty confident that the signal is actually correct. I am going to be out of office for the next two weeks, so I am not sure how responsive I can be.

jrrk:
Not to worry, I am thinking of rewriting my testing to make it more of a black box approach.

PR #296, opened 2019-08-06
update the bram contents of bitstream for changes in the bootrom code

furkanturan:
This is a commit to update the bitstream contents for the changes on bootrom code, without resynthesizing the design. It takes seconds to update the bitstream, so speeds up the baremetal application development.

The steps it follows are given below:

- `run.tcl` reports all the bram instances and their locations to `reports/ariane.bram.log`
- `make $(bit)` (in fpga dir) finds the target brams for bootrom, and reports them to ` work-fpga/ariane_xilinx.bootrom.mmi`
- `make fpga_update` (in top dir) executes the `make bram_update` (in fpga dir). That starts with executing the same make target in `fpga/src/bootrom` to compile the boot code and generate corresponding `bootrom.mem` file. Then, it executes the `updatemem` tool to replace the bram contents of the bitstream with this mem file.

The updated bitstream is written to `work-fpga/ariane_xilinx_updated.bit`.

For making this possible, the structure of `bootrom.sv` is changed. The 64-bit words of the bootrom code are divided into two 32-bit words, which are concatenated later. This implementation forces Vivado to instantiate two BRAMs each are 32-bit in size. These two BRAM contents are updated with mem file. If the bootrom content is not divided to two 32-bit words by hand, then Vivado does the division itself. In that case, it divides the contents to two BRAMs respectively having sizes of 36-bits and 28-bits. In this case, the first BRAM stores 32-bit of the words into its data storage, while the remaining 4-bits in into parity storage. Furthermore, the `updatemem` tool is not capable of updating the parity storage. As a result, we prefer to make the bootrom implementation by hand to avoid un-updateable BRAM use.


JeanRochCoulon:
As this PR is open since two years. I would be pleased to status on it, and close it if nobody need this feature.
@furkanturan do you plan to merge it with the master branch ?
@sjthales @jquevremont. Would you have any interest in that FPGA feature ?
Regards

sjthales:
This PR is quite old, there are likely modifications to perform to integrate it to the current repository.

On my side, it is rare that I modify the bootrom so I don't need this feature.

JeanRochCoulon:
Thanks @sjthales for your answer. Without answer from @furkanturan, I would close this PR into one week from now.
@zarubaf, do not hesitate to comment !

furkanturan:
Sorry for my late reply. We used the Ariane platform for a project at that time, and committed some minor feature changes to accelerate the workflow. But, to be honest, the project was done quickly and we did not touch the source code afterwards. 

In the end, it has been years and I do not remember the details, so I am fine with closing the PR.

Kind Regards,

PR #307, opened 2019-08-20
add default assignment for miss_handler.sv:active_serving_o

niwis:
This resolves undefined behavior of `active_serving_o`


zarubaf:
Thanks.

PR #315, opened 2019-09-03
sync: Replace `sync_*` by `clint_sync_*`

zarubaf:
This should avoid naming conflicts with Amazon's F1 instances.


zarubaf:
@Jbalkind Can you check whether this solves your problem?

grigoriy-chirkov:
Yep, it does fix our problems, thanks!

PR #335, opened 2019-10-09
Add VC707 compatibility

jmason827:
These changes enable the production of a VC707 bitstream.


fpga/scripts/write_cfgmem.tcl
```diff
@@ -25,4 +25,9 @@ if {$argc < 2 || $argc > 4} {
 lassign $argv mcsfile bitfile
 
 # https://scholar.princeton.edu/jbalkind/blog/programming-genesys-2-qspi-spi-x4-flash
-write_cfgmem -format mcs -interface SPIx4 -size 256  -loadbit "up 0x0 $bitfile" -file $mcsfile -force
+
+# GENESYS II
+# write_cfgmem -format mcs -interface SPIx4 -size 256  -loadbit "up 0x0 $bitfile" -file $mcsfile -force
+
+# VC707
+write_cfgmem -format mcs -interface bpix16 -size 128 -loadbit "up 0x0 $bitfile" -file $mcsfile -force
```
zarubaf: Can we make this dependent on the board variable as well? Otherwise this breaks for the Gensys II flow.

fpga/scripts/program.tcl
```diff
@@ -18,9 +18,14 @@
 open_hw
 
 connect_hw_server -url localhost:3121
-open_hw_target {localhost:3121/xilinx_tcf/Digilent/200300A8CD43B}
+open_hw_target {localhost:3121/xilinx_tcf/Digilent/210203A5FC70A}
```
zarubaf: Although that is not part of the standard flow I'd appreciate a switch on board variable as well. So that me and you can use this script :-)

jmason827:
OK, I think that should do it. Let me know if your flow still works.

PR #336, opened 2019-10-15
Fixup `trst` in the ariane_xilinx.sv portlist

jmason827:
So we mapped `trst` to an active-high button on the VC707 -- I've changed the `ariane_xilinx.sv` portlist to reflect this.

I wasn't sure how you wanted to handle something like this, stylistically speaking. Let me know your thoughts.


zarubaf:
Sorry for taking so long and thanks for submitting this patch!

PR #340, opened 2019-10-24
Fix empty match file in vivado flow

jimmysitu:
Hi, All

I found it is better to remove the wildcard here when we use a relative path.
The previous PR #334 is still match empty file in some vivado version.


zarubaf:
Thanks, merged.

PR #344, opened 2019-11-05
Make fpu optional

JeanRochCoulon:
As FPU size is significant (114 equivalent kgates in our techno), it is critical to have the capability to remove it if it is not used by our product.

To disable the FPU, RVD and RVF are set to zero. With this configuration, the resulting RTL cannot be elaborated by the Synopsys dc_shell tool.

The proposed modification solves the issue.

Regards,
Jean-Roch

### DIFF
--- a/include/ariane_pkg.sv
+++ b/include/ariane_pkg.sv
@@ -153,15 +153,15 @@ package ariane_pkg;
     localparam int unsigned DEPTH_COMMIT = 8;
 `endif
 
-
-`ifdef PITON_ARIANE
+`define FPU_ENABLE // to enable both F and D extensions
+`ifdef FPU_ENABLE
     // Floating-point extensions configuration
     localparam bit RVF = 1'b1; // Is F extension enabled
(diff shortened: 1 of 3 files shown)


zarubaf:
Thanks for bringing this up and identifying the problem. I completely agree on your statement that the FPU should be removable. Unfortunately, our coding style discourages `ifdefs` and I'd like to try avoide them as much as possible. I've brought up an alternative pull request in #352. I'd appreciate your comment on this and also if you could try it out on your side.

JeanRochCoulon:
Yes, your coding proposal is smarter. I have tried it on my side. With the RTL change, the dc_shell tool (Synopsys) does not complain anymore in the two following cases: (RVF=0, RVD=0) and (RVF=1, RVD=1). The FPU disabling saves 118kgates, VERY GOOD!  So the change seems good to me. Thanks

zarubaf:
Thanks a lot for catching this. Merged into master.

PR #357, opened 2019-12-02
ariane.cfg: enable address translation per default when debugging

niwis:
Enable the OpenOCD option to perform memory accesses according to the current privilege level as per https://github.com/riscv/riscv-openocd/pull/386.
i.e. when debug mode is entered from a context running on virtual memory, address translation is performed for memory reads and writes through the debugger.


niwis:
This would throw an invalid subcommand error during startup and abort.

zarubaf:
Okay. Do you think we can pack it into a `try/catch` block and indicate an error to the user to get the latest openocd if he wants to use the virtual address feature?

PR #359, opened 2019-12-03
commit_stage.sv: remove redundant code block

niwis:
Or should the code block above be removed instead?


zarubaf:
I think that should be right. Can you try booting Linux with the version, just to be sure. I think the CI doesn’t check this very thoroughly.

PR #369, opened 2020-01-30
address length: make configurable physical and virtual address length

JeanRochCoulon:
**Why to request a pull for this modification?**
In RISCV specification, two types of address are defined: physical and virtual address. Physical address is max 56 bits and virtual address is 39bit (sv39) or 48bits (sv48).
In Ariane implementation, physical and virtual addresses are implemented on 56 and 64bits.
Address length has impact on a lot of functionalities, scoreboard or caches for instance. By implementing two RTL variables, PLEN (as physical length) and VLEN (as virtual length), address length becomes configurable. By setting PLEN=32 and VLEN=39, kgate reduction is about 40kgates.
I think it would be nice to integrate such a modification in main line to stick to the RISCV specification and reduce the gate count.

**Modification:**
-	Only Ariane has been modified, not testbench. No impact on Ariane IOs.
-	36 files have been modified.
-	As STD_CACHE is deprecated, it has not been modified. Keep the original configuration (VLEN=64 and PLEN=56) when using STD_CACHE.
-      The following constraint must be met: PLEN > VLEN
-	Regression suite is PASS when VLEN/PLEN=64/56 and 39/32.
-	Test traces have been compare cycle by cycle. No discrepancy has been detected between before and after modification.
-	No additional Verilator warnings due to this modification
-	Synthesis shows a 10% of gate reduction.

++,
Jean-Roch

### DIFF
--- a/src/load_store_unit.sv
+++ b/src/load_store_unit.sv
@@ -83,9 +83,14 @@ module load_store_unit #(
     // ------------------------------
     // virtual address as calculated by the AGU in the first cycle
     logic [riscv::VLEN-1:0]   vaddr_i;
+    logic [63:0]              vaddr64;
+    logic                     overflow;
     logic [7:0]               be_i;
 
-    assign vaddr_i = $unsigned($signed(fu_data_i.imm[riscv::VLEN-1:0]) + $signed(fu_data_i.operand_a[riscv::VLEN-1:0]));
+    assign vaddr64 = $unsigned($signed(fu_data_i.imm) + $signed(fu_data_i.operand_a));
+    assign vaddr_i = vaddr64[riscv::VLEN-1:0];
+    // we work with SV39, so if VM is enabled, check that all bits [64:riscv::38] are equal
+    assign overflow = !((&vaddr64[63:riscv::VLEN-1]) == 1'b1 || (|vaddr64[63:riscv::VLEN-1]) == 1'b0);
 
     logic                     st_valid_i;
     logic                     ld_valid_i;
@@ -331,8 +336,7 @@ module load_store_unit #(
             end
         end
 
-        // we work with SV39, so if VM is enabled, check that all bits [riscv::VLEN-1:38] are equal
-        if (en_ld_st_translation_i && !((&lsu_ctrl.vaddr[riscv::VLEN-1:38]) == 1'b1 || (|lsu_ctrl.vaddr[riscv::VLEN-1:38]) == 1'b0)) begin
+        if (en_ld_st_translation_i && lsu_ctrl.overflow) begin
 
             if (lsu_ctrl.fu == LOAD) begin
                 misaligned_exception = {
@@ -357,7 +361,7 @@ module load_store_unit #(
     // new data arrives here
     lsu_ctrl_t lsu_req_i;
 
-    assign lsu_req_i = {lsu_valid_i, vaddr_i, fu_data_i.operand_b, be_i, fu_data_i.fu, fu_data_i.operator, fu_data_i.trans_id};
+    assign lsu_req_i = {lsu_valid_i, vaddr_i, overflow, fu_data_i.operand_b, be_i, fu_data_i.fu, fu_data_i.operator, fu_data_i.trans_id};
 
     lsu_bypass lsu_bypass_i (
         .lsu_req_i          ( lsu_req_i   ),
--- a/src/load_store_unit.sv
+++ b/src/load_store_unit.sv
@@ -82,21 +82,21 @@ module load_store_unit #(
     // Address Generation Unit (AGU)
     // ------------------------------
     // virtual address as calculated by the AGU in the first cycle
-    logic [63:0] vaddr_i;
-    logic [7:0]  be_i;
+    logic [riscv::VLEN-1:0]   vaddr_i;
+    logic [7:0]               be_i;
 
-    assign vaddr_i = $unsigned($signed(fu_data_i.imm) + $signed(fu_data_i.operand_a));
+    assign vaddr_i = $unsigned($signed(fu_data_i.imm[riscv::VLEN-1:0]) + $signed(fu_data_i.operand_a[riscv::VLEN-1:0]));
 
     logic                     st_valid_i;
     logic                     ld_valid_i;
     logic                     ld_translation_req;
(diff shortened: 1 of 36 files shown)


src/load_store_unit.sv
```diff
@@ -318,33 +318,33 @@ module load_store_unit #(
...
-                    lsu_ctrl.vaddr,
+                    {{64-riscv::VLEN{1'b0}},lsu_ctrl.vaddr},
                     1'b1
                 };
             end
         end
 
-        // we work with SV39, so if VM is enabled, check that all bits [63:38] are equal
-        if (en_ld_st_translation_i && !((&lsu_ctrl.vaddr[63:38]) == 1'b1 || (|lsu_ctrl.vaddr[63:38]) == 1'b0)) begin
+        // we work with SV39, so if VM is enabled, check that all bits [riscv::VLEN-1:38] are equal
```
zarubaf: From privileged spec: 4.4.1
> Instruction fetch addresses and load and store effective addresses, which are 64 bits, must have bits 63–39 all equal to bit 38, or else a page-fault exception will occur.

We will likely miss this if `VLEN < 64`, right?

JeanRochCoulon: Yes, thank you for your feedback. Discarding the check when VLEN<64 would be a regression, not acceptable. Please look at my commit, I have just pushed a RTL modification to enable the check including the case VLEN<64.

JeanRochCoulon:
First of all, thank you for your comments. I was obliged to read back Riscv specifications and other things to clarify the point. 

Indeed, I agree, the VLEN/PLEN=39/32 is not the legal configuration.

From Privilege spec, in 4.3, 4.4 and 4.5 chapters, according to the sv32, sv39 or sv48 configuration, we understand that VLEN can be 32, 39 or 48 bits long. In same chapters, we understand that PLEN can be 34 or 56 bits long.

Privilege spec 3.6.1 chapter says that “Not all physical address bits may be implemented”. RiscV spec allows the PLEN reduction if needed by RiscV implementations.

Wikipedia https://en.wikipedia.org/wiki/Virtual_memory says that “In system with virtual  Software within the operating system may extend these capabilities to provide a virtual address space that can exceed the capacity of real memory and thus reference more memory than is physically present in the computer”. VLEN > PLEN is possible.

In Ariane, the supported configuration is sv39. VLEN/PLEN=64/56 is a possible configuration. But according to the above points, VLEN/PLEN=39/32 is also possible. Allowing the VLEN/PLEN configuration would help industrials to customize Ariane according to their needs, by reducing gate count for instance. 

Please do not hesitate to comment, your feedbacks are appreciated.

src/mmu.sv
```diff
@@ -198,9 +198,9 @@ module mmu #(
         // AXI decode error), or when PTW performs walk due to ITLB miss and raises
         // an error.
         if (enable_translation_i) begin
-            // we work with SV39, so if VM is enabled, check that all bits [63:38] are equal
-            if (icache_areq_i.fetch_req && !((&icache_areq_i.fetch_vaddr[63:38]) == 1'b1 || (|icache_areq_i.fetch_vaddr[63:38]) == 1'b0)) begin
-                icache_areq_o.fetch_exception = {riscv::INSTR_ACCESS_FAULT, icache_areq_i.fetch_vaddr, 1'b1};
+            // we work with SV39, so if VM is enabled, check that all bits [riscv::VLEN-1:38] are equal
+            if (icache_areq_i.fetch_req && !((&icache_areq_i.fetch_vaddr[riscv::VLEN-1:38]) == 1'b1 || (|icache_areq_i.fetch_vaddr[riscv::VLEN-1:38]) == 1'b0)) begin
+                icache_areq_o.fetch_exception = {riscv::INSTR_ACCESS_FAULT, {{64-riscv::VLEN{1'b0}}, icache_areq_i.fetch_vaddr}, 1'b1};
```
zarubaf: Are you sure zero extending is the right thing to do with the `mtval` field.  I would have said that sign extending is the right thing to do here. Happy to get some pointers in case I am wrong.

src/mmu.sv
```diff
@@ -225,7 +225,7 @@ module mmu #(
                 // we got an access error
                 if (iaccess_err) begin
                     // throw a page fault
-                    icache_areq_o.fetch_exception = {riscv::INSTR_PAGE_FAULT, icache_areq_i.fetch_vaddr, 1'b1};
+                    icache_areq_o.fetch_exception = {riscv::INSTR_PAGE_FAULT, {{64-riscv::VLEN{1'b0}}, icache_areq_i.fetch_vaddr}, 1'b1};
```
zarubaf: Same thinking for the other `tval` fields, here and further down.

zarubaf:
Completely agree with your observation. Now that the all ones/zeros check is in place there is only one minor thing I am not 100% sure about: The address in the `tbal/badaddr` field should IMO be sign-extended, not zero-extended. But I am not fully sure. Happy to get your input.

src/mmu.sv
JeanRochCoulon: By following the icache_areq_o.fetch_exception signal through the (VLENAriane hierarchy, I have found out only one code line which uses fetch_exception.tval. It is in issue_read_operands.sv file,
assign orig_instr = riscv::instruction_t'(issue_instr_i.ex.tval[31:0]);
To my mind, the icache_areq_o.fetch_exception is used to log the instruction which has generated the fault. In that case, as sign extension is done with bit number VLEN-1, it does not make sense to do it. Please highlight if I am wrong.

Your comment helped me to challenge the "address length" impact on the software interface. As it is very recommended to keep the same values in CSR register to limit the software impact,  two commits allow to keep the same CSR register values than before "address length" modification:
- 8f577fa6a to sign extend the flu/load/store/fpu_exceptions
- 870d to sign extend the pc inside csr registers

zarubaf:
Thanks for addressing this. I will just quickly try building an FPGA image and booting Linux again and then merge. Is there anything you want to add/request from your side regarding this PR?

JeanRochCoulon:
No, I have nothing to add to the PR. I am curious to know about FPGA and Linux validation results. Thanks

zarubaf:
Looks good from a Linux booting side. Going to merge. Thanks!

PR #372, opened 2020-02-03
Uncached loads bug (buggy, do not merge)

jrrk2:
Dear Florian,
  This simulation demonstrates the bug with your fix to issue-356. You can find everything you need here to reproduce the problem, you can also find waveforms and simulation log here:

https://www.cl.cam.ac.uk/~jrrk2/ariane/uncached_loads_bug.zip

You can see the following erroneous behaviour:

          3206410000: SPI write (addr=0000000020000020,data=0000000000000000)
          3206510000: SPI read (addr=0000000020000020,data=0000000000000000)
          3206530000: SPI read (addr=0000000020000020,data=000000000000000a)
          3206630000: SPI write (addr=0000000020000028,data=0000000000000000)

The CPU starts to read from the peripheral while there are still three writes waiting to drain. This is different to what I assumed before, that reads were coming from the write buffer. These $display commands are actually recording the behaviour at the peripheral.

As mentioned earlier, I proposed a fix to this fix, but something else is still going wrong.

I hope this helps to expedite the investigation. Let me know ASAP if I can help further.


zarubaf:
Closing because of a fix was merged with #368

PR #375, opened 2020-02-06
Proposal for protection from accessing missing physical memory regions

jrrk2:
At the moment Ariane can encounter a fatal error if it accesses a non-existent physical region. Neither does it support the ability of the AXI bus to return an error due to non-existence of memory. This proposal introduces an extra layer of protection that traps if non-existent memory is accessed, before generating an AXI transaction. This brings the behaviour roughly into line with what Rocket does.

### DIFF
--- a/include/riscv_pkg.sv
+++ b/include/riscv_pkg.sv
@@ -289,6 +289,7 @@ package riscv;
     localparam logic [63:0] ENV_CALL_MMODE        = 11; // environment call from machine mode
     localparam logic [63:0] INSTR_PAGE_FAULT      = 12; // Instruction page fault
     localparam logic [63:0] LOAD_PAGE_FAULT       = 13; // Load page fault
+    localparam logic [63:0] PHYSICAL_PROT_FAULT   = 14; // Physical protection fault (non-standard)
     localparam logic [63:0] STORE_PAGE_FAULT      = 15; // Store page fault
     localparam logic [63:0] DEBUG_REQUEST         = 24; // Debug request
 
--- a/src/mmu.sv
+++ b/src/mmu.sv
@@ -257,6 +257,7 @@ module mmu #(
     logic        dtlb_hit_n,      dtlb_hit_q;
     logic        dtlb_is_2M_n,    dtlb_is_2M_q;
     logic        dtlb_is_1G_n,    dtlb_is_1G_q;
+    logic match_any_physical_region;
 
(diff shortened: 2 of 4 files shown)


zarubaf:
Thanks for the pull request. The changes look good to me. Can you elaborate a bit on the non-standard extension? Is this something Rocket also provides? Will this eventually be standardized?

jrrk2:
This is an implementation of RISCV privilege spec section 3.5.1, it is non-standard in that I am not sure what exception should be triggered when the access fails. The important thing is to fault precisely when the address is known rather than relying on the AXI sub-system.

zarubaf:
Ah, I see. I would say, that either an illegal load/store/instruction fetch exception would be triggered (as for all PMAs which are violated). What do you think? I think this is a good proposal if we just throw an illegal access exception.

jrrk2:
I feel neutral about the precise exception number. The main thing is to make it a precise exception, to not forward invalid addresses to the AXI bus, and for it to be simple enough to not have much impact on timing.

zarubaf:
Okay, thanks for proposing. Would you be up to changing the exception number to either a load or store illegal access exception depending on the type of request? I am happy to merge this then.

include/riscv_pkg.sv
```diff
@@ -289,6 +289,7 @@ package riscv;
     localparam logic [63:0] ENV_CALL_MMODE        = 11; // environment call from machine mode
     localparam logic [63:0] INSTR_PAGE_FAULT      = 12; // Instruction page fault
     localparam logic [63:0] LOAD_PAGE_FAULT       = 13; // Load page fault
+    localparam logic [63:0] PHYSICAL_PROT_FAULT   = 14; // Physical protection fault (non-standard)
```
zarubaf:
```suggestion
```

src/mmu.sv
```diff
@@ -331,9 +332,17 @@ module mmu #(
                         lsu_exception_o = {riscv::LOAD_PAGE_FAULT, {25'b0, update_vaddr}, 1'b1};
                     end
                 end
+            end // if (ptw_active && !walking_instr)
             end
+        // if it didn't match any physical region throw a (non-standard) `Physical Protection Fault`
+        if (lsu_valid_o && !match_any_physical_region) begin
```
zarubaf: We should probably distinguish between load and stores here and throw the appropriate illegal access exception.

jrrk:
I’m off sick at the moment, however I ticked the ‘allow maintainer to modify pull request’

PR #378, opened 2020-02-19
scoreboard.sv: support FU NONE

niwis:
If an instruction does not require a functional unit we want to commit it right away.


zarubaf:
Thanks! Can you confirm that Linux ist still booting?

niwis:
Yup

zarubaf:
Thanks!

PR #388, opened 2020-03-10
Benderization of Ariane

WRoenninger:
This PR introduces [bender](https://github.com/fabianschuiki/bender) for dependency management of Ariane.
* Git submodules have been removed and the sources placed into the folder deps by bender. Most of the dependencies have been updated, see `CHANGELOG`. Notably the AXI infrastructure has been revamped.
* The Makefile has been adapted to use the compilation scripts generated by bender for FPGA and Vsim.
* Verilator simulation gets compiled by parsing `scripts/sources.json` in the Makefile through a python3 script in `scripts/parse_bender_verilator.py`.
* Peripherals module is now the same for FPGA and Testharness (src/fpga/peripherals_xilinx.sv). In the Testharness it is configured to not use any Xilinx IPs and answers with salve errors on the AXI.
Note / Todos:
* AXI has still to be updated to a release tag, as the dwc is not officially released jet.
* README has to be updated.

PR #392, opened 2020-03-13
Add Bender Dependencies (1)

WRoenninger:
First part of #388. This adds the soucre dependencies in a seperate folder `deps/` generated by bender in Part 2.


.gitignore
```diff
@@ -31,3 +31,7 @@ stdout/
 work-dpi/
 tb/riscv-isa-sim/
 work-*/*
+deps/fpu/src/common_cells*
+deps/fpu/src/fpu_div_sqrt_mvp*
+deps/fpu/tb*
```
zarubaf: Why do we ignore those files?

WRoenninger: These folders are git submodules inside `fpnew`, which get automatically checked out with bender.

WRoenninger: `common_cells` and `fpu_div_sqrt_mvp` are checked out separately in `deps/`

zarubaf: Is there a reason to:
1. Don't check them in?
2. I think you can disable the recursive checkout in the `Bender.yml`, please help @fabianschuiki

WRoenninger: Recursive checkout disable is currently not supported by bender.

WRoenninger: The checked out submodules get checked out by bender anyways, except the test bench of `fpnew`, which  does not matter in the context of Ariane.

Issue #381, opened 2020-02-21
Wrong Illegal instruction fault for compressed fld ft0, xxx(sp)


xushengj:
If a compressed instruction fld rd, xxx(sp) use ft0 as rd, currently Ariane will generate an illegal instruction fault, caused by the check at src/compressed_decoder.sv:205:
`if (instr_i[11:7] == 5'b0)  illegal_instr_o = 1'b1;`
However, the restriction that destination shouldn't be zero should only be applied to GPR.

Suggest to remove src/compressed_decoder.sv:205 to fix the fault.

zarubaf:
Very nice catch, thanks! Would you be willing to contribute a patch? Thanks!

PR #395, opened 2020-03-13
Removed the check in c.fldsp against ft0 destination

xushengj:
PR for issue #381 

If a compressed instruction fld rd, xxx(sp) use ft0 as rd, currently Ariane will generate an illegal instruction fault. However, the restriction that destination shouldn't be zero should only be applied to GPR; floating point register ft0 is not hard-wired to zero and can be used as destination.

This pull request removed the check in compressed decoder that causes this issue.


zarubaf:
Thanks!

PR #402, opened 2020-04-03
Initial support for Riviera-PRO 2020.04

pawelkudlakaldec:


zarubaf:
Sorry for the long radio silence. Is this PR still a valid request of yours (because you started to submit smaller patches)?

If so, I think we need to proceed in a slightly different approach and making the patches smaller. I, unfortunately, I also can't accept simulator specific patches (at least not in synthesizable RTL). If we violate the LRM we should re-write the offending RTL. If there is one worrying construct we can try to find another way to express it, but the `ifdefs` are imho a very hacky solution.

I will close for now, feel free to re-open or create a new PR.

danielmlynek:
We are aware of disadvantage of having per simulator ifdefs. Sometimes there is no ither way to support many tools.
We will prepare smaller pull request with no ifdefs usage so we can move forward.
If some ifdef cannot be avoided for us we will make it clear

PR #411, opened 2020-04-11
Adding mcounteren and scounteren registers

Jbalkind:
Wanted to address some interest in having `rdcycle` in S and U modes since the privileged spec allows this kind of delegation. Unfortunately I don't have a very minimal test case for this to check both modes, but I can see that the previous illegal access exceptions no longer appear when using `rdcycle` provided that `mcounteren`/`scounteren` have the right values.

Probably needs some further testing.

Also the changes I made are all in csr_regfile.sv. I don't know what changes need to be done to enable readtime to work properly, too.

### DIFF
--- a/src/csr_regfile.sv
+++ b/src/csr_regfile.sv
@@ -115,12 +115,14 @@ module csr_regfile #(
     logic [63:0] mideleg_q,   mideleg_d;
     logic [63:0] mip_q,       mip_d;
     logic [63:0] mie_q,       mie_d;
+    logic [63:0] mcounteren_q,mcounteren_d;
     logic [63:0] mscratch_q,  mscratch_d;
     logic [63:0] mepc_q,      mepc_d;
     logic [63:0] mcause_q,    mcause_d;
     logic [63:0] mtval_q,     mtval_d;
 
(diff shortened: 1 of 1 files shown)


src/csr_regfile.sv
```diff
@@ -841,6 +847,15 @@ module csr_regfile #(
             if (csr_addr_i[11:4] == 8'h7b && !debug_mode_q) begin
                 privilege_violation = 1'b1;
             end
+            // check counter-enabled counter CSR accesses
+            // counter address range is C00 to C1F
+            if (csr_addr_i[11:5] == 7'b1100000) begin
```
zarubaf, line 852: I'd prefer something which is a bit more descriptive:
```suggestion
            if (csr_addr_i inside {[riscv::CSR_CYCLE:riscv::CSR_HPM_COUNTER_31]}) begin
```

Jbalkind:
I tried 2 core OpenPiton+Ariane on Genesys2 and Fei tried 3 core on VC707. Both still boot Linux to a user-mode shell as intended

zarubaf:
Thanks a lot, a very useful addition.

PR #416, opened 2020-04-27
move 1 package import into module

danielmlynek:


src/cache_subsystem/std_nbdcache.sv
```diff
@@ -13,7 +13,6 @@
 // Description: Nonblocking private L1 dcache
 
 import ariane_pkg::*;
-import std_cache_pkg::*;
 
 module std_nbdcache #(
```
zarubaf, line 17: I would also prefer to move it here, just for consistency:
```suggestion
module std_nbdcache import std_cache_pkg::*; #(
```

zarubaf: If that's not possible I'll also accept the current version ;-)

pawelkudlakaldec: Your proposition is OK for us. Changes have been committed.

PR #431, opened 2020-05-29
Separate NC and NI Load/Stores, and fix bugs

morenes:
The changes involve separating Non-Cacheable(NC) and Non-idempotent(NI) operations, cause Florian fix treat them as the same. The behavior is as follows:

Regarding Load operations:
--NC but Idempotent Loads would not need to wait until the WB is drained, and can be executed speculatively, as normal cacheable loads.
--NC and NI Loads would wait for the WB to be drained only if there are NI stores ongoing, and they cannot be issued speculatively.

Regarding Store operations:
--NC but Idempotent Stores could enter the WB in the same rules as a normal Cacheable Store.
--NC and NI Stores wouldn't enter the WB if there is another NC and NI Store inside it, but they could enter otherwise, without needing to halt other Cacheable operations to enter the WB, since they belong to different mem regions, they wouldnt forward (load) or coalesce (stores).

Note: NI operations are always NC too, but not viceversa.


morenes:
I would still need to refactor the name of some signals, but I created the PR to check that the CI checks worked!

zarubaf:
Thanks for putting this together. Let me know when you want to have a review from me, I'll have a reduced response time the next two weeks as I am on holidays. 

One very high-level observation we use SV `logic` type for `wire` and `reg`.

morenes:
yeah, I know, but logic produces bad behavior when it is assigned at declaration time
logic signal_a = signal_b;
doesnt work as It could be expected, and thats why i used wire,
is that a problem? I could refactor it so I have the logic declaration and assign separate if necessary.
I think the commit is now ready for review If you can have a look.
FYI, I reached Jonathan Kimmit, but he mentioned he left the LowRisc organization and he did not really reach the bottom of the issues…
I observed that the CI passed and our ends works too.
Thanks in advance for the review,

Marcelo

zarubaf:
We are mostly in line with the lowRISC coding style which recommends `logic` over the other two constructs 
https://github.com/lowRISC/style-guides/blob/master/VerilogCodingStyle.md#preferred-systemverilog-constructs

So, yes, please re-factor them. Just define the signal and use `assign` to achieve similar behavior.

src/cache_subsystem/wt_dcache_wbuffer.sv
```diff
@@ -134,11 +135,11 @@ module wt_dcache_wbuffer #(
...
 ///////////////////////////////////////////////////////
-
-  assign miss_nc_o = nc_pending_q;
-
+  wire [ariane_pkg::DCACHE_TAG_WIDTH-1:0] miss_tag = miss_paddr_o[ariane_pkg::DCACHE_INDEX_WIDTH+:ariane_pkg::DCACHE_TAG_WIDTH];
+  wire is_nc_miss = !ariane_pkg::is_inside_cacheable_regions(ArianeCfg, {{64-DCACHE_TAG_WIDTH{1'b0}}, miss_tag, {DCACHE_INDEX_WIDTH{1'b0}}});
+  assign miss_nc_o = !cache_en_i || is_nc_miss; 
   // noncacheable if request goes to I/O space, or if cache is disabled
-  assign addr_is_nc = (~cache_en_i) | (~ariane_pkg::is_inside_cacheable_regions(ArianeCfg, {{64-DCACHE_TAG_WIDTH{1'b0}}, req_port_i.address_tag, {DCACHE_INDEX_WIDTH{1'b0}}}));
+  wire is_ni_i = ariane_pkg::is_inside_nonidempotent_regions(ArianeCfg, {{64-DCACHE_TAG_WIDTH{1'b0}}, req_port_i.address_tag, {DCACHE_INDEX_WIDTH{1'b0}}});
```
zarubaf, line 142: `_i` is reserved for primary input and outputs

src/cache_subsystem/wt_dcache_wbuffer.sv
```diff
@@ -390,11 +390,16 @@ module wt_dcache_wbuffer #(
   assign req_port_o.data_rdata  = '0;
 
   assign rd_hit_oh_d = rd_hit_oh_i;
+  
+  wire ni_inside = |ni_pending_q;
+  wire ni_conflict = is_ni_i && ni_inside;
```
zarubaf, line 395:
```suggestion
  wire ni_conflict = is_ni_i & ni_inside;
```

src/cache_subsystem/wt_dcache_wbuffer.sv
```diff
@@ -390,11 +390,16 @@ module wt_dcache_wbuffer #(
...
+  assign not_ni_o = !ni_inside;
```
zarubaf, line 396:
```suggestion
  assign not_ni_o = ~ni_inside;
```

src/cache_subsystem/wt_dcache_wbuffer.sv
```diff
@@ -390,11 +390,16 @@ module wt_dcache_wbuffer #(
...
+  assign empty_o    = !(|valid);
```
zarubaf, line 397:
```suggestion
  assign empty_o    = ~(|valid);
```

src/cache_subsystem/wt_dcache_wbuffer.sv
```diff
@@ -451,11 +457,11 @@ module wt_dcache_wbuffer #(
     if (req_port_i.data_req && rdy) begin
       // in case we have an NC address, need to drain the buffer first
```
zarubaf, line 458:
```suggestion
      // in case we have an NI address, need to drain the buffer first
```

src/cache_subsystem/wt_dcache_wbuffer.sv
```diff
@@ -451,11 +457,11 @@ module wt_dcache_wbuffer #(
...
       // in case we are serving an NC address,  we block until it is written to memory
```
zarubaf, line 459:
```suggestion
      // in case we are serving an NI address,  we block until it is written to memory
```

src/cache_subsystem/wt_dcache_wbuffer.sv
```diff
@@ -134,11 +135,11 @@ module wt_dcache_wbuffer #(
 ///////////////////////////////////////////////////////
 // misc
 ///////////////////////////////////////////////////////
-
-  assign miss_nc_o = nc_pending_q;
-
+  wire [ariane_pkg::DCACHE_TAG_WIDTH-1:0] miss_tag = miss_paddr_o[ariane_pkg::DCACHE_INDEX_WIDTH+:ariane_pkg::DCACHE_TAG_WIDTH];
+  wire is_nc_miss = !ariane_pkg::is_inside_cacheable_regions(ArianeCfg, {{64-DCACHE_TAG_WIDTH{1'b0}}, miss_tag, {DCACHE_INDEX_WIDTH{1'b0}}});
+  assign miss_nc_o = !cache_en_i || is_nc_miss; 
   // noncacheable if request goes to I/O space, or if cache is disabled
```
zarubaf, line 141: The comments don't apply to the below code anymore. Please move it to the right code.

src/load_store_unit.sv
```diff
@@ -383,6 +387,33 @@ module load_store_unit #(
         .*
     );
 
+genvar j;
```
zarubaf, line 390: What's the purpose of this? Looks like debug infrastructure.

src/load_unit.sv
```diff
@@ -34,6 +34,7 @@ module load_unit import ariane_pkg::*; #(
     input  logic [riscv::PLEN-1:0]   paddr_i,             // physical address in
     input  exception_t               ex_i,                // exception which may has happened earlier. for example: mis-aligned exception
     input  logic                     dtlb_hit_i,          // hit on the dtlb, send in the same cycle as the request
+    input  logic [riscv::PLEN-13:0]  dtlb_ppn_i,          // ppn on the dtlb, send in the same cycle as the request
```
zarubaf, line 37: I am a bit worried on the timing impact of this design choice. That's basically a sneak path from address translation back to the LSU. But I currently don't see another way around it except a fundamental re-design of the load unit.

src/load_unit.sv
```diff
@@ -75,13 +76,16 @@ module load_unit import ariane_pkg::*; #(
     // directly output an exception
     assign ex_o = ex_i;
 
+    wire [riscv::PLEN-1:0] paddr_pre = {dtlb_ppn_i, 12'd0};
```
zarubaf, line 79:
```suggestion
    wire [riscv::PLEN-1:0] paddr_early = {dtlb_ppn_i, 12'd0};
```

src/load_unit.sv
```diff
@@ -75,13 +76,16 @@ module load_unit import ariane_pkg::*; #(
...
+    //wire paddr_nc = !is_inside_cacheable_regions(ArianeCfg, paddr_pre);
```
zarubaf, line 80: Remove stale comments.

src/load_unit.sv
```diff
@@ -75,13 +76,16 @@ module load_unit import ariane_pkg::*; #(
...
+    wire not_commit_time = commit_tran_id_i != lsu_ctrl_i.trans_id;
+    wire inflight_stores = (!dcache_wbuffer_not_ni_i || !store_buffer_empty_i);
     logic stall_nc;  // stall because of non-empty WB and address within non-cacheable region.
     // should we stall the request e.g.: is it withing a non-cacheable region
     // and the write buffer (in the cache and in the core) is not empty so that we don't forward anything
     // from the write buffer (e.g. it would essentially be cached).
-    assign stall_nc = (~(dcache_wbuffer_empty_i | store_buffer_empty_i) & is_inside_cacheable_regions(ArianeCfg, paddr_i))
-                    // this guards the load to be executed non-speculatively (we wait until our transaction id is on port 0
-                    | (commit_tran_id_i != lsu_ctrl_i.trans_id & is_inside_nonidempotent_regions(ArianeCfg, paddr_i));
+    assign stall_nc = (inflight_stores || not_commit_time) && paddr_ni;
```
zarubaf, line 88: Please clean-up this section:

- Let's not over introduce unnecessary signal names (signals which are not needed anymore such as `not_commit_time`, `inflight_stores`, `paddr_ni`
- Keep the comments descriptive, up-to-date, and in the right place.

src/load_unit.sv
```diff
@@ -121,7 +125,7 @@ module load_unit import ariane_pkg::*; #(
                             // translation valid but this is to NC and the WB is not yet empty.
                             end else if (dtlb_hit_i && stall_nc) begin
                                 state_d = ABORT_TRANSACTION_NC;
-                            end else begin
+                            end else begin //TLB miss
```
zarubaf, line 128:
```suggestion
                            end else begin // TLB miss
```

src/load_unit.sv
```diff
@@ -121,7 +125,7 @@ module load_unit import ariane_pkg::*; #(
                             // translation valid but this is to NC and the WB is not yet empty.
                             end else if (dtlb_hit_i && stall_nc) begin
                                 state_d = ABORT_TRANSACTION_NC;
```
zarubaf, line 127: Please re-name the state to `_NI`.

morenes:
Can you elaborate on why NC stores shouldn't merge in the WB?

zarubaf:
Just because it changes the ST->ST ordering. That could become problematic when for example a store to one bit of the register activates another register (like write-protection for example). A sequence that deactivates write-protection writes and activates write-protection again could become messy if the stores to the write-protection addresses merge (essentially nullifying them).

Yet you could argue that one should declare that NI, but actually the registers are perfectly idempotent.

src/cache_subsystem/wt_dcache_wbuffer.sv
morenes: What is the rationale for that? why do bitwise operation over boolean signals? I (and Arm Code Conventions) find very misleading to use that

src/load_store_unit.sv
morenes: oh, yes, I added this accidentally, it was my performance counters to see the average load latency. I wonder if we should include this under some macro, like `PERF_COUNTERS... Or I can just remove it

src/load_unit.sv
morenes: I think the same, I think the timing should be fine, since we are using the hit signal, which comes from the lookup comparison + an or reduction, so I guess If the Hit signal meets timing, this should meet timing too.

src/load_unit.sv
morenes: Honestly I think it eases debugging a lot to have intermediate signals, and It doesnt harm at all, only adds clarity

src/cache_subsystem/wt_dcache_mem.sv
```diff
@@ -210,36 +208,40 @@ module wt_dcache_mem #(
...
-  if (Axi64BitCompliant) begin : gen_axi_off
-      assign wr_cl_off     = (wr_cl_nc_i) ? '0 : wr_cl_off_i[DCACHE_OFFSET_WIDTH-1:3];
-  end else begin  : gen_piton_off
-      assign wr_cl_off     = wr_cl_off_i[DCACHE_OFFSET_WIDTH-1:3];
-  end
-
-  assign rdata         = (wr_cl_vld_i)  ? wr_cl_data_i[wr_cl_off*64 +: 64] :
-                                          rdata_cl[rd_hit_idx];
+  assign wbuffer_rdata = wbuffer_data_i[wbuffer_hit_idx].data;
+  assign wbuffer_be    = (wbuffer_miss) ? '0 : wbuffer_data_i[wbuffer_hit_idx].valid;
```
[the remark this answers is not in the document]
morenes: right now the diff is only the wbuffer_miss, which I extract from the lzc to reuse the logic and avoid the or reduction on the hit signal, Idk why git selects the whole area as diff

Jbalkind:
Florian I see that more as an issue of RISC-V using the name "non-idempotent" than one of the actual intent of the PMA class. I think if you consider the term "side-effecting" rather than "non-idempotent" then what you called NC should clearly be NI. The spec (section A.4.1) says "Idempotency PMAs: Idempotency PMAs are used to specify memory regions for which loads and/or stores may have side effects" - I would take that as bad naming of NI on their part and that those registers should be NI as they are side effecting.

Interestingly, for both cacheability and idempotency PMAs, A.4.1 says for both that they "do not affect the memory model", implying that ST-ST reordering may be allowed if there is no suitable fencing or aq/rl...

zarubaf:
Sgtm, then let’s just declare them as NI/w. side effect. 

So that would mean that the programmer would need to fence for such scenario. I highly doubt that any programmer and/or Linux device code does that but I could be wrong of course. For sure BBL has the UART divisor latch register which exhibits such behavior and does not fence.

Jbalkind:
Yeah I think we're best to maintain the present NI behaviour of not issuing to avoid reordering given this, but can relax things for NC (assuming it's used for bootroms, scratchpads, that sort of thing). There is something in the spec about I/O ordering (as opposed to memory ordering) but I didn't have time to read it yet.

zarubaf:
Okay, after re-reading the specification there is a special I/O PMA which handles that (each I/O transfer appeared as if a `fence io, io` would be called before and after the instruction). 
I think we can happily live with the fact that we provide this kind of ordering for NI accesses and that we provide a weak ordering for the NC access. We should just note it somewhere.

@morenes Do you have an update on the Linux booting status?

src/cache_subsystem/wt_dcache_mem.sv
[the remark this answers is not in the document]
zarubaf: Because you are committing white-space changes.

Jbalkind:
Ah should we just be specifying the I/O PMA for all the things we do NI to atm? We could easily change that in the core instantiation.

zarubaf:
Thanks closing this in favor of #492

PR #445, opened 2020-06-12
co-simulation with dromajo

kabylkas:
Hi 👋 

If short, with this change you can run co-simulation with the following:
```
make verilate DROMAJO=1
make run-dromajo-verilator BIN=/path/to/bin
```

As we had been discussing via email, this patch implements [dromajo](https://github.com/chipsalliance/dromajo) co-simulation flow that is depicted in the figure below.
image
1. Load the binary of interest into Dromajo.
2. Run Dromajo stand alone and let a couple of instructions to complete.
3. Dump the checkpoint. This is the whole architectural state of the reference model. Dromajo dumps the main and boot memories. In addition, it generates a boot code. If you were to run that code it will restore the whole architectural state. This means that you can bring any two or more cores into complete synced architectural state by running this piece of code.
4. Load the checkpoint into the RTL memory and the instance of Dromajo in RTL. Dromajo gets linked to a simulator as a shared library. RTL communicates to Dromajo through set of DPI calls.
5. Run the RTL simulation and perform co-simulation.

Please let me know if you want me to remove/add/change anything.

(P.S. Sorry, I have completed this a while back but barely had time to clean it up for the pull request.)

### DIFF
--- a/bootrom/dromajo_bootrom.sv
+++ b/bootrom/dromajo_bootrom.sv
@@ -1,0 +1,53 @@
+/* 
+ * File: dromajo_bootrom.sv
+ *
+ * Description: bootrom that gets synced with dromajo for
+ * cosimulation purposes.
+ *
+ */
+
+module dromajo_bootrom (
+   input  logic         clk_i,
+   input  logic         req_i,
+   input  logic [63:0]  addr_i,
+   output logic [63:0]  rdata_o
+);
+    localparam int RomSize = 4096;
+    logic [63:0] mem[RomSize-1:0];
+
+    initial begin
+      integer hex_file, num_bytes;
+      longint address, value;
+      string f_name;
+      // init to 0
+      for (int k=0; k<RomSize; k++) begin
+        mem[k] = 0;
+      end
+
+      // sync with dromajo
+      if ($value$plusargs("checkpoint=%s", f_name)) begin
+        hex_file = $fopen({f_name,".bootram.hex"}, "r");
+        while (!$feof(hex_file)) begin
+          num_bytes = $fscanf(hex_file, "%d %h\n", address, value);
+          $display("%d %h", address, value);
+          mem[address] = value;
+        end
+        $display("Done syncing boot ROM with dromajo...\n");
+      end else begin
+        $display("Failed syncing boot ROM: provide path to a checkpoint.\n");
+      end
+
+    end
+
+    logic [$clog2(RomSize)-1:0] addr_q;
+
+    always_ff @(posedge clk_i) begin
+        if (req_i) begin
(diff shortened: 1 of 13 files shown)


bootrom/dromajo_bootrom.sv
```diff
@@ -0,0 +1,53 @@
+/*
```
zarubaf, line 1: Can you add a license header in this file, please?

src/csr_regfile.sv
```diff
@@ -400,8 +400,6 @@ module csr_regfile #(
                     dcsr_d = csr_wdata[31:0];
                     // debug is implemented
                     dcsr_d.xdebugver = 4'h4;
-                    // privilege level
-                    dcsr_d.prv = priv_lvl_q;
```
zarubaf, line 404: Is this a bug you fixed? I can't seem to remember now :(

src/csr_regfile.sv
```diff
@@ -989,7 +990,11 @@ module csr_regfile #(
             // floating-point registers
             fcsr_q                 <= 64'b0;
             // debug signals
+`ifdef DROMAJO
+            debug_mode_q           <= 1'b1;
```
zarubaf, line 994: Why do you need to enable the debug mode after reset?

tb/dpi/dromajo_cosim_dpi.cc
```diff
@@ -0,0 +1,88 @@
+#include <svdpi.h>
```
zarubaf, line 1: We need a license header here as well.

src/csr_regfile.sv
kabylkas: This was actually fixed by you. You sent me a patch via email that I applied. Here is what you said in the email: 
>You are absolutely right about the bug, which incorrectly implements the update logic of `dcsr_d.prv`.
>
>The confusion came from the fact that the core should update its `dcsr_d.prv` to the current running privilege level when entering debug mode. I've attached a patch which, as you suggested, removes the wrong update logic in the CSR write process and should now correctly handle the update when entering debug mode (4.9.1 Debug Control and Status of the debug specification).
>
>Let me know if that makes sense. Don't hesitate to contact me if you have any questions.

src/csr_regfile.sv
kabylkas: This is related to the bullet point 3 in the PR description. When dromajo generates a bootcode it generates series of csr writes. Some of those csr writes can be executed only in debug mode, such as dpc register, for example. Also, the last instruction in the boot code is dret.

src/csr_regfile.sv
zarubaf: Perfect, sorry for the confusion 🙈

PR #472, opened 2020-07-09
PMP integration

Moschn:
This PR adds PMP support to Ariane/cv6. A few comments on the implementation:

- Adds a new PMP access check unit
- The unit is instantiated in three places in the pipeline:
  - two in the MMU (one for instruction fetch and one for data access)
  - one in the PTW that checks pt accesses
- Because Ariane does not support unaligned data accesses, we have decided to only implement granularity 3 (2^3 = 8 bytes aligned). 
- The number of PMP entries is configurable in `ariane_cfg_t`. The default is set to 8 for now (I am open to set it to 0 to not cause any breakage for users). If the number is set to 0, all of the PMP units are disabled and there should not be any hardware overhead.

This has been tested with:
- the PMP test in riscv-tests. However, the PMP test in riscv-tests works with granularity 2 (4 byte access). So in order to check the ariane implementation one has to change the test and change the granularity to 3 (or 8).
- a custom version of the keystone enclave framework
- OpenSBI boots the new core. However the configuration of OpenSBI would have to be modified to indicate that Ariane does support PMP. Without this modification, it does not work.
- Linux obviously still boots because it does not interact with the PMP entries (verified on the FPGA).


Moschn:
Actually, the riscv-test PMP benchmark fails. Let me investigate why...

Moschn:
Retested. The PMP benchmark from riscv-tests works with the following changes:

```diff
--- a/benchmarks/pmp/pmp.c
+++ b/benchmarks/pmp/pmp.c
@@ -42,7 +42,7 @@ static void init_pt()
 #else
   uintptr_t vm_choice = SATP_MODE_SV32;
 #endif
-  write_csr(sptbr, ((uintptr_t)l1pt >> RISCV_PGSHIFT) |
+  write_csr(satp, ((uintptr_t)l1pt >> RISCV_PGSHIFT) |
                    (vm_choice * (SATP_MODE & ~(SATP_MODE<<1))));
   write_csr(pmpaddr2, -1);
   write_csr(pmpcfg0, (PMP_NAPOT | PMP_R) << 16);
@@ -55,7 +55,7 @@ INLINE uintptr_t va2pa(uintptr_t va)
   return va - SCRATCH + (uintptr_t)scratch;
 }

-#define GRANULE (1UL << PMP_SHIFT)
+#define GRANULE (1UL << 3)

 typedef struct {
   uintptr_t cfg;
@@ -98,7 +98,7 @@ INLINE int pmp_ok(pmpcfg_t p, uintptr_t addr, uintptr_t size)
   return hits == 0 || hits >= size;
 }

-INLINE void test_one(uintptr_t addr, uintptr_t size)
+void test_one(uintptr_t addr, uintptr_t size)
 {
   uintptr_t new_mstatus = (read_csr(mstatus) & ~MSTATUS_MPP) | (MSTATUS_MPP & (MSTATUS_MPP >> 1)) | MSTATUS_MPRV;
   switch (size) {
```

zarubaf:
Thanks for this really, really nice contribution! Everything looks good to me! I am all in favor of keeping the PMPs configured to 8 and issue a PR to the OpenSBI repository!

Issue #301, opened 2019-08-09
Architecturally visible exception in mispredicted branch


niwis:
Hi,

We observe a `ret` instruction where the return address is wrongly predicted as `0x0` by the return address stack. The speculative execution at `0x0` triggers an `INSTRUCTION_PAGE_FAULT`, since `0x0` is not mapped. However, since it is mispredicted, it should be flushed once the misprection is detected. Yet, we see the exception occur nonetheless.

We've included a buffer that tracks some internal signals of the frontend:
<details><summary>Frontend Buffer</summary>

| Time | icache\_dreq\_o\.vaddr | priv\_lvl | Instr\[0\] | Instr\[1\] | Active Signals                                                                                                               |
|------|------------------------|-----------|------------|------------|-----------------------------------------------------------------------------------------------------------------------|
| 0    | 0xffffffff810a0000     | S         | 0x0        | 0x0        | \-                                                                                                                    |
| \-1  | 0x2c430                | U         | 0x0        | 0x0        | \-                                                                                                                    |
| \-2  | 0x2c42c                | U         | 0x87aad3df | 0x0        | \-                                                                                                                    |
| \-3  | 0x2c42a                | U         | 0xd3dfe0ef | 0x87aa     | if\_ready, icache\_dreq\_o\.kill\_s2, bp\_valid, instr\_valid\[0\], instr\_valid\[1\], rvi\_call\[0\], rvi\_jump\[0\] |
| \-4  | 0x2d6f4                | U         | 0xe0ef4501 | 0x0        | if\_ready, instr\_valid\[0\]                                                                                          |
| \-5  | 0x2d6f0                | U         | 0x45814601 | 0x4581     | if\_ready, instr\_valid\[0\], instr\_valid\[1\]                                                                       |
| \-6  | 0x2d6ec                | U         | 0xfec42783 | 0x86be     | if\_ready, instr\_valid\[0\], instr\_valid\[1\]                                                                       |
| \-7  | 0x2d6e8                | U         | 0x2783     | 0x0        | if\_ready                                                                                                             |
| \-8  | 0x2d6e4                | U         | 0x71398082 | 0x7139     | if\_ready, rvc\_return\[0\]                                                                                           |
| \-9  | 0x2d6e2                | U         | 0x71398082 | 0x7139     | if\_ready, rvc\_return\[0\]                                                                                           |
[...]

PR #474, opened 2020-07-13
abort fetch_req when kill_s2=1

masgia:
Signed-off-by: Massimiliano Giacometti <massimiliano.giacometti@hensoldt-cyber.com>

This pull request is to solve https://github.com/openhwgroup/cva6/issues/301


zarubaf:
Hopefully, this bug is fixed now that we switched to a single instruction cache implementation. Can you confirm that? Please feel free to re-open in case that doesn't solve your problem.

PR #485, opened 2020-07-27
Code style correction of cva6

### LINKED ISSUE #477: Prefixing of module names and file names
In CV32E40P all module names and file names were prefixed with cv32e40p_ to avoid naming collisions at the chip level, see https://github.com/openhwgroup/cv32e40p/issues/276.

Can the same be done for all RTL files in CVA6 (I would propose a cva6_ prefix)?

In CV32E40P the following naming style is used:

- Filename : cv32e40p_\<something\>.sv
- Module name: cv32e40p_\<something\>
- Instantiation: cv32e40p_\<something\> () \<something\>_i

whutddk:
## fix #477 
Hello all, according to #477 , I have add prefixs "cva6_" to most of the RTL files, modules, and instantiation. So far, my tests only prove that the flow ```make fpga```, which is for ```Genesys II``` is working properly. I will also test the other flow soon, because macro may block some bugs, which  miss prefixs. 

----------------------
The mainly changes are as followed:

* file name, module name, instantiation name 
* ariane -> cva6
* riscv_pkg -> cva6_riscv_pkg
* ariane_pkg -> cva6_pkg
* std_cache_pkg -> cva6_std_cache_pkg
* wt_cache_pkg -> cva6_wt_cache_pkg

---------------------

The RTL codes in submodule are not modified, by the way.


Jbalkind:
Your modifications under the openpiton/ directory will break openpiton downstream. Please submit an appropriate PR to openpiton if you are going to change those filenames and interface wires.

whutddk:
@Jbalkind Yes I understand, and modifying the code style is a huge job and easy to make mistakes. All related repositories should be changed at the same time to avoid errors. I will try to do that soon. And only after that, we can merge the PR.

PR #489, opened 2020-07-31
remove global import

### LINKED ISSUE #482: Do not use global imports of packages in the compilation unit scope unless necessary
The use of global imports of packages (as for example done here https://github.com/openhwgroup/cva6/blob/c57b275df595601cae85eed6b3b4d47639a6f510/src/ariane.sv#L25) should be avoided. For further justification please see the corresponding issue in CV32E40P https://github.com/openhwgroup/cv32e40p/issues/384. In CV32E40P all such imports were fixed in https://github.com/openhwgroup/cv32e40p/pull/389 and ideally the same is done in CVA6.

whutddk:
## fix #482 
Global imports have been fixed in this PR.

pkg name in this PR may conflict with the pkg name in #485  , but don't worry, I will solve this no matter which PR is merged first.

### DIFF
--- a/include/ariane_pkg.sv
+++ b/include/ariane_pkg.sv
@@ -481,7 +481,7 @@ package ariane_pkg;
         logic [TRANS_ID_BITS-1:0] trans_id;
     } fu_data_t;
 
-    function automatic logic isBranch (input fu_op op);
+    function automatic logic op_is_branch (input fu_op op);
         unique case (op) inside
             EQ, NE, LTS, GES, LTU, GEU: return 1'b1;
             default                   : return 1'b0; // all other ops
--- a/include/ariane_pkg.sv
+++ b/include/ariane_pkg.sv
@@ -481,7 +481,7 @@ package ariane_pkg;
         logic [TRANS_ID_BITS-1:0] trans_id;
     } fu_data_t;
 
-    function automatic logic is_branch (input fu_op op);
+    function automatic logic isBranch (input fu_op op);
         unique case (op) inside
             EQ, NE, LTS, GES, LTU, GEU: return 1'b1;
             default                   : return 1'b0; // all other ops
--- a/src/branch_unit.sv
+++ b/src/branch_unit.sv
@@ -65,7 +65,7 @@ module branch_unit (
             resolved_branch_o.target_address = (branch_comp_res_i) ? target_address : next_pc;
             resolved_branch_o.is_taken = branch_comp_res_i;
             // check the outcome of the branch speculation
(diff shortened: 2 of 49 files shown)


zarubaf:
Thank you very much for bringing this up and providing a fix. Can we change the pull request to move the global imports to the module body i.e., `module ariane import ariane_pkg::*; #(`. That should greatly reduce the diff and make it easier to merge :-D

Thanks!

whutddk:
I have moved the global imports to the module body already. But the name `is_branch` in `frontend.sv line 139` is collided with `ariane_pkg::is_branch`. As a result, I renamed `ariane_pkg::is_branch` to `ariane_pkg::isBranch`.

By the way, should I clean up the package prefixes of the logic? I think it will be easier to fix #477 and complete #485 after that, because the diff 
will greatly reduce.

include/ariane_pkg.sv
```diff
@@ -481,7 +481,7 @@ package ariane_pkg;
         logic [TRANS_ID_BITS-1:0] trans_id;
     } fu_data_t;
 
-    function automatic logic is_branch (input fu_op op);
+    function automatic logic isBranch (input fu_op op);
```
zarubaf, line 484: Can we rename the function name to `op_is_branch` (our style guide recommends function to be `snake_case`)?
```suggestion
    function automatic logic op_is_branch (input fu_op op);
```

src/branch_unit.sv
```diff
@@ -65,7 +65,7 @@ module branch_unit (
             resolved_branch_o.target_address = (branch_comp_res_i) ? target_address : next_pc;
             resolved_branch_o.is_taken = branch_comp_res_i;
             // check the outcome of the branch speculation
-            if (ariane_pkg::is_branch(fu_data_i.operator) && branch_comp_res_i != (branch_predict_i.cf == ariane_pkg::Branch)) begin
+            if (ariane_pkg::isBranch(fu_data_i.operator) && branch_comp_res_i != (branch_predict_i.cf == ariane_pkg::Branch)) begin
```
zarubaf, line 68:
```suggestion
            if (ariane_pkg::op_is_branch(fu_data_i.operator) && branch_comp_res_i != (branch_predict_i.cf == ariane_pkg::Branch)) begin
```

zarubaf (changes requested):
Thanks for contributing your time! Let's remove the commented import statements and I propose another function name ;-) Then we are good to merge.
Regarding the package prefix. Can we do this on a step-by-step basis so that the diff gets maintainable? (In a different PR)

whutddk:
Hi, I have resolved the issues in the review.

zarubaf (approved):
Terrific. LGTM

PR #491, opened 2020-08-05
PMP fixes

Moschn:
During some work with some other tools form the ASIC flow, some issues with the PMP syntax came up. This PR fixes the SV syntax so that hopefully more tools will be able to process it. The main problem was the over/underflow that happens for `pmp_conf_t[NrPMPEntries-1:0]` when `NrPMPEntries=0`.

Details:
- Always wire the full width of the PMP configuration signals through the arch
- The csr file now explicitly only sets the entries below `NrPMPEntries`
- Fix the bender file list

### DIFF
--- a/Bender.yml
+++ b/Bender.yml
@@ -60,6 +60,9 @@ sources:
     - src/mult.sv
     - src/load_unit.sv
     - src/issue_read_operands.sv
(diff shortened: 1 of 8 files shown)


zarubaf (approved):
Thanks, LGTM!

PR #498, opened 2020-08-14
FPGA Optimization for Performance Counters

ganoam:
This commit modifies the performance counter implementation to allow
inference of DSP slices to absorb counter logic on FPGas. By making use
of dedicated circuitry we are able to save both logic and flip-flop
resources. The optimization only works for counter widths up to 48 bit,
due to restrictions of the DSP slices.
The largest benefit has been observed for 32 bit wide counters.
In order to avoid premature overflows, the mcycle and minstret counters
are left untouched (64 bit).

Summary of changes:

- Counter logic for mhpmcounters is moved to a new separate module.
- The preprocessor variable TARGET_XILINX (automatically set by Bender)
  is used to include the required synthesis pragma and synchronous
  reset.
- DSP inference is supported for Xilinx FPGA devices featuring DSP48E1
  slices or similar and for counter widths up to 48 bits. The benefits
  are the largest for 32 bits wide counters.
- A new top-level parameter MHPMCounterWidth is introduced
  to control the width of the performance counters (excluding mcycle
  and minstret).

Concrete Savings: (Xilinx Kintex-7, xc7k325tffg900-2)

14 counters, 32 bit wide

```
            LUT   FF      DSP48E1
---------------------------------
baseline:   39874 22349   27
optimized:  38694 21922   41
--------------------------------
Diff        -1180 -427    +14
            -3.0% -1.9%
```

14 counters, 48 bit wide

```
            LUT   FF      DSP48E1
---------------------------------
baseline:   39841 22539   27
optimized:  39533 21905   41
--------------------------------
Diff        -308  -634    +14
            -0.8% -2.8%
```

Modified write adress decoder


src/perf_counters.sv
```diff
@@ -44,87 +46,68 @@ module perf_counters (
...
-            perf_counter_d[riscv::CSR_MCALL] = perf_counter_q[riscv::CSR_MCALL] + 1'b1;
+          // The standard software calling convention uses register x1 to hold the return address
+          // on a call.
+          // The unconditional jump is decoded as ADD op
+          perf_counter_inc[riscv::CSR_MCALL] =
+              (commit_instr_i[i].fu == CTRL_FLOW && commit_instr_i[i].op == '0
 
           // Return from call
-          if (commit_instr_i[i].op == JALR && (commit_instr_i[i].rd == 'd1 || commit_instr_i[i].rd == 'd1))
-            perf_counter_d[riscv::CSR_MRET] = perf_counter_q[riscv::CSR_MRET] + 1'b1;
```
ganoam, line 90: removed redundant 'or'

src/cva6_counter.sv
```diff
@@ -0,0 +1,70 @@
+module cva6_counter #(
```
ganoam, line 1: What do you want to name this module..?

zarubaf: That is fine with me. Can you add a license header, please?

src/perf_counters.sv
zarubaf: I think that is actually a bug and should be `x5` as this can be used as an alternative link register.

src/cva6_counter.sv
zarubaf: Can we please, instead of using the defines have two separate files:

1. One for RTL/ASIC synthesis
2. One optimized for FPGA

For the FPGA we'll then include the right file. There is a general interest in getting rid of as many defines as possible and introducing more seems like a step backwards ;-D

zarubaf (changes requested):
Thanks a lot for the proposed optimizations, that is really useful. There is only a minor comment on how to (not) handle the defines. Apart from that, it looks good to me.

zarubaf:
Ping @ganoam - I think this PR also needs a rebase :-)

zarubaf:
Closing due to inactivity.

Issue #470, opened 2020-07-08
Wrong exception tval for instruction fetch fault if instruction is crossing page boundary


xushengj:
When there is an instruction fetch causing page fault, according to riscv privileged instruction set document, mtval/stval "will ontain the virtual address of the portion of the instruction that caused the fault", but it seems that in the current implementation the PC is stored in the tval field. This causes hang when running some programs having a 32 bits instruction split across two pages; the fault appears to be in the first page while it should be on the second page.

The attachment is our current crafted binary that can trigger this hang in verilator and fpga running Linux, but successfully complete in QEMU.
[Test8.zip](https://github.com/openhwgroup/cva6/files/4893666/Test8.zip)

For now we can work around this issue by disabling c extension during compilation of user level programs, which force all instructions to be 4 bytes aligned. We will try to make a PR for this issue.

zarubaf:
Thanks, that absolutely makes sense. Let's discuss further on the PR if necessary.

PR #500, opened 2020-08-23
Fixed tval of misaligned 32-bit instr's page fault

xushengj:
Fixed wrong tval for instruction page fault exception when fetching the second half of a 4-byte instruction split across two pages.
This PR fixes issue #470

### DIFF
--- a/src/frontend/frontend.sv
+++ b/src/frontend/frontend.sv
@@ -418,6 +418,7 @@ module frontend import ariane_pkg::*; #(
       .instr_i             ( instr                ), // from re-aligner
       .addr_i              ( addr                 ), // from re-aligner
       .exception_i         ( icache_ex_valid_q    ), // from I$
+      .exception_addr_i    ( icache_vaddr_q       ),
       .predict_address_i   ( predict_address      ),
       .cf_type_i           ( cf_type              ),
       .valid_i             ( instruction_valid    ), // from re-aligner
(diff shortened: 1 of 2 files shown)


zarubaf (approved):
Thank you very much. That looks good to me. Any reason why the status of this PR is `draft`?

xushengj:
The main reason why this PR is marked as draft is that I'm struggling whether the assignment to fetch_entry_o.ex.tval should use sign extension or zero extension. The spec says that the tval should be the virtual address and I suppose a sign extension makes more sense. However, the current version use zero extension, which I'm not sure about the reason. I think a kernel module that can trigger the same issue is needed to test whether zero extension can work for faults on kernel code pages, and I haven't created the test module and done any test on that yet.

xushengj:
umm...  I'm not an expert in linux kernel and I'm not sure if the distinction of sign/zero extension matters for kernel modules. We may instead merge this PR first, and use another patch later on if this becomes an issue.

zarubaf:
I was under the impression that sign-extension is what we want. But looking at https://github.com/openhwgroup/cva6/blob/8678fedda2d7f49784cc59eaccc3466776a98bdf/src/load_store_unit.sv#L340 I can see that we also zero extend the `tval` here (which doesn't mean that it is correct). Either way I am okay with merging.

zarubaf:
Can you make sure to also rebase on the latest master and try to get the Eclipse contributor agreement check green? That requires your commits to be signed ``git commit -s` and a contributor agreement from the Eclipse foundation.

xushengj:
Yes. I have rebased the commit and signed the agreement.

zarubaf:
Terrific, thanks!

PR #502, opened 2020-08-24
Fix CI

zarubaf:


Jbalkind:
I don't like it much but we do use build-riscv-gcc.sh and install-verilator.sh from Ariane. Could you keep from deleting the files from now at least?

zarubaf:
Sure, thanks for letting me know!

PR #492, opened 2020-08-05
Separate NC and NI Load/Stores and fix NI bugs (final-squashed)

morenes:
PR that puts all commits into one, to make the changes more clear
https://github.com/openhwgroup/cva6/pull/431

PR #511, opened 2020-08-26
Movera ni

morenes:
The changes involve separating Non-Cacheable(NC) and Non-idempotent(NI) operations, cause Florian fix treat them as the same. The behavior is as follows:

Regarding Load operations:
--NC but Idempotent Loads would not need to wait until the WB is drained, and can be executed speculatively, as normal cacheable loads.
--NC and NI Loads would wait for the WB to be drained only if there are NI stores ongoing, and they cannot be issued speculatively.

Regarding Store operations:
--NC but Idempotent Stores could enter the WB in the same rules as a normal Cacheable Store.
--NC and NI Stores wouldn't enter the WB if there is another NC and NI Store inside it, but they could enter otherwise, without needing to halt other Cacheable operations to enter the WB, since they belong to different mem regions, they wouldnt forward (load) or coalesce (stores).

Note: NI operations are always NC too, but not viceversa.

New version of PR:
https://github.com/openhwgroup/cva6/pull/492


zarubaf:
Terrific works on my side, merging!

PR #512, opened 2020-08-27
clean up the package prefixes

whutddk:
I have cleaned up the package prefixes of the logic and use the global imports in the module body.

The function `ebreak()` in 'riscv::' is collided with the logic `ebreak` in `decorder.sv`, and `ebreak()`, `nop()`, `illegal()` can be found in package `dm`. So, I temporarily remove them in package `riscv`.


whutddk:
However, I am wondering if we are walking in the right direction. If we remove all package prefixes, the definition becomes short, ambiguous, and easy to collide with other definitions. But if we add all the package prefixes and avoid using global import in modules, the name will become too long especially after we add prefixes `cva6_`.

PR #519, opened 2020-09-04
tb_wb_dcache: small bugfixes

niwis:
- update signal names in wave.do
- fix local imports
- do not invalidate memory while there is an outgoing read reply


zarubaf (approved):
LGTM! Thanks!

PR #520, opened 2020-09-04
axi_intf.sv: import axi_pkg in AXI_LITE interface

niwis:
Fix artifact from recent change from global to local imports


zarubaf:
Thanks @niwis

PR #523, opened 2020-09-11
mmu: fix port width mismatch

niwis:
Fix width of `pmpaddr_i`. Note that PMP address registers only encode bits `2` - `PLEN-1` of the physical address.


Moschn:
I noticed this as well during my last PMP debug session. LGTM

zarubaf (approved):
LGTM

PR #529, opened 2020-09-23
Tentative fix for IFILL no speculate on NI addr

morenes:
Signed-off-by: Marcelo Orenes <movera@princeton.edu>

-Propagate speculative signal to icache, so the translated addr is not provided until the speculative flag goes low


morenes:
Closing this for the sake of #533 which is cleaner and it is signed!

PR #530, opened 2020-09-24
tb_wt_icache: fix icache testbench

niwis:
Minor fixes for the icache testbench to make it functional again:

- fix local imports
- adjust `CachedAddrEnd` to prevent an overflow of `CachedRegionLength`
- increase time padding between testcases
- update signal names in `wave.do`


zarubaf:
LGTM

PR #531, opened 2020-09-24
wt_icache: simplify FSM

niwis:
Remove redundant `TLB_MISS` state from icache FSM to
- reduce code size,
- decrease ITLB miss penalty by at least one cycle.


zarubaf (approved):
Nice, LGTM. I really like simplifications :-)

PR #533, opened 2020-09-25
icache fill speculative NI access fix

morenes:
-Propagate speculative signal to icache, so the translated addr is not provided until the speculative flag goes low
-Final version of #529


zarubaf (approved):
Thanks, that LGTM.

zarubaf:
I don't have access to your repository so I rebased the PR and opened it under #536. Thanks.

PR #524, opened 2020-09-14
Fix lint

JeanRochCoulon:
@zarubaf 

Sebastien Jack has detected some lint errors which break the FPGA generation in 32bits configuration. This PR fix some of them. 

Git logs:
load_store_unit.sv, load_unit.sv, mmu.sv, tlb.sv: fix lint warnings
issue_stage.sv: fix lint warnings on rs1_forwarding and rs2_forwarding signals
frontend.sv: fix lint error on icache_vaddr_q signal

A lint error related to PMP cannot be solved easily: PMP has been designed to be compliant to rv64, and not rv32. In 32bits, new registers (pmpconfig) need to be defined and current registers need to be reduced from 64 to 32bits. How can think we can fix these LINT errors?  Does Moritz Schneider available to adapt the PMP to 32bits, it would be great?

PR #537, opened 2020-10-02
Pmp 32 bit fixes

Moschn:
This PR fixes the incorrect length of PMP csrs and enables 32bit compatibility for `pmpcfg` (see https://github.com/openhwgroup/cva6/pull/524#issuecomment-691944602).


zarubaf (approved):
LGTM, actually makes sense that _physical_ memory protection should work on _physical memory_ addresses. Sorry for making this merge mess...

JeanRochCoulon:
Many thanks for this commit. The lint errors causes failure during FPGA synthesis, I will try with the PMP correction.
BTW, it have noticed that the PMP is always disable by boot. As no test is executed with PMP ON, this limits the code coverage and the PMP verification. Would it be possible to share PMP tests? It would be helpful to add these test to CVA6 regression suite. Cheers

PR #539, opened 2020-10-06
mmu: zero-extend paddr to match tval width

niwis:
`exception_t.tval` is `XLEN`, while `*_paddr` is only `PLEN` wide. This results in an overflow of `cause` into `tval` and wrong exception information. Therefore, zero-extend physical addresses when assigned to `lsu_exception_o.tval`.


zarubaf (approved):
Thanks, LGTM

Issue #538, opened 2020-10-05
PMP risc-v benchmark fails with page fault


Moschn:
The official riscv-tests have recently been updated to automatically detect the granularity of the underlying implementation and it would be nice to add them to our CI. However, the PMP benchmark fails with a page fault and AFAIK is unrelated to PMP (Note that this report is based on #537 and #523). If the `INLINE` keyword is removed from the function `test_one` the benchmark/test successfully completes. See patch below:

```patch
diff --git a/benchmarks/pmp/pmp.c b/benchmarks/pmp/pmp.c
index ec07e4a..4a7fbd4 100644
--- a/benchmarks/pmp/pmp.c
+++ b/benchmarks/pmp/pmp.c
@@ -17,8 +17,11 @@ uintptr_t handle_trap(uintptr_t cause, uintptr_t epc, uintptr_t regs[32])
   if (cause == CAUSE_ILLEGAL_INSTRUCTION)
     exit(0); // no PMP support

-  if (!trap_expected || cause != CAUSE_LOAD_ACCESS)
+  if (!trap_expected || cause != CAUSE_LOAD_ACCESS) {
+    printf("trap expected: %d\n", trap_expected);
+    printf("cause: %d, mtval: %p\n", cause, read_csr(mtval));
     exit(1);
+  }
   trap_expected = 0;
   return epc + insn_len(epc);
 }
@@ -97,6 +100,7 @@ INLINE int pmp_ok(pmpcfg_t p, uintptr_t addr, uintptr_t size)
   return hits == 0 || hits >= size;
 }

+// Remove INLINE here and it will work
 INLINE void test_one(uintptr_t addr, uintptr_t size)
 {
   uintptr_t new_mstatus = (read_csr(mstatus) & ~MSTATUS_MPP) | (MSTATUS_MPP & (MSTATUS_MPP >> 1)) | MSTATUS_MPRV;
```

The output from verilator:
```
This emulator compiled with JTAG Remote Bitbang client. To enable, use +jtag_rbb_enable=1.
Listening on port 41709
trap expected: 0
cause: 13, mtval: 0x80002dac
*** FAILED *** (tohost = 1)
../orig/riscv-tests/benchmarks/pmp.riscv *** FAILED *** (code = 1) after 370946 cycles
```

I could not figure out why it fails yet (and I am hoping someone with more experience takes over ;) ).

niwis:
The responsible code segment looks something like this:
```
[...]

PR #540, opened 2020-10-06
load unit: do not retire exception prematurely

niwis:
Fix for https://github.com/openhwgroup/cva6/issues/538:
Hold back exception from address translation when transmitting a valid, earlier load result. This is actually the behavior described in the comment of [line 282](https://github.com/niwis/ariane/blob/0249cfef00de2d6edb8e2dcb3b2fe84cab7a8ef3/src/load_unit.sv#L282):

> exceptions can retire out-of-order -> but we need to give priority to non-excepting load and stores so we simply check if we got an rvalid if so we prioritize it by not retiring the exception - we simply go for another round in the load FSM

### DIFF
--- a/src/load_unit.sv
+++ b/src/load_unit.sv
@@ -57,6 +57,16 @@ module load_unit import ariane_pkg::*; #(
         fu_op                     operator;
     } load_data_d, load_data_q, in_data;
 
+    // there is a D$ request pending while we may accept further load requests
+    logic dreq_pending_d, dreq_pending_q;
+    // store valid_i for one cycle
+    logic valid_d, valid_q;
+    // store lsu_ctrl_i.trans_id for one cycle
+    logic [TRANS_ID_BITS-1:0] trans_id_d, trans_id_q;
+
+    // save valid bit
+    assign valid_d = valid_i;
+    assign trans_id_d = lsu_ctrl_i.trans_id;
     // page offset is defined as the lower 12 bits, feed through for address checker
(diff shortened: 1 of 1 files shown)


niwis:
There were more thorough modifications required. From my understanding, the current implementation of the load unit makes two false assumptions:
1. **Same cycle translation request and exception:** The address translation exception arrives the same cycle as `translation_req_o` is set, and therefore `valid_i` and `lsu_ctrl_i.trans_id` are used to handle exceptions. This is incorrect, since the MMU needs exactly one cycle to return an exception. Therefore, `valid_i` and `lsu_ctrl_i.trans_id` from the previous cycle should be used, which is realized in this PR by storing both signals in a register.
2. **`req_port_i.data_rvalid & ex_i.valid` => do not commit `ex_i`** (somehow the summary of [line 282](https://github.com/niwis/ariane/blob/0249cfef00de2d6edb8e2dcb3b2fe84cab7a8ef3/src/load_unit.sv#L282)). The assumption here is that `req_port_i.data_rvalid` and `ex_i.valid` are responses to two different loads. However, this is doesn't have to be the case: e.g. if we have a cache hit, `req_port_i.data_rvalid` and `ex_i.valid` could arrive for the same load at the same time (after one cycle). Therefore, this PR checks for a pending D$ request instead.
Still needs to be verified on FPGA.

niwis:
Superseded by https://github.com/openhwgroup/cva6/pull/561.

PR #547, opened 2020-10-15
Makefile: support spike not located in RISCV directory

ASintzoff:
With this modification, spike can be located in any directory.
By default, it is still located in RISCV directory.


zarubaf (approved):
Great thanks! LGTM

PR #561, opened 2020-11-24
load_unit: fix exception forwarding

niwis:
When we receive a d-cache reply and an exception from the MMU in the same cycle, we need to check whether both belong to the same lsu request before forwarding the exception. This is the case iff `state_q == SEND_TAG`, since this is when we receive the translation result (i.e. physical tag / MMU exception) for the current in-flight d-cache request.

Otherwise, the exception belongs to the next request that is already being translated (e.g. `state_q == WAIT_GNT`). In this case of a collision, prioritize and return the valid load, discard the MMU exception, and go for another round in the load FSM.

Test bench passes and works on FPGA.


zarubaf:
Those are great news! Awesome!

zarubaf (approved):
LGTM. Thanks!

PR #573, opened 2020-12-15
Add RVFI to Ariane

JeanRochCoulon:
The RVFI has been implemented to generate execution trace. This RFVI implementation is synthetizable which is convenient to keep the trace functionility in gate simulation. Only the signals useful to generate instruction execution trace have been implemented. In the near future, our goal is to implement all the signals.

Makefile, rvfi_tracer_pck.sv, ariane.sv, ariane_testharness.sv :
    - rvfi_tracer.sv: add RVFI tracer and debug support. rvfi_tracer is a module added in ariane-testharness.sv.
    - RVFI ports are added to ariane module and connected to rvfi_tracer module
    - RVFI_TRACE enables RVFI trace generation

### DIFF
--- a/include/rvfi_tracer_pkg.sv
+++ b/include/rvfi_tracer_pkg.sv
@@ -1,24 +1,16 @@
-//#############################################################################
-//#
-//# Copyright 2020 Thales DIS design services SAS
-//#
-//# Licensed under the Solderpad Hardware Licence, Version 2.0 (the "License");
-//# you may not use this file except in compliance with the License.
-//# You may obtain a copy of the License at
-//#
-//#     https://solderpad.org/licenses/
-//#
-//# Unless required by applicable law or agreed to in writing, software
-//# distributed under the License is distributed on an "AS IS" BASIS,
-//# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-//# See the License for the specific language governing permissions and
-//# limitations under the License.
-//#
-//#############################################################################
-//#
-//# Original Author: Jean-Roch COULON (jean-roch.coulon@invia.fr)
-//#
-//#############################################################################
+// Copyright 2020 Thales DIS design services SAS
+//
+// Licensed under the Solderpad Hardware Licence, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// SPDX-License-Identifier: Apache-2.0 WITH SHL-2.0
+// You may obtain a copy of the License at https://solderpad.org/licenses/
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// Original Author: Jean-Roch COULON (jean-roch.coulon@invia.fr)
 
 package rvfi_tracer_pkg;
 
--- a/include/rvfi_tracer_pkg.sv
+++ b/include/rvfi_tracer_pkg.sv
@@ -1,0 +1,56 @@
+//#############################################################################
+//#
+//# Copyright 2020 Thales DIS design services SAS
+//#
+//# Licensed under the Solderpad Hardware Licence, Version 2.0 (the "License");
+//# you may not use this file except in compliance with the License.
+//# You may obtain a copy of the License at
+//#
+//#     https://solderpad.org/licenses/
+//#
+//# Unless required by applicable law or agreed to in writing, software
+//# distributed under the License is distributed on an "AS IS" BASIS,
+//# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//# See the License for the specific language governing permissions and
+//# limitations under the License.
+//#
+//#############################################################################
+//#
+//# Original Author: Jean-Roch COULON (jean-roch.coulon@invia.fr)
+//#
+//#############################################################################
+
+package rvfi_tracer_pkg;
+
+  localparam NRET = 1;
+  localparam ILEN = 32;
+
+  typedef struct packed {
+    logic [NRET-1:0]                 rvfi_valid;
+    logic [NRET*63:0]                rvfi_order;
+    logic [NRET*ILEN-1:0]            rvfi_insn;
+    logic [NRET-1:0]                 rvfi_trap;
+    logic [NRET-1:0]                 rvfi_halt;
+    logic [NRET-1:0]                 rvfi_intr;
+    logic [NRET*2-1:0]               rvfi_mode;
+    logic [NRET*2-1:0]               rvfi_ixl;
+    logic [NRET*5-1:0]               rvfi_rs1_addr;
+    logic [NRET*5-1:0]               rvfi_rs2_addr;
+    logic [NRET*riscv::XLEN-1:0]     rvfi_rs1_rdata;
+    logic [NRET*riscv::XLEN-1:0]     rvfi_rs2_rdata;
+    logic [NRET*5-1:0]               rvfi_rd_addr;
+    logic [NRET*riscv::XLEN-1:0]     rvfi_rd_wdata;
+
+    logic [NRET*riscv::XLEN-1:0]     rvfi_pc_rdata;
+    logic [NRET*riscv::XLEN-1:0]     rvfi_pc_wdata;
+
+    logic [NRET*riscv::XLEN-1:0]     rvfi_mem_addr;
+    logic [NRET*(riscv::XLEN/8)-1:0] rvfi_mem_rmask;
+    logic [NRET*(riscv::XLEN/8)-1:0] rvfi_mem_wmask;
+    logic [NRET*riscv::XLEN-1:0]     rvfi_mem_rdata;
+    logic [NRET*riscv::XLEN-1:0]     rvfi_mem_wdata;
+  } rvfi_instr_t;
+
+  typedef rvfi_instr_t [ariane_pkg::NR_COMMIT_PORTS-1:0] rvfi_port_t;
+
+endpackage
--- a/include/rvfi_tracer_pkg.sv
+++ b/include/rvfi_tracer_pkg.sv
@@ -4,11 +4,6 @@
 // you may not use this file except in compliance with the License.
 // SPDX-License-Identifier: Apache-2.0 WITH SHL-2.0
 // You may obtain a copy of the License at https://solderpad.org/licenses/
-// Unless required by applicable law or agreed to in writing, software
-// distributed under the License is distributed on an "AS IS" BASIS,
-// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-// See the License for the specific language governing permissions and
-// limitations under the License.
 //
 // Original Author: Jean-Roch COULON (jean-roch.coulon@invia.fr)
 
@@ -18,29 +13,29 @@ package rvfi_tracer_pkg;
   localparam ILEN = 32;
 
   typedef struct packed {
-    logic [NRET-1:0]                 rvfi_valid;
-    logic [NRET*63:0]                rvfi_order;
-    logic [NRET*ILEN-1:0]            rvfi_insn;
-    logic [NRET-1:0]                 rvfi_trap;
-    logic [NRET-1:0]                 rvfi_halt;
-    logic [NRET-1:0]                 rvfi_intr;
-    logic [NRET*2-1:0]               rvfi_mode;
-    logic [NRET*2-1:0]               rvfi_ixl;
-    logic [NRET*5-1:0]               rvfi_rs1_addr;
-    logic [NRET*5-1:0]               rvfi_rs2_addr;
-    logic [NRET*riscv::XLEN-1:0]     rvfi_rs1_rdata;
-    logic [NRET*riscv::XLEN-1:0]     rvfi_rs2_rdata;
-    logic [NRET*5-1:0]               rvfi_rd_addr;
-    logic [NRET*riscv::XLEN-1:0]     rvfi_rd_wdata;
+    logic [NRET-1:0]                 valid;
+    logic [NRET*63:0]                order;
+    logic [NRET*ILEN-1:0]            insn;
+    logic [NRET-1:0]                 trap;
+    logic [NRET-1:0]                 halt;
+    logic [NRET-1:0]                 intr;
+    logic [NRET*2-1:0]               mode;
+    logic [NRET*2-1:0]               ixl;
+    logic [NRET*5-1:0]               rs1_addr;
+    logic [NRET*5-1:0]               rs2_addr;
+    logic [NRET*riscv::XLEN-1:0]     rs1_rdata;
+    logic [NRET*riscv::XLEN-1:0]     rs2_rdata;
+    logic [NRET*5-1:0]               rd_addr;
+    logic [NRET*riscv::XLEN-1:0]     rd_wdata;
 
-    logic [NRET*riscv::XLEN-1:0]     rvfi_pc_rdata;
-    logic [NRET*riscv::XLEN-1:0]     rvfi_pc_wdata;
+    logic [NRET*riscv::XLEN-1:0]     pc_rdata;
+    logic [NRET*riscv::XLEN-1:0]     pc_wdata;
 
-    logic [NRET*riscv::XLEN-1:0]     rvfi_mem_addr;
-    logic [NRET*(riscv::XLEN/8)-1:0] rvfi_mem_rmask;
-    logic [NRET*(riscv::XLEN/8)-1:0] rvfi_mem_wmask;
-    logic [NRET*riscv::XLEN-1:0]     rvfi_mem_rdata;
-    logic [NRET*riscv::XLEN-1:0]     rvfi_mem_wdata;
+    logic [NRET*riscv::XLEN-1:0]     mem_addr;
+    logic [NRET*(riscv::XLEN/8)-1:0] mem_rmask;
+    logic [NRET*(riscv::XLEN/8)-1:0] mem_wmask;
+    logic [NRET*riscv::XLEN-1:0]     mem_rdata;
+    logic [NRET*riscv::XLEN-1:0]     mem_wdata;
   } rvfi_instr_t;
 
   typedef rvfi_instr_t [ariane_pkg::NR_COMMIT_PORTS-1:0] rvfi_port_t;
--- a/src/ariane.sv
+++ b/src/ariane.sv
@@ -42,7 +42,9 @@ module ariane import ariane_pkg::*; #(
   // firesim trace port
   output traced_instr_pkg::trace_port_t trace_o,
 `endif
-  // Risc-V Format Interface port
+  // RISC-V formal interface port (`rvfi`):
+  // Can be left open when formal tracing is not needed.
+  // The `rvfi_tracer` module can be connected to dump the trace content into a file.
   output rvfi_tracer_pkg::rvfi_port_t  rvfi_o,
 `ifdef PITON_ARIANE
   // L15 (memory side)
--- a/src/ariane.sv
+++ b/src/ariane.sv
@@ -42,6 +42,10 @@ module ariane import ariane_pkg::*; #(
   // firesim trace port
   output traced_instr_pkg::trace_port_t trace_o,
 `endif
+`ifdef RVFI_TRACE
+  // Risc-V Format Interface port
+  output rvfi_tracer_pkg::rvfi_port_t  rvfi_o,
+`endif
 `ifdef PITON_ARIANE
   // L15 (memory side)
   output wt_cache_pkg::l15_req_t       l15_req_o,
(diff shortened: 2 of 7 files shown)


rickoco:
@JeanRochCoulon pls log into your Eclipse account and be sure the select Thales as your company and also add your GitHubID to your Eclipse profile https://accounts.eclipse.org/ 

Doing so will resolve the eclipsefnd/eca errors...

tb/rvfi_tracer.sv
```diff
@@ -0,0 +1,88 @@
...
+  parameter int unsigned SIM_FINISH = 1000000,
+  parameter int unsigned DEBUG_START = 0,
+  parameter int unsigned DEBUG_STOP  = 0
+)(
+  input logic [31:0]                    cycles,
+  input rvfi_tracer_pkg::rvfi_port_t    rvfi_i
+);
+
+  int f;
+  initial f = $fopen("trace_ip.dasm", "w");
```
zarubaf, line 35: We somehow need to distinquish the trace file based on the `hartid`. Otherwise, we run into trouble when using CVA6 in a multi-core system.

tb/rvfi_tracer.sv
```diff
@@ -0,0 +1,88 @@
...
+    end
+    if (cycles > SIM_FINISH) $finish(1);
+  end
+
+  //Trace custom signals
+  //Define signals to be traced by adding them into debug and name arrays
+  string name[0:10];
+  logic[63:0] debug[0:10], debug_previous[0:10];
+  //assign debug[0] = i_ariane.csr_regfile_i.dscratch1_q;
+  //assign  name[0] = "i_ariane.csr_regfile_i.dscratch1_q";
```
zarubaf, line 78:
```suggestion
```

tb/rvfi_tracer.sv
```diff
@@ -0,0 +1,88 @@
...
+
+  always_ff @(posedge i_ariane.clk_i) begin
+    if (cycles > DEBUG_START && cycles < DEBUG_STOP)
+      for (int index=0; index<100; index=index+1)
```
zarubaf, line 82:
```suggestion
      for (int index = 0; index < 100; index++)
```

tb/rvfi_tracer.sv
```diff
@@ -0,0 +1,88 @@
...
+)(
+  input logic [31:0]                    cycles,
+  input rvfi_tracer_pkg::rvfi_port_t    rvfi_i
+);
+
+  int f;
+  initial f = $fopen("trace_ip.dasm", "w");
+  final $fclose(f);
+
+  //Generate the trace based on RVFI
```
zarubaf, line 38:
```suggestion
  // Generate the trace based on RVFI
```

tb/rvfi_tracer.sv
```diff
@@ -0,0 +1,88 @@
...
+  logic [63:0] pc64;
+  always_ff @(posedge i_ariane.clk_i) begin
+    for (int i = 0; i < NR_COMMIT_PORTS; i++) begin
+      pc64 = {{riscv::XLEN-riscv::VLEN{rvfi_i[i].rvfi_pc_rdata[riscv::VLEN-1]}}, rvfi_i[i].rvfi_pc_rdata};
+      //print instruction information if is valid or is trapped
```
zarubaf, line 43:
```suggestion
      // print the instruction information if the instruction is valid or a trap is taken
```

tb/rvfi_tracer.sv
```diff
@@ -0,0 +1,88 @@
...
+      if (rvfi_i[i].rvfi_valid) begin
+        //Instruction information
```
zarubaf, line 45:
```suggestion
        // Instruction information
```

tb/rvfi_tracer.sv
```diff
@@ -0,0 +1,88 @@
...
+        $fwrite(f, "core   0: 0x%h (0x%h) DASM(%h)\n",
+          pc64, rvfi_i[i].rvfi_insn, rvfi_i[i].rvfi_insn);
+        //Destination register information
```
zarubaf, line 48:
```suggestion
        // Destination register information
```

tb/rvfi_tracer.sv
```diff
@@ -0,0 +1,88 @@
...
+        $fwrite(f, "%h 0x%h (0x%h)",
+          rvfi_i[i].rvfi_mode, pc64, rvfi_i[i].rvfi_insn);
+        //Decode instruction to know if destination register is FP register
```
zarubaf, line 51:
```suggestion
        // Decode instruction to know if destination register is FP register
```

tb/rvfi_tracer.sv
```diff
@@ -0,0 +1,88 @@
...
+            rvfi_i[i].rvfi_rd_addr, rvfi_i[i].rvfi_rd_wdata);
+        else $fwrite(f, "\n");
+      end else if (rvfi_i[i].rvfi_trap)
+        $fwrite(f, "exception : 0x%h\n", pc64);
+    end
+    if (cycles > SIM_FINISH) $finish(1);
+  end
+
+  //Trace custom signals
+  //Define signals to be traced by adding them into debug and name arrays
```
zarubaf, line 74:
```suggestion
  // Trace any custom signals
  // Define signals to be traced by adding them into debug and name arrays
```

include/rvfi_tracer_pkg.sv
```diff
@@ -0,0 +1,56 @@
...
+//# distributed under the License is distributed on an "AS IS" BASIS,
+//# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//# See the License for the specific language governing permissions and
+//# limitations under the License.
+//#
+//#############################################################################
+//#
+//# Original Author: Jean-Roch COULON (jean-roch.coulon@invia.fr)
+//#
+//#############################################################################
```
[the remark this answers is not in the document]
zarubaf: @JeanRochCoulon Might be a style question, but I would appreciate if we could keep the license headers approximately the same style i.e., without all the hashes `#`. Not a hard point for me if you happen to like them very much :-)

zarubaf (commented):
LGTM, a couple of small nits and there is a small license discussion I just want to get a second opinion (i think it is alright!). Thanks for the great contribution!

include/rvfi_tracer_pkg.sv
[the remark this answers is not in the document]
MikeOpenHWGroup: @JeanRochCoulon, @zarubaf, it is OK to mix and match Solderpad 0.51 and 2.0, although for new submissions, it would be better to use 2.0.

Also, please add an SPDX header like this:
```
// SPDX-License-Identifier: Apache-2.0 WITH SHL-2.0
```

Sorry to provide this information to you so late @JeanRochCoulon, we are just learning about this recently ourselves.

MikeOpenHWGroup:
Hi @JeanRochCoulon and @zarubaf, I have one very big comment to make.   In CV32E40P we worked very hard to remove compiler directives from the RTL (`ifdef`, etc.).  The tracer for CV32E40P is therefore not part of the RTL of the core at all.  It is bound into the RTL using a [wrapper](https://github.com/openhwgroup/cv32e40p/blob/master/bhv/cv32e40p_wrapper.sv) that instantiated the tracer, core and other things as well.   I strongly recommend we do this for the CVA6 as well.

Yes, I know this is a lot of work.  You can ask @Silabs-ArjanB and @davideschiavone about it.  Even with significant effort, there are still some in the code (I hope to have them all gone in cv32e40p_v2).

include/rvfi_tracer_pkg.sv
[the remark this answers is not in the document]
JeanRochCoulon: The license description of the two following files trace_rvfi_pkg.sv, trace_rvfi.sv have been updated

src/ariane.sv
```diff
@@ -894,4 +898,37 @@ module ariane import ariane_pkg::*; #(
 `endif // VERILATOR
 //pragma translate_on
 
+`ifdef RVFI_TRACE
+  logic trap, exception;
+  always @(commit_ack, commit_instr_id_commit, debug_mode, wdata_commit_id, priv_lvl, ex_commit)
```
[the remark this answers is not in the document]
JeanRochCoulon: ok, the always has been updated. As I used to write code in VHDL, I was not aware of this always_comb possibility!

src/ariane.sv
```diff
@@ -894,4 +898,37 @@ module ariane import ariane_pkg::*; #(
...
+    for (int i = 0; i < NR_COMMIT_PORTS; i++) begin
+      trap = commit_instr_id_commit[i].valid && commit_instr_id_commit[i].ex.valid && ex_commit.valid;
+      exception = trap &&
```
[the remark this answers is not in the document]
JeanRochCoulon: The RVFI describes the rfvi_trap signal generation as it: 
"_rvfi_trap must be set for an instruction that cannot be decoded as a legal instruction, such as 0x00000000.
In addition, rvfi_trap must be set for a misaligned memory read or write in PMAs that don't allow misaligned access, or other memory access violations. rvfi_trap must also be set for a jump instruction that jumps to a misaligned instruction._"
That's why exception signal does not correspond to all exceptions, but only exception which are linked to the memory. To avoid confusion, I have renamed _trap_ into _exception_ and _exception_ into _mem_exception_. I think it is more in line with RVFI description. Your feedback is welcome.

src/ariane.sv
```diff
@@ -894,4 +898,37 @@ module ariane import ariane_pkg::*; #(
...
+         commit_instr_id_commit[i].ex.cause == riscv::ST_ADDR_MISALIGNED ||
+         commit_instr_id_commit[i].ex.cause == riscv::ST_ACCESS_FAULT ||
+         commit_instr_id_commit[i].ex.cause == riscv::INSTR_PAGE_FAULT ||
+         commit_instr_id_commit[i].ex.cause == riscv::LOAD_PAGE_FAULT ||
+         commit_instr_id_commit[i].ex.cause == riscv::STORE_PAGE_FAULT);
+
+      rvfi_o[i].rvfi_valid    = (commit_ack[i] && !commit_instr_id_commit[i].ex.valid) ||
+        (trap && (commit_instr_id_commit[i].ex.cause == riscv::ENV_CALL_MMODE ||
+                  commit_instr_id_commit[i].ex.cause == riscv::ENV_CALL_SMODE ||
+                  commit_instr_id_commit[i].ex.cause == riscv::ENV_CALL_UMODE));
```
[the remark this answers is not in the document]
JeanRochCoulon: Trap are of two different types:
- In case of memory trap (rvfi_trap), the instruction is not traced because it is not executed. But the exception is reported.
- In case of CALL trap, the instruction is traced (it is executed) because it is executed. That's why they are not contributing to rvfi_trap but rvfi_valid.

src/ariane.sv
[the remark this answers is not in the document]
JeanRochCoulon: mem_exception correspond to instructions which are not executed. While CALL exception corresponds to instructions which are executed

JeanRochCoulon:
@MikeOpenHWGroup @zarubaf  removing the compiler directive from RTL would break the execution trace at gate level. Indeed, in CV32_wrapper, signal values are probed by providing the hierarchy path, for instance core_i.id_stage_i.controller_i.ctrl_fsm_cs. Also if it works at RTL level, it does not at all at gate level because the hierarchy path does not exist anymore. Unfortunately I do not see solution to remove the compiler directive and to have the execution trace functional in RTL and gate. I would be pleased to get your opinion.

JeanRochCoulon:
@rickoco @zarubaf I have solved the ECA errors, thank you for the feedbacks

zarubaf:
> Hi @JeanRochCoulon and @zarubaf, I have one very big comment to make. In CV32E40P we worked very hard to remove compiler directives from the RTL (ifdef, etc.). The tracer for CV32E40P is therefore not part of the RTL of the core at all. It is bound into the RTL using a wrapper that instantiated the tracer, core and other things as well. I strongly recommend we do this for the CVA6 as well.

That is what is happening here as well - or did I miss something? The goal is to have a synthesisable tracer so we can't use dot references and need to somehow export the necessary signals. The only subject to discussion would be whether we want to unconditionally enable the tracer port in the top-level. Is that what you are referring to?

include/rvfi_tracer_pkg.sv
```diff
@@ -0,0 +1,48 @@
...
+//
+// Licensed under the Solderpad Hardware Licence, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// SPDX-License-Identifier: Apache-2.0 WITH SHL-2.0
+// You may obtain a copy of the License at https://solderpad.org/licenses/
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
```
zarubaf, line 11:
```suggestion
```
My understanding was that you wouldn't necessarily need this when having the pointer to the SPDX Identifier and a link to the license :-)

zarubaf: Looks much better by the way. Thanks!

tb/rvfi_tracer.sv
```diff
@@ -0,0 +1,78 @@
...
+//
+// Licensed under the Solderpad Hardware Licence, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// SPDX-License-Identifier: Apache-2.0 WITH SHL-2.0
+// You may obtain a copy of the License at https://solderpad.org/licenses/
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
```
zarubaf, line 11:
```suggestion
```

include/rvfi_tracer_pkg.sv
```diff
@@ -0,0 +1,48 @@
...
+    logic [NRET*riscv::XLEN-1:0]     rvfi_rd_wdata;
+
+    logic [NRET*riscv::XLEN-1:0]     rvfi_pc_rdata;
+    logic [NRET*riscv::XLEN-1:0]     rvfi_pc_wdata;
+
+    logic [NRET*riscv::XLEN-1:0]     rvfi_mem_addr;
+    logic [NRET*(riscv::XLEN/8)-1:0] rvfi_mem_rmask;
+    logic [NRET*(riscv::XLEN/8)-1:0] rvfi_mem_wmask;
+    logic [NRET*riscv::XLEN-1:0]     rvfi_mem_rdata;
+    logic [NRET*riscv::XLEN-1:0]     rvfi_mem_wdata;
```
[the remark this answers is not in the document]
zarubaf: I would say that the `rvfi_` in the struct is redundant as it is already part of the struct's name. I would prefer something more concise, do you agree?

src/ariane.sv
```diff
@@ -894,4 +898,38 @@ module ariane import ariane_pkg::*; #(
...
+      rvfi_o[i].rvfi_insn     = commit_instr_id_commit[i].ex.tval[31:0];
+      // when rvfi_trap, the instruction is not executed
+      rvfi_o[i].rvfi_trap     = mem_exception;
+      rvfi_o[i].rvfi_mode     = debug_mode ? 2'b10 : priv_lvl;
+      rvfi_o[i].rvfi_ixl      = riscv::XLEN == 64 ? 2 : 1;
+      rvfi_o[i].rvfi_rs1_addr = commit_instr_id_commit[i].rs1;
+      rvfi_o[i].rvfi_rs2_addr = commit_instr_id_commit[i].rs2;
+      rvfi_o[i].rvfi_rd_addr  = commit_instr_id_commit[i].rd;
+      rvfi_o[i].rvfi_rd_wdata = ariane_pkg::is_rd_fpr(commit_instr_id_commit[i].op) == 0 ? wdata_commit_id[i] : commit_instr_id_commit[i].result;
+      rvfi_o[i].rvfi_pc_rdata = commit_instr_id_commit[i].pc;
```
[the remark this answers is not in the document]
zarubaf: Necessary change if we go for the reduced names. Here the name duplication becomes quite obvious :-)

zarubaf (commented):
A couple of more nits. Let's wait for @MikeOpenHWGroup comment on how to resolve his comment. Also, let's try to synthesize the FPGA image, just to sanity check. Otherwise LGTM.

JeanRochCoulon:
> 
> 
> > Hi @JeanRochCoulon and @zarubaf, I have one very big comment to make. In CV32E40P we worked very hard to remove compiler directives from the RTL (ifdef, etc.). The tracer for CV32E40P is therefore not part of the RTL of the core at all. It is bound into the RTL using a wrapper that instantiated the tracer, core and other things as well. I strongly recommend we do this for the CVA6 as well.
> 
> That is what is happening here as well - or did I miss something? The goal is to have a synthesisable tracer so we can't use dot references and need to somehow export the necessary signals. The only subject to discussion would be whether we want to unconditionally enable the tracer port in the top-level. Is that what you are referring to?

Florian, your reasoning is correct. 
If the scope is to prevent from adding compiler directive, I would see two solutions:
- The rvfi_tracer is instantiated inside Ariane, $fwrite would be encapsulated with translate_on and off commands. In that case, RVFI functionality is added without defining new Ariane ports.
- An unconditional tracer port is added to Ariane module and the rvfi_tracer module is out of Ariane.
To my mind the second is better, because less logic gate dedicated to tracer remains in Ariane and all incremental code needed to implement the linked functionalities (step end compare, formal verification,... ) could be added outside Ariane, in testbench.
I do not see other straight forward solution, but maybe @MikeOpenHWGroup or @zarubaf would you think about something else?

tb/rvfi_tracer.sv
zarubaf: @JeanRochCoulon Here we just need to pass down (either through a port or parameter) a unique id.

zarubaf:
> > > Hi @JeanRochCoulon and @zarubaf, I have one very big comment to make. In CV32E40P we worked very hard to remove compiler directives from the RTL (ifdef, etc.). The tracer for CV32E40P is therefore not part of the RTL of the core at all. It is bound into the RTL using a wrapper that instantiated the tracer, core and other things as well. I strongly recommend we do this for the CVA6 as well.
> > 
> > 
> > That is what is happening here as well - or did I miss something? The goal is to have a synthesisable tracer so we can't use dot references and need to somehow export the necessary signals. The only subject to discussion would be whether we want to unconditionally enable the tracer port in the top-level. Is that what you are referring to?
> 
> Florian, your reasoning is correct.
> If the scope is to prevent from adding compiler directive, I would see two solutions:
> 
> * The rvfi_tracer is instantiated inside Ariane, $fwrite would be encapsulated with translate_on and off commands. In that case, RVFI functionality is added without defining new Ariane ports.
> * An unconditional tracer port is added to Ariane module and the rvfi_tracer module is out of Ariane.
> 
> To my mind the second is better, because less logic gate dedicated to tracer remains in Ariane and all incremental code needed to implement the linked functionalities (step end compare, formal verification,... ) could be added outside Ariane, in testbench. I have added a commit to implement this solution.
> I do not see other straight forward solution, but maybe @MikeOpenHWGroup or @zarubaf would you think about something else?

Agree, but I am pretty sure that the FPGA build will fail now. We need to add the `rvfi` port here: https://github.com/openhwgroup/cva6/blob/6fd1a734e0a79a28b3965cb02a4df5ea399b39ab/fpga/src/ariane_xilinx.sv#L424

@JeanRochCoulon are you able to run FPGA synthesis on your end? I would also suggest adding a comment to `ariane.sv` where the tracer port is located on how to instantiate it (for example if unused can be left open).

src/ariane.sv
```diff
@@ -42,6 +42,8 @@ module ariane import ariane_pkg::*; #(
   // firesim trace port
   output traced_instr_pkg::trace_port_t trace_o,
 `endif
+  // Risc-V Format Interface port
```
zarubaf, line 45: Something along those lines:
```suggestion
  // RISC-V formal Interface port (`rvfi`):
  // Can be left open when formal tracing is not needed.
  // The `rvfi_tracer` module can be connected to dump the trace content into a file.
```

JeanRochCoulon:
> 
> 
> > * The rvfi_tracer is instantiated inside Ariane, $fwrite would be encapsulated with translate_on and off commands. In that case, RVFI functionality is added without defining new Ariane ports.
> > * An unconditional tracer port is added to Ariane module and the rvfi_tracer module is out of Ariane.
> > 
> > To my mind the second is better, because less logic gate dedicated to tracer remains in Ariane and all incremental code needed to implement the linked functionalities (step end compare, formal verification,... ) could be added outside Ariane, in testbench. I have added a commit to implement this solution.
> > I do not see other straight forward solution, but maybe @MikeOpenHWGroup or @zarubaf would you think about something else?
> 
> Agree, but I am pretty sure that the FPGA build will fail now. We need to add the `rvfi` port here:
> 
> https://github.com/openhwgroup/cva6/blob/6fd1a734e0a79a28b3965cb02a4df5ea399b39ab/fpga/src/ariane_xilinx.sv#L424-L437
> 
> @JeanRochCoulon are you able to run FPGA synthesis on your end? I would also suggest adding a comment to `ariane.sv` where the tracer port is located on how to instantiate it (for example if unused can be left open).

Yes, you are right, the FPGA build will fail. On my side, th eFPGA build is planned but not computed yet. Unfortunately I cannot help on this point.

tb/rvfi_tracer.sv
JeanRochCoulon: I have added the hart_is input inside rvfi_tracer, output trace file has been renamed to contain the hart_id. Thanks for the remark, it is better now.

JeanRochCoulon:
Hello @zarubaf, @MikeOpenHWGroup, 
25 days since the last comment... Do you have other feedback? or could you pull it on the main line. I am working on gate simulation (and using the RVFI), and I would pleased to stack my new modification on granted database. Thanks

zarubaf:
Sorry, that fell through the Christmas cracks. The only remaining problem is the FPGA buid, we need to solve this before I can pull this PR in.

JeanRochCoulon:
If I can help?   Please tell me which command I need to execute to see if the FPGA compilation is ok, we already know the top module need to be modified.

zarubaf:
It should really just be `make fpga` if you have a reasonable new Vivado version installed. Let me know if that doesn’t work for you!

PR #574, opened 2020-12-26
Fixed multithreading and optimization options for verilator target

xushengj:
This PR primarily fixed the following problems in the verilator target:
1.  The verilator version test did not work correctly on our
    System (Ubuntu 18.04). The verilator model was always
    single-threaded before the change.
2.  From our experiments, 4 threads is a better pick than 2 threads.
    On our workstation with i7-7700, running verilator model with
    4 threads use about 60% of wall time of the case in 2 threads, and
    more threads are having diminishing returns. On our server,
    8 threads are slower than 4 threads.
3.  The original Makefile did not add optimization flags to verilator
    output. Adding the optimization flag makes building the verilator
    target significantly slower (>=10 mins) but makes the simulation
    significantly faster.
4.  The VL_DEBUG macro is added to the CFLAGS of the verilator target.
    It can assist catching buggy logic at runtime. IRCC we added
    VL_DEBUG to catch a combinational loop we acccidentally introduced
    in our modification to the processor.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -77,27 +77,27 @@ jobs:
       name: run riscv benchmarks (Write-Back Cache)
       script:
         - ci/build-riscv-tests.sh
-        - make -j${NUM_JOBS} run-benchmarks-verilator defines=WB_DCACHE
+        - make run-benchmarks-verilator defines=WB_DCACHE
 
     # asm tests
     - stage: test
       name: run asm tests (Write-Back Cache)
       script:
         - ci/build-riscv-tests.sh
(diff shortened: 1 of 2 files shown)


zarubaf:
It is a bit strange that the CI tests do not pass anymore. Can you confirm that behavior locally?

xushengj:
Sorry for the late reply. No, we cannot reproduce that. For some reason, after I merge the changes from the upstream, all the CI tests now pass. No idea why...

zarubaf:
Great thanks! LGTM!

PR #584, opened 2021-01-18
According to IEEE 1800-2017 (9.2.2.4 Sequential logic always_ff proce…

pawelkudlakaldec:
…dure)

"Variables on the left-hand side of assignments within an always_ff procedure, including variables from the contents of a called function, shall not be written to by any other process."
Thus there is a proposal of changing always_ff with always because Mem_DP is driven by two processes: initial and always_ff what is forbidden.

### DIFF
--- a/src/dromajo_ram.sv
+++ b/src/dromajo_ram.sv
@@ -71,7 +71,7 @@ module dromajo_ram
     end
   end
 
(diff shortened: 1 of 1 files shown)


zarubaf (approved):
LGTM.

PR #593, opened 2021-01-24
instr_queue: Prevent latch inference in always_comb

LucaZulberti:
Using CVA6 as an IP in Vivado 2018.2, a latch was inferred durring OOC synthesis.

Signed-off-by: Luca Zulberti <zulberti.luca@gmail.com>


zarubaf (approved):
Awesome, thanks!

PR #595, opened 2021-01-26
axi_adapter/shim: no need to have the fixed burst type

LucaZulberti:
The AXI transactions can be _regular_ as stated in AXI specification section A3.4.3.

It seems there is no need to perform FIXED transactions, am I wrong?
This patch makes the CPU compliant also with _AXI Regular Transactions_. For example, now it is compatible with the AXI SmartConnect IP of Vivado.

I have created a draft PR since I have not tested this patch a lot. Do you have any concerns about this change?

### DIFF
--- a/src/axi_adapter.sv
+++ b/src/axi_adapter.sv
@@ -17,332 +17,332 @@
 //import std_cache_pkg::*;
 
 module axi_adapter #(
-    parameter int unsigned DATA_WIDTH            = 256,
-    parameter logic        CRITICAL_WORD_FIRST   = 0, // the AXI subsystem needs to support wrapping reads for this feature
-    parameter int unsigned AXI_ID_WIDTH          = 10,
-    parameter int unsigned CACHELINE_BYTE_OFFSET = 8
+  parameter int unsigned DATA_WIDTH            = 256,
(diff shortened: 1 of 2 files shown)


zarubaf:
You are definitely right and the patch seems okay to me. In fact, I think this makes Ariane compliant with much more of our AXI infrastructure. 

Just a small nit you might want to use the constants from the [`axi_pkg`](https://github.com/pulp-platform/axi/blob/3f5d5b540a6403c5c611e507a14e99d9e7f22edf/src/axi_pkg.sv#L28), just to make the design intent a bit clearer (and shame on me who hard-coded this in the first place :-))

LucaZulberti:
Perfect! I will update the PR this evening or tomorrow. ;-)

LucaZulberti:
Now it should be ok. I did some cleaning, if you like, I can convert the PR.

zarubaf (approved):
Thanks a lot. LGTM!

zarubaf:
> Now it should be ok. I did some cleaning, if you like, I can convert the PR.

Please go ahead and convert the PR.

LucaZulberti:
Ready ;-) 
(next time I will convert a PR as soon as it's ready for review)

PR #598, opened 2021-01-29
Remove unnecessary ariane_soc_pkg dependency

LucaZulberti:
With my work, I had the necessity to remove the dependency of ariane_soc package from the CVA6 core.
It was straightforward because *_slv_t types are needed only in ariane_testharness.

I'm not familiar with your build system (Bender, etc..) and I'm not sure this is the best solution to be maintained.
If you are interested in isolating better the code of CVA6 from the SoC where it is instantiated, this patch aims to obtain this result.

Draft until interested and improved.

### DIFF
--- a/Bender.yml
+++ b/Bender.yml
@@ -23,7 +23,6 @@ sources:
     - src/register_interface/src/reg_intf.sv
     - src/register_interface/src/reg_intf_pkg.sv
     - include/axi_intf.sv
-    - tb/ariane_soc_pkg.sv
     - include/ariane_axi_pkg.sv
     - src/fpu/src/fpnew_pkg.sv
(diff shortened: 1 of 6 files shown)


zarubaf:
Thanks, @LucaZulberti, that LGTM. 

@MikeOpenHWGroup Can you confirm that this solves the dependency issue? If so, let's bring this in before #602.

MikeOpenHWGroup:
I can confirm that this resolves, the dependency issue, so yes, please merge this in ahead of #602.

Thanks @LucaZulberti.

zarubaf (approved):
Thanks @LucaZulberti, that looks fantastic to me. I've also tried building the FPGA which works. So I am going ahead with merging.

PR #600, opened 2021-01-30
cache_ctrl: remove forward reference

LucaZulberti:
DesignCompiler 2019.12 reports it as a possible error in future versions.


zarubaf (approved):
Thanks for fixing this, LGTM!

Issue #451, opened 2020-06-20
Request for contribution


NazerkeT:
Dear Florian,
Congratulations with passing Ariane to Open HW Group!
Wish Ariane all the best!

I am motivated by the idea behind open source development and PULP project, so want to contribute to Ariane as a summer project. I want to start my journey by adding support for Bit Manipulation Extension. Nevertheless, first, I want to understand whether is it really necessary and useful for Ariane, or is it something extra that you don't want to consider for now? If B extension is something wanted, then is it possible to make some plan on contribution as for A extension support to make things clear, or could I be addressed to someone interested in this?) 

On the other hand, if B extension is not necessary for the present, then I am open to any new idea on contributions as a project. For example, is adding support for out of order executions interesting for Ariane? Or is there any other extension support or some architecture improvements that were planned/desired?

P.S. Basically, I have written an email to you too, but I am in worries whether is it got into spam or not. I am sorry if I am writing too much)

zarubaf:
Hi @NazerkeT,

thanks for your interest in contributing to Ariane and the hardware ecosystem in general. I think the B extension would make an excellent contribution to Ariane/CVA6 and to RISC-V. 

Another interesting feature could be an optional L2 TLB or having an optional SV48 address translation mode.

Let me know what you think and where you would be excited to contribute?

Cheers,
Florian

NazerkeT:
Hi @zarubaf,
All the options are exciting and appealing to me :). However, as long as I have already started to study RISCV Bit Manipulation, I think I better, first, go with adding support for the B extension. For now, I have completed one cycle implementation for `pcnt, clz, ctz`(some debug is needed for `ctz`). I've implemented them as functions, as I was initially planning for possible reusability of above-mentioned instructions. However, I rarely saw function implementations in ALU designs (after searching over some RISCV core repos including Ariane). Should I avoid implementation using functions? 

One more question, before processing to other instructions and considering possible architecture changes, could I be shared with what instructions are desired to be implemented in Ariane? And are there any other requirements on extension support to follow as of specification/technical task? 

Best regards,

NazerkeT:
Hi @zarubaf !
[...]

PR #602, opened 2021-01-31
Stand-alone manifest for CVA6

MikeOpenHWGroup:
Please do not merge this (yet).

The primary purpose of this pull-request is to review `cva6_manifest.flist`, a stand-alone manifest for the CVA6 core (without any of the external TB or FPGA components).  This manifest is currently used in core-v-verif (see [core-v-verif #451](https://github.com/openhwgroup/core-v-verif/pull/451)), so I am at least 90% confident that is it correct and complete.

Ultimately this manifest should be in a better form such as a YAML structure, but let's focus on content first.  I have three specific comments/questions:
1. Is the list of files in the manifest complete?
2. The RTL should not be dependent on a file in the tb directory.  So `tb/ariane_soc_pkg.sv` will need to find another home, or the RTL-specific content could be extracted and moved to a file in another directory.
3. The core RTL should not be dependent on an FPGA-specific module.  This affects `src/fpga-support/rtl/SyncSpRamBeNx64.sv`.

The other changes in this pull-request exist to support Metrics dsim.  I am still working through some of this and will open specific issues and/or pull-requests on this later.


LucaZulberti:
Hi @MikeOpenHWGroup, is the solution given by #598 a good starting point for your second comment?

tb/ariane_tb.cpp
```diff
@@ -19,7 +19,7 @@
 #include "verilator.h"
 #include "verilated.h"
 #include "verilated_vcd_c.h"
-#include "Variane_testharness__Dpi.h"
+//#include "Variane_testharness__Dpi.h"
```
zarubaf, line 22: Why is this needed?

zarubaf (commented):
> Ultimately this manifest should be in a better form such as a YAML structure, but let's focus on content first. 
Yes, please - maybe we can also use something standardized - this is effectively manifest number 5 in this repo...
> I have three specific comments/questions:
> 
> 1. Is the list of files in the manifest complete?
Looks okay for the write-through cache version. Of course, the write-back cache needs different files.
> 2. The RTL should not be dependent on a file in the tb directory.  So `tb/ariane_soc_pkg.sv` will need to find another home, or the RTL-specific content could be extracted and moved to a file in another directory.
See #598.
> 3. The core RTL should not be dependent on an FPGA-specific module.  This affects `src/fpga-support/rtl/SyncSpRamBeNx64.sv`.
It is a behavioral description of an SRAM, just a bit tuned so that the Xilinx and Altera tools can infer BRAMs for it. The `fpga-support` repo is anyway deprecated afaik, so we can flatten and re-name if that helps the situation. Don't think this is really high-priority, tbh.

tb/ariane_tb.cpp
MikeOpenHWGroup: Opps.  Good catch, thanks.  That should not have been pushed in.  Restored.

MikeOpenHWGroup:
> Ultimately this manifest should be in a better form such as a YAML structure, but let's focus on content first.
>> Yes, please - maybe we can also use something standardized - this is effectively manifest number 5 in this repo...

Completely agree.  Standardizing the manifests is a priority for me.  I feel bad for adding manifest number 5 and promise to eliminate it as soon as possible.

> 3. The core RTL should not be dependent on an FPGA-specific module. This affects src/fpga-support/rtl/SyncSpRamBeNx64.sv.
>> It is a behavioral description of an SRAM, just a bit tuned so that the Xilinx and Altera tools can infer BRAMs for it. The fpga-support repo is anyway deprecated afaik, so we can flatten and re-name if that helps the situation. Don't think this is really high-priority, tbh.

No time like the present.  :wink: Either we fix it now, or I will open another specific issue for it.

PR #607, opened 2021-02-04
Flush all state when reset is not deasserted

berendo:
This PR ensures that while the CPU reset is asserted, the simulation instruction tracer flushes its internal state, so that it can resume tracing correctly once reset is deasserted. Otherwise the trace logging becomes inconsistent with the actual CPU state if the CPU is ever reset externally and then resumes execution.

### DIFF
--- a/src/util/instr_tracer.sv
+++ b/src/util/instr_tracer.sv
@@ -66,7 +66,11 @@ module instr_tracer (
     forever begin
       automatic ariane_pkg::bp_resolve_t bp_instruction = '0;
       // new cycle, we are only interested if reset is de-asserted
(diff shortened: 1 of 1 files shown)


zarubaf (approved):
Great, thanks! LGTM

zarubaf:
Can you briefly rebase on the latest master please? I've fixed the failing CI :-)

berendo:
@zarubaf done! Thank you for fixing the CI failure - I had noticed that a bunch of other PRs were failing and figured it was something that was being worked on. I'll keep an eye on the current CI run. 🤞

PR #612, opened 2021-02-16
Added fixes for Errors/Warnings for compiling with xcelium on CVA6

OttG:
**List of Error Fixed for xrun command:**
* In `ariane_pkg.sv` and `riscv_pkg.sv` the slicing of an enumeration constant is illegal for xcelium. The size of `MODE_SV` has been fixed to the size of `vm_mode_t` which removes the need for slicing;
* In `pmp.sv` assigning the port `conf_addr_prev_i ` of `pmp_entry` with a ternary operator caused an error. The assignment has been done outside the `pmp_entry` module.

**Warnings:**
* In `ariane_pkg.sv` removed extra semicolons at the end of endfunctions.

### DIFF
--- a/include/riscv_pkg.sv
+++ b/include/riscv_pkg.sv
@@ -39,8 +39,8 @@ package riscv;
     localparam ModeW      = (XLEN == 32) ? 1 : 4;
     localparam ASIDW      = (XLEN == 32) ? 9 : 16;
     localparam PPNW       = (XLEN == 32) ? 22 : 44;
-    localparam logic [ModeW-1:0] MODE_SV = (XLEN == 32) ? ModeSv32[ModeW-1:0] : ModeSv39[ModeW-1:0];
-    localparam SV         = (MODE_SV == ModeSv32[ModeW-1:0]) ? 32 : 39;
+    localparam logic [3:0] MODE_SV = (XLEN == 32) ? ModeSv32 : ModeSv39;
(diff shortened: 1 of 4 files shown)


include/riscv_pkg.sv
```diff
@@ -39,8 +39,8 @@ package riscv;
     localparam ModeW      = (XLEN == 32) ? 1 : 4;
```
zarubaf, line 39: Is `ModeW` used anymore? Otherwise, we can get rid of it imho. Or at least we can tie it to `4` for all cases.

src/pmp/src/pmp.sv
```diff
@@ -32,13 +32,22 @@ module pmp #(
...
 
         for (genvar i = 0; i < NR_ENTRIES; i++) begin
+
+            logic [PMP_LEN-1:0] conf_addr_prev;
+	
+	          if(i==0)begin
+            	assign conf_addr_prev = '0;
+ 	          end else begin
+	            assign conf_addr_prev = conf_addr_i[i-1];
+	          end
```
zarubaf, line 42: Nitpick: That should also work, right? A bit more concise imho.
```suggestion
	          assign conf_addr_prev = (i == 0) ? '0 : conf_addr_i[i-1]
```

zarubaf (changes requested):
Awesome, LGTM in general. Only two small nits.

PR #614, opened 2021-02-17
Added fixes for Errors/Warnings for compiling with xcelium on CVA6

OttG:
**List of Error Fixed for xrun command:**
* In `ariane_pkg.sv` and `riscv_pkg.sv` the slicing of an enumeration constant is illegal for xcelium. The size of `MODE_SV` has been fixed to the size of `vm_mode_t` which removes the need for slicing;
* In `pmp.sv` assigning the port `conf_addr_prev_i ` of `pmp_entry` with a ternary operator caused an error. The assignment has been done outside the `pmp_entry` module.

**Warnings:**
* In `ariane_pkg.sv` removed extra semicolons at the end of endfunctions.

### DIFF
--- a/include/ariane_pkg.sv
+++ b/include/ariane_pkg.sv
@@ -485,7 +485,7 @@ package ariane_pkg;
             EQ, NE, LTS, GES, LTU, GEU: return 1'b1;
             default                   : return 1'b0; // all other ops
         endcase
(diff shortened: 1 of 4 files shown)


zarubaf (approved):
Thanks. LGTM!

PR #616, opened 2021-02-18
Sv32 MMU and C.JAL addition for CV32A6

sjthales:
Added Sv32 MMU for the CV32A6 core.
Added support for C.JAL instruction which was not supported for RV32C.


ASintzoff:
Could you split the commit "adding mmu sv32, adding c.jal support for RV32C" to extract the c.jal support in a dedicated commit? This will separate the topics and we could get the c.jal support faster if sv32 needs more discussion and work.

zarubaf:
Yes please split! The Jal can go immediately!

sjthales:
Ok I  do that

PR #618, opened 2021-02-19
peripherals: assign default value to unused interrupts

niwis:
Otherwise, the PLIC registers will be poisoned.


zarubaf (approved):
Thanks. LGTM

PR #625, opened 2021-03-01
Add custom CSRs for flushing caches

E-Poulin:

### DIFF
--- a/src/ariane.sv
+++ b/src/ariane.sv
@@ -179,9 +179,11 @@ module ariane import ariane_pkg::*; #(
   logic                     tw_csr_id;
   logic                     tsr_csr_id;
   irq_ctrl_t                irq_ctrl_csr_id;
-  logic                     dcache_en_csr_nbdcache;
   logic                     csr_write_fflags_commit_cs;
+  logic                     dcache_en_csr;
   logic                     icache_en_csr;
+  logic                     dcache_flush_csr;
+  logic                     icache_flush_csr;
   logic                     debug_mode;
   logic                     single_step_csr_commit;
   riscv::pmpcfg_t [15:0]    pmpcfg;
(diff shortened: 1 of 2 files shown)


src/ariane.sv
```diff
@@ -661,15 +665,15 @@ module ariane import ariane_pkg::*; #(
...
     .icache_miss_o         ( icache_miss_cache_perf      ),
     .icache_areq_i         ( icache_areq_ex_cache        ),
     .icache_areq_o         ( icache_areq_cache_ex        ),
     .icache_dreq_i         ( icache_dreq_if_cache        ),
     .icache_dreq_o         ( icache_dreq_cache_if        ),
     // D$
-    .dcache_enable_i       ( dcache_en_csr_nbdcache      ),
-    .dcache_flush_i        ( dcache_flush_ctrl_cache     ),
+    .dcache_enable_i       ( dcache_en_csr               ),
+    .dcache_flush_i        ( dcache_flush_ctrl_cache | dcache_flush_csr     ),
```
zarubaf, line 676: I think you should wait for `dcache_flush_ack_cache_ctrl` to go high or else you could be missing a flush event.

src/csr_regfile.sv
```diff
@@ -1089,8 +1092,8 @@ module csr_regfile import ariane_pkg::*; #(
             mcounteren_q           <= {riscv::XLEN{1'b0}};
             mscratch_q             <= {riscv::XLEN{1'b0}};
             mtval_q                <= {riscv::XLEN{1'b0}};
-            dcache_q               <= {{riscv::XLEN-1{1'b0}}, 1'b1};
-            icache_q               <= {{riscv::XLEN-1{1'b0}}, 1'b1};
+	    dcache_q               <= {{riscv::XLEN-2{1'b0}}, 1'b0, 1'b1}; // dcache flush disable by default and dcache enable by default
```
zarubaf, line 1095: The spacing is off, are you using tabs? Please make sure you set your editor to:

- Trim trailing whitespaces
- Used spaces for indentation

zarubaf (commented):
Thanks for tackling this. In general this LGTM, I do think that the flushing isn't entirely correct as the request to flush the cache could not be registered in the data cache. But if it works for you we can merge (anyway a non-standard feature).

src/ariane.sv
E-Poulin: Thank your for pointing this mistake. I'm going to correct it.

Jbalkind:
Just wondering what the exact intent is with this flush - only to flush the L1s? Has this been tested with the write-through subsystem eg in the OpenPiton+Ariane config?

Just to note, the cache management ops proposal is finally becoming formed which may subsume this capability eventually: https://github.com/riscv/riscv-CMOs/pull/21

E-Poulin:
> Just wondering what the exact intent is with this flush - only to flush the L1s? Has this been tested with the write-through subsystem eg in the OpenPiton+Ariane config?
> 
> Just to note, the cache management ops proposal is finally becoming formed which may subsume this capability eventually: [riscv/riscv-CMOs#21](https://github.com/riscv/riscv-CMOs/pull/21)

Yes, the intent of this functionality is to flush only the L1s. It has not been tested with OpenPiton. 
Thank you for your comment and link about riscv-CMOs. I was not aware of the progress of it. You're right, this functionality may be replaced in the future by CMOs.

Jbalkind:
Thanks!

Issue #169, opened 2019-01-28
Mounting a larger file system on the FPGA board


janhoogerbrugge:
Hi,

I want to run some benchmarks on the FPGA board. How to get these on the FPGA board? If I follow
the procedure on https://github.com/pulp-platform/ariane-sdk, then the first partition is too small
and if I make it larger then it does not fit anymore in memory when it loaded with rootfs.  So, I want
to use the second partition of the sdcard. But how to mount this partition on the board. Is there
something still necessary in the RTL code to access the sdcard or does something in the 
kernel to be configured?

Regards,
Jan

Moschn:
How big is your benchmark?

Indeed, it is not possible to mount the second partition of the sd card currently. I think there is a problem with interrupts/the plic. 

However, it should be possible to create a bigger partition. You just have to adapt the bootloader aswell, since only the first 16mb of the partition are copied to RAM. This can be changed in [this line](https://github.com/pulp-platform/ariane/blob/master/fpga/src/bootrom/src/main.c#L11) where the second parameter times 512 is the amount of bytes that are copied (2 * 16384 * 512 = 16mb).

janhoogerbrugge:
The benchmark is SPEC2006 (integer only). If I want to run all the benchmarks in one go I need about 1G. If I would run the benchmarks one by one (or a few at a time) then perhaps a few 100 MB is also fine. I will experiment with the line in the source code you referred to and see how far I will get.

Thanks!

janhoogerbrugge:
One more question. The command to format the sdcard is:

  % sudo sgdisk --clear --new=1:2048:67583 --new=2 --typecode=1:3000 --typecode=2:8300 /dev/sdb # create a new gpt partition table and two partitions: 1st partition: 32mb (ONIE boot), second partition: rest (Linux root)

So there a 32M partition is created but 16M are loaded by the bootloader. Shouldn't these numbers not be the same or will everything go fine as long as you do not put more data in the partition than the bootloader will load?

Moschn:
Yes, you are correct. The sizes should probably be the same. I think the rationale behind only copying 16mb was to save a little bit of time during the boot.

janhoogerbrugge:
I replaced
  
  sd_copy(dest, boot->first_lba, size);

by 

  sd_copy(dest, boot->first_lba, boot->last_lba - boot->first_lba + 1)

in the bootloader so that I do not have to re-synthesize every time I change the size of the
[...]

PR #627, opened 2021-03-03
Copy actual partition size instead of hardcoded 16MB

davidschrammel:
Changed the size of the copied partition as discussed in #169.
I have tested this on my setup and it seems to be working as expected.

Should I also include the updated bootrom.h and bootrom.sv files in this PR or do you want to build them yourself? Since these a binary blobs, it makes sense that an official maintainer creates these.

At some point we might also want to change the name of `gpt_find_boot_partition`, since it does not just _find_ but also _copy_ the boot partition. Also, the `size` argument is now unused:
https://github.com/openhwgroup/cva6/blob/7760227f887c59a2ae1b89e70fb20582a2d5c5b3/fpga/src/bootrom/src/gpt.c#L7
https://github.com/openhwgroup/cva6/blob/0bbc3cb5cf43cd3bf77d3872e46dd0d6ededf28f/openpiton/bootrom/linux/src/main.c#L14
https://github.com/openhwgroup/cva6/blob/0bbc3cb5cf43cd3bf77d3872e46dd0d6ededf28f/fpga/src/bootrom/src/main.c#L11


zarubaf:
This is great, thanks! Please go ahead and re-generate the bootrom if that is possible. I trust that your `riscv-gcc` will do similar things to mine :-)

P.S.: Can you please sign your commit using `git commit -s`? It seems that your user is already covered by an ECA so that should be the only missing step.

zarubaf:
Thanks a lot!

PR #631, opened 2021-03-12
ptw: Kill pending dcache request when flushed

niwis:
Instead of waiting for the dcache request to complete and then to discard the result, kill it and directly return to the `IDLE` state.

This prevents pollution of the data cache and makes the ptw (and dcache) faster ready for new requests.


zarubaf:
Sorry for the hassle, but can you try closing this PR and re-opening it in a new one? It seems that Travis is back, but didn't pick-up old PRs.

niwis:
Sure!

PR #632, opened 2021-03-16
Makefile: Remove excess questa cmd for target sim

niwis:
Remove the static `vsim -do` argument for target `sim`. This argument takes precedence over `questa-cmd` (which defines the desired `-do` sequence) and seems to be an artifact from https://github.com/openhwgroup/cva6/commit/296ce392113003533af863518b29114b9953d836. It results in a batch-like behavior for questasim simulations in GUI mode - in particular closing QuestaSim immediately after terminating the simulation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -334,8 +334,7 @@ generate-trace-vsim:
 	make generate-trace
 
 sim: build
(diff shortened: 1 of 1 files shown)


zarubaf (approved):
LGTM, thanks!

PR #643, opened 2021-03-29
Index width adaptation to cache size and associativity

E-Poulin:


zarubaf (approved):
Thanks. LGTM

PR #650, opened 2021-04-13
:wastebasket: Remove redundant prefix in riscv_pkg

flaviens:
Some tools such as sv2v have some difficulties to deal with these redundant prefixes `riscv::` in the _include/riscv_ package. For instance,sv2v treats these as circular package dependency. Since they bring nothing, it may be useful to remove them.


zarubaf (approved):
Oops, indeed. Good catch!

PR #652, opened 2021-04-14
src/ariane.sv: Verilator: change single-letter string to byte

flaviens:
The string in this block is restrained to one letter. Using a byte (also displayable using `%s`) instead makes some tools' life easier, and makes it arguably more readable.


zarubaf (approved):
Thanks, LGTM!

PR #665, opened 2021-05-05
Performance Counters Fix

OttG:
## Performance counter fix
* Fixed the for loop exit condition, now both commit ports are properly tracked;
* Added JALR to calls counter and modified the typo for the second destination register from x1 to x5 (AFAIK it can be used as alternate register to store the return address);
* For returns from call we only check if the instruction is a JALR and the destination register is x0.

### DIFF
--- a/src/perf_counters.sv
+++ b/src/perf_counters.sv
@@ -67,7 +67,7 @@ module perf_counters import ariane_pkg::*; (
         perf_counter_d[riscv::CSR_MDTLB_MISS] = perf_counter_q[riscv::CSR_MDTLB_MISS] + 1'b1;
 
       // instruction related perf counters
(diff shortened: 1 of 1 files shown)


zarubaf (approved):
Great, LGTM!

zarubaf:
It is interesting that the `rsort` benchmark failed. Can you reproduce this locally?

OttG:
Somehow this change breaks stuff.. I tried to decouple the increments from the check (this actually also correctly increases the registers if both retired instruction are from the same category) but now the benchmakrs work but the ISA tests fail. I'm not sure if it's a problem in Verilator, the tests I'm working with in xcelium are terminating correctly. At anyrate, I can't see why this would cause problems.

PR #667, opened 2021-05-10
Compile corev_apu

MikeOpenHWGroup:
This PR updates the Makefile so that the `make verilate` target compiles the apu and the core.


zarubaf (approved):
Great, LGTM!

PR #668, opened 2021-05-12
icache: Reset FSM to FLUSH

niwis:
@suehtamacv encountered this unexpected behavior. The current implementation resets the icache FSM to `IDLE`, where the icache checks if it was just enabled (rising edge of `en_i`), and flushes (initialises) itself if so. This assumes that `en_i` is deasserted during reset, which is not necessarily the case (e.g. hardwired `en_i`), possibly leaving the cache uninitialised. Hence, reset to `FLUSH` directly.

### DIFF
--- a/src/cache_subsystem/cva6_icache.sv
+++ b/src/cache_subsystem/cva6_icache.sv
@@ -457,7 +457,7 @@ end else begin : gen_piton_offset
       cmp_en_q      <= '0;
       cache_en_q    <= '0;
       flush_q       <= '0;
-      state_q       <= IDLE;
+      state_q       <= FLUSH;
       cl_offset_q   <= '0;
       repl_way_oh_q <= '0;
       inv_q         <= '0;


zarubaf:
I guess this is okay. Why doesn't the assumption on `en_i` hold? That is driven from a reset register, isn't it?

suehtamacv:
> I guess this is okay. Why doesn't the assumption on `en_i` hold? That is driven from a reset register, isn't it?

In my case in particular, it is driven from a tie high cell. But the assumption should hold indeed. I will investigate this further. This being said, I guess it makes sense to initialize the cache to the `FLUSH` state?

zarubaf:
Yes, I think this makes sense. I am basically asking because there have been some people doing post-layout simulation and observing something similar I guess. The tool shouldn't infer a tie-high where we are waiting for an edge, shouldn't it? Please keep me posted :-D Merging this anyway as I think it makes sense.

zarubaf (approved):
LGTM

niwis:
> I guess this is okay. Why doesn't the assumption on `en_i` hold? That is driven from a reset register, isn't it?

Yes, but the driving register is reset to `1`, hence no rising edge I assume:

https://github.com/openhwgroup/cva6/blob/6963fe9667d68c06170940f01bcb0ed61e816074/src/csr_regfile.sv#L1093

zarubaf:
You don’t need a rising edge on the `en_i`, the “rising edge” is produced by the one cycle latency here:

https://github.com/openhwgroup/cva6/blob/6963fe9667d68c06170940f01bcb0ed61e816074/src/cache_subsystem/cva6_icache.sv#L155

Not sure why it should work in simulation but doesn’t in post-synthesis. That is something we should be tracking down.

PR #678, opened 2021-06-07
Fix warning messages when using 32 bits version

E-Poulin:

### DIFF
--- a/src/csr_regfile.sv
+++ b/src/csr_regfile.sv
@@ -30,7 +30,7 @@ module csr_regfile import ariane_pkg::*; #(
     input  logic [NrCommitPorts-1:0]              commit_ack_i,   // Commit acknowledged a instruction -> increase instret CSR
     // Core and Cluster ID
     input  logic[riscv::VLEN-1:0] boot_addr_i,                // Address from which to start booting, mtvec is set to the same address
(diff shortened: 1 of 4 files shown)


src/csr_regfile.sv
```diff
@@ -30,7 +30,7 @@ module csr_regfile import ariane_pkg::*; #(
     input  logic [NrCommitPorts-1:0]              commit_ack_i,   // Commit acknowledged a instruction -> increase instret CSR
     // Core and Cluster ID
     input  logic[riscv::VLEN-1:0] boot_addr_i,                // Address from which to start booting, mtvec is set to the same address
-    input  logic[riscv::XLEN-1:0] hart_id_i,                  // Hart id in a multicore environment (reflected in a CSR)
+    input  logic[riscv::VLEN-1:0] hart_id_i,                  // Hart id in a multicore environment (reflected in a CSR)
```
zarubaf, line 33: That should be `xlen` shouldn't it?

zarubaf (changes requested):

src/csr_regfile.sv
E-Poulin: Yes, you're right. I've checked the RISCV privileged manual and it's `xlen`. I'm gonna change it.

zarubaf (approved):
👍

PR #685, opened 2021-06-17
Add verification support on cva6_reorg branch to 

ASintzoff:
- support of rfvi
- support of vcs
- support of vsim
- support of FPGA

Note this PR is strongly linked to the corresponding PR on core-v-verif repository.

Modifications done here are needed to have cva6 verification provided in core-v-verif working after the cva6 reorganization.


core/include/riscv_pkg.sv
```diff
@@ -28,7 +28,11 @@ package riscv;
        ModeSv64 = 11
     } vm_mode_t;
 
+`ifdef CV32A6
```
MikeOpenHWGroup, line 31: Is there a way we can do this without using compiler macros (`defines)?

MikeOpenHWGroup, line 31: Is there a way we can do this without compiler macros (`defines)?

MikeOpenHWGroup (commented):
This looks fine to me @ASintzoff, thanks for this!  I have made a couple of review comments, but these are for discussion, they do not need to gate the merge.   Having said that, I will not approve the merge since this should come from @zarubaf.

.gitmodules
```diff
@@ -42,7 +42,7 @@
 	url = https://github.com/pulp-platform/axi_riscv_atomics.git
 [submodule "corev_apu/riscv-dbg"]
 	path = corev_apu/riscv-dbg
-	url = https://github.com/pulp-platform/riscv-dbg.git
+	url = https://github.com/JeanRochCoulon/riscv-dbg.git
```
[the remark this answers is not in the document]
ASintzoff: I would prefer to not fork if possible.
IMO, it seems better to modify riscv-dbg according to https://github.com/pulp-platform/riscv-dbg/pull/83

core/include/riscv_pkg.sv
ASintzoff: Good point. @JeanRochCoulon will address it. Details will follow.

JeanRochCoulon: XLEN is defined in a package. As there is no way to override a parameter in a package, we can create different versions of the same package and choose which version to compile. By the way, Flist.cva6 should be duplicated into Flist.cv32a6 and Flist.cv64a6. In that way, all the supported configurations (to be referred to cva6 requirement specification) would correspond to a dedicated FLIST file as FLIST.cva64a6_noFD for instance (naming to be fine-tuned of course).

@zarubaf @MikeOpenHWGroup, I can't wait for having your feedback !?!

.gitmodules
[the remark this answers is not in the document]
zarubaf: I think @JeanRochCoulon and I agree that the problem he was seeing because of a bug in the debug unit that should have been fixed by later commits. Can you try updating to the latest `riscv-dbg` module? We need to merge this before we can merge this PR as the public CVA6 can't depend on "private" forks of members.

Makefile
```diff
@@ -36,13 +36,7 @@ root-dir := $(dir $(mkfile_path))
 
 support_verilator_4 := $(shell ($(verilator) --version | grep '4\.') > /dev/null 2>&1 ; echo $$?)
 ifeq ($(support_verilator_4), 0)
-	ifndef verilator_threads
```
zarubaf, line 39: Why is this removed, isn't it working correctly?

core/ariane.sv
```diff
@@ -42,17 +42,153 @@ module ariane import ariane_pkg::*; #(
   // firesim trace port
   output traced_instr_pkg::trace_port_t trace_o,
 `endif
+  // RISC-V formal interface port (`rvfi`):
```
zarubaf, line 45: I would prefer structs on the top ports. Also, I guess we want to make the inclusion of the tracer port optional to remain backward compatibility.

core/ariane.sv
```diff
@@ -42,17 +42,153 @@ module ariane import ariane_pkg::*; #(
...
+  output logic[riscv::XLEN-1:0]        rvfi_mem_wdata_0,
 `ifdef PITON_ARIANE
   // L15 (memory side)
   output wt_cache_pkg::l15_req_t       l15_req_o,
   input  wt_cache_pkg::l15_rtrn_t      l15_rtrn_i
 `else
   // memory side, AXI Master
-  output ariane_axi::req_t             axi_req_o,
-  input  ariane_axi::resp_t            axi_resp_i
+  output ariane_axi::id_t              aw_id_o,
```
zarubaf, line 95: Strongly prefer the struct variant. Why would you unroll?

corev_apu/tb/ariane_testharness.sv
```diff
@@ -722,6 +811,17 @@ module ariane_testharness #(
...
+  rvfi_tracer  #(
+    .SIM_FINISH(2000000),
+    .HART_ID(hart_id),
+    .DEBUG_START(0),
+    .DEBUG_STOP(0)
+  ) rvfi_tracer_i (
+    .clk_i(clk_i),
+    .rst_ni(rst_ni),
+    .rvfi_i(rvfi)
+  ) ;
```
zarubaf, line 823:
```suggestion
  );
```

core/include/riscv_pkg.sv
zarubaf: Technically you could fork that out into two separate packages `cv64a_config_pkg.sv` and `cv32a_config_pkg.sv` and include the one that you are looking for.

zarubaf: My feeling is that, at a certain point, we are better off if we generate part of the RTL from templates as we are hitting on the limits of SV.

zarubaf (changes requested):
Generally looks okay to me:

- We need to get the debug thing sorted.
- CI is failing
- I really don't like the unrolled interfaces, what is the rationale?
- Small nits

Makefile
ASintzoff: In 32-bit configuration, tests in riscv-tests and riscv-compliance test suites are randomly stucked. Several executions of the same test are stopping at different places. Stopping means no more instruction logged in the trace but `cycles` still incremented.

corev_apu/tb/ariane_testharness.sv
ASintzoff: Good point

.gitmodules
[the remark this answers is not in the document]
ASintzoff: With latest version of `riscv-dbg`, the problem disappears. PR has been adapted accordingly.

core/ariane.sv
JeanRochCoulon: Florian,

When structure are used in RTL, dc_shell from Synopsys flattens the structure by generating a logic bit vector. A signal referring to rvfi_inst_t typedef will be mapped by dc_shell on a single logic bit vector of more than 100 bits. While it could remain transparent when the signal is internal, it become a problem when it is in IO.

For instance, to execute a gate simulation, we replace the ariane RTL module by the ariane gate module, but testharness’s RVFI and AXI structures cannot be connected to the ariane’s two logic bit vectors.

I ask for a command line option to the Synopsys support, but without success.

We already get the same problem, a long time ago on LEON Sparc. We had flattened the structures.

My proposal is to expand as minimal the ariane input and output which prevent us from executing gate simulation connected to the testharness, that’s to say RVFI and AXI.

Regards

core/include/riscv_pkg.sv
JeanRochCoulon: As you will see, a RTL proposal has been pushed.  A new package has been created to host the cva6 configuration. Two Flist.cva6xxx files have been created, one to point on the package file where XLEN=64 and the other to the package file with XLEN=32. According to the variant, the Makefile selects the right Flist.cva6xxx file. Tell me if it works for you! Cheers

core/ariane.sv
zarubaf: I understand your use-case, then let's add a wrapper around the main `ariane.sv` (`ariane_wrapepr.sv`?) that exposes the interfaces as flat ports. In that way, we do not break the interface for existing users.

core/ariane.sv
JeanRochCoulon: @zarubaf    As discussed, the structures generate issues at dc_shell synthesis time. To solve the problem, a wrapper has been added in core testbench (testbench used for UVM) following the suggestion, testharness is kept unchanged.

core/Flist.cv64a6_imacfd_sv39
```diff
@@ -0,0 +1,153 @@
+//////////////////////////////////////////////////////////////////////////////
```
zarubaf, line 1: Super nitpick: The isa string is `imafdc` not `imacfd`

core/ariane.sv
zarubaf: Adding an interface to the SystemVerilog module breaks its interface so we should ifdef similar to the firesim trace port.

core/include/cv32a6_imac_sv0_config_pkg.sv
```diff
@@ -0,0 +1,15 @@
...
+// you may not use this file except in compliance with the License.
+// SPDX-License-Identifier: Apache-2.0 WITH SHL-2.0
+// You may obtain a copy of the License at https://solderpad.org/licenses/
+//
+// Original Author: Jean-Roch COULON (jean-roch.coulon@thalesgroup.com)
+
+
+package cva6_config_pkg;
+
+    localparam cva6_config_XLEN = 32;
```
zarubaf, line 13: Nit: According to the style guide localparams are supposed to be CamelCase.
```suggestion
    localparam CVA6ConfigXlen = 32;
```

core/include/cv64a6_imacfd_sv39_config_pkg.sv
```diff
@@ -0,0 +1,15 @@
...
+// you may not use this file except in compliance with the License.
+// SPDX-License-Identifier: Apache-2.0 WITH SHL-2.0
+// You may obtain a copy of the License at https://solderpad.org/licenses/
+//
+// Original Author: Jean-Roch COULON (jean-roch.coulon@thalesgroup.com)
+
+
+package cva6_config_pkg;
+
+    localparam cva6_config_XLEN = 64;
```
zarubaf, line 13: Same nit here

zarubaf (commented):
Almost there.

core/Flist.cv64a6_imacfd_sv39
JeanRochCoulon: Yes, super nitpick!!!! :-)     Fixed by commit

core/include/cv32a6_imac_sv0_config_pkg.sv
JeanRochCoulon: Fixed by commit

core/include/cv64a6_imacfd_sv39_config_pkg.sv
JeanRochCoulon: Fixed by commit

JeanRochCoulon:
@zarubaf    Concerning the RFVI interface add-on, I am confused: I though @MikeOpenHWGroup would like to limit/remove all the ifdef from CVA6 ?!  Mike, could you confirm it?

core/include/riscv_pkg.sv
```diff
@@ -14,8 +14,16 @@
  *
  * Description: Common RISC-V definitions.
  */
+
+import cva6_config_pkg::*;
```
zarubaf, line 18: Let's move that into the package. It otherwise pollutes the namespace

zarubaf:
> @zarubaf Concerning the RFVI interface add-on, I am confused: I though @MikeOpenHWGroup would like to limit/remove all the ifdef from CVA6 ?! Mike, could you confirm it?

What we can't change is the top-level interface. (well we can but that would mean a major version bump - that is not necessary imho). Other approach would be to promote `ariane.sv` to a wrapper around `ariane_v2.sv` which ties off the ports - ugly. Imho we should go with the ifdef.

core/include/riscv_pkg.sv
JeanRochCoulon: done

JeanRochCoulon:
Ok, understood, thanks for the clarification, I have added a RVFI_TRACE ifdef in ariane module.

JeanRochCoulon:
@zarubaf  @MikeOpenHWGroup   Very good, we have solved the riscv-dbg issue, managed the different configurations (will be useful), and added RVFI. Splendid.  
Asap the merge is done, we tranfer Thales guys (and who wants) on the new organization.  Cheers

corev_apu/fpga/src/ariane_xilinx.sv
```diff
@@ -424,16 +424,59 @@ ariane_axi::resp_t   axi_ariane_resp;
...
-    .clk_i        ( clk                 ),
-    .rst_ni       ( ndmreset_n          ),
-    .boot_addr_i  ( ariane_soc::ROMBase ), // start fetching from ROM
-    .hart_id_i    ( '0                  ),
-    .irq_i        ( irq                 ),
-    .ipi_i        ( ipi                 ),
-    .time_irq_i   ( timer_irq           ),
-    .debug_req_i  ( debug_req_irq       ),
-    .axi_req_o    ( axi_ariane_req      ),
-    .axi_resp_i   ( axi_ariane_resp     )
```
zarubaf, line 436: Wait, this should be reverted to the struct-based version.

zarubaf (approved):
Just the small FPGA fix and then let's merge!

corev_apu/fpga/src/ariane_xilinx.sv
ASintzoff: Modification removed from commit Support FPGA generation

zarubaf (approved):
Thanks for all the work. LGTM - merging.

PR #690, opened 2021-06-25
Make the cache size and assoc configurable

E-Poulin:
The current replacement policy prevented to have a configurable cache size and associativity. Instead, I replaced it with the Galois LFSR present in common_cells.


zarubaf (approved):
Awesome - thanks!

PR #698, opened 2021-07-27
pmp: Fix csr

### LINKED ISSUE #538: PMP risc-v benchmark fails with page fault

### LINKED ISSUE #697: PMP and G value
From the specs:
`In general, the PMP grain is 2^G+2 bytes and must be the same across all PMP regions. When G ≥ 1, the NA4 mode is not selectable. When G ≥ 2 and pmpcfg_i.A[1] is set, i.e. the mode is NAPOT, then bits pmpaddr_i[G-2:0] read as all ones. When G ≥ 1 and pmpcfg_i.A[1] is clear, i.e. the mode is OFF or TOR, then bits pmpaddr_i[G-1:0] read as all zeros.`

While in csr_registers.sv
```
// PMPADDR
// Important: we only support granularity 8 bytes (G=2)
// -> last bit of pmpaddr must be set 0/1 based on the mode:
// NA4, NAPOT: 1
// TOR, OFF:   0
riscv::CSR_PMPADDR0:         csr_rdata = {10'b0, pmpaddr_q[0][riscv::PLEN-3:1], (pmpcfg_q[0].addr_mode[1] == 1'b1 ? 1'b1 : 1'b0)};
```
8-bytes granularity would correspond to G=1 in my understanding.
Moreover, if the code is consistent with the rule which states that pmpaddr[G-2:0] read as all ones, it is not compliant with the one which states that pmpaddr[G-1:0] read as all zeros (it should be the last 2 bits, not the last one).

Also, it looks like NA4 is selectable.
I can write pmpcfg.a=NA4, and the PMP implementation behaves accordingly.
For example, with the following configuration
pmpcfg0.r = 1;
pmpcfg0.w = 1;
pmpcfg0.x = 0;
pmpcfg0.a = 2; // NA4
pmpcfg0.o = 0;
pmpcfg0.l = 0;
pmpaddr0=0x40000000;
I can read and write address 0x40000000 but I get an exception when accessing address 0x40000004.

Moschn:
- Add proper granularity (G=1)
- Disallow NA4
- Fix read of pmpaddr
- Bump riscv-tests to support pmp benchmark that detects granularity (https://github.com/riscv/riscv-tests/pull/295)
- Add PMP benchmark to CI

Fixes #538 
Fixes #697

### DIFF
--- a/ci/build-riscv-tests.sh
+++ b/ci/build-riscv-tests.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -e
 ROOT=$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)
-VERSION="7cc76ea83b4f827596158c8ba0763e93da65de8f"
+VERSION="eeacd5507db7a0f50ca8c4f27aff220fcbb60bdf"
 
 cd $ROOT/tmp
 
--- a/ci/riscv-benchmarks.list
+++ b/ci/riscv-benchmarks.list
@@ -5,3 +5,4 @@ qsort.riscv
 rsort.riscv
 towers.riscv
 vvadd.riscv
+pmp.riscv
--- a/src/csr_regfile.sv
+++ b/src/csr_regfile.sv
@@ -925,7 +925,7 @@ module csr_regfile import ariane_pkg::*; #(
             // check counter-enabled counter CSR accesses
             // counter address range is C00 to C1F
             if (csr_addr_i inside {[riscv::CSR_CYCLE:riscv::CSR_HPM_COUNTER_31]}) begin
-                unique case (csr_addr.csr_decode.priv_lvl)
+                unique case (priv_lvl_o)
                     riscv::PRIV_LVL_M: privilege_violation = 1'b0;
                     riscv::PRIV_LVL_S: privilege_violation = ~mcounteren_q[csr_addr_i[4:0]];
                     riscv::PRIV_LVL_U: privilege_violation = ~mcounteren_q[csr_addr_i[4:0]] & ~scounteren_q[csr_addr_i[4:0]];
--- a/src/csr_regfile.sv
+++ b/src/csr_regfile.sv
@@ -1152,8 +1152,11 @@ module csr_regfile import ariane_pkg::*; #(
             // pmp
             for(int i = 0; i < 16; i++) begin
                 if(i < NrPMPEntries) begin
+                    // We only support >=8-byte granularity, NA4 is disabled
                     if(pmpcfg_q[i].addr_mode != riscv::NA4) 
                         pmpcfg_q[i] <= pmpcfg_d[i];
+                    else
+                        pmpcfg_q[i] <= pmpcfg_q[i];
                     pmpaddr_q[i] <= pmpaddr_d[i];
                 end else begin
                     pmpcfg_q[i] <= '0;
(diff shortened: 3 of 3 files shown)


Moschn:
I guess the riscv-tests repo changed something else that brakes now...

I can remove the pmp benchmark and the riscv-test bump from this PR but it would be nice to have automatic testing of the PMPs in the CI

Moschn:
This commit broke the test `rv64mi-p-csr` (equivalent to `rv64si-p-csr`): https://github.com/riscv/riscv-tests/commit/3e175a094ff15ea28ae173af8f00a36c5ed9d296

Not exactly sure why reading the `cycle` csr produces an illegal instruction.

zarubaf:
That is really interesting! The trap should be fairly localized to the `csr_regfile`, can you see the illegal instruction being raised there?

Moschn:
After some investigation, I found and fixed the problem: 

The privilege check should verify that the __current__ privilege level has access to the counter CSRs. Before it checked if the privilege level of the CSR instruction has access to the counters. Essentially, the m-mode could not access any counter csr if mcounteren/scounteren was not set.

zarubaf:
Awesome, thanks a lot!

PR #709, opened 2021-08-20
PMP improvement

masgia:
- PMPCFG.A=NA4 not allowed
- access allowed in S/U mode when no PMP is configured (from the specs: if no PMP entry matches an S-mode or U-mode access, but at least one PMP entry is implemented, the access fails)

Note: the new signal between the commit_stage and the csr_registers is necessary to remove the combinatorial loop
update_access_exception → csr_regfile_i/csr_exception_o → csr_exception_csr_commit → commit_stage_i/csr_exception_i → commit_stage_i/exception_o → commit_stage_i/csr_wdata_o[4] → csr_wdata_commit_csr[4] → csr_regfile_i/csr_wdata_i[4] → update_access_exception

### DIFF
--- a/src/csr_regfile.sv
+++ b/src/csr_regfile.sv
@@ -1154,7 +1154,7 @@ module csr_regfile import ariane_pkg::*; #(
             for(int i = 0; i < 16; i++) begin
                 if(i < NrPMPEntries) begin
                     // We only support >=8-byte granularity, NA4 is disabled
-                    if(pmpcfg_q[i].addr_mode != riscv::NA4) 
+                    if(pmpcfg_d[i].addr_mode != riscv::NA4) 
                         pmpcfg_q[i] <= pmpcfg_d[i];
                     else
                         pmpcfg_q[i] <= pmpcfg_q[i];


Moschn:
PMP CSR registers are supposed to be WARL (write any read legal) so we should __not__ throw an exception on a write of an illegal value. Instead, we should just not update the CSR register (current behavior).

The extra check you add in the pmp_entry to check if PMP is implemented should also not be necessary as no `pmp_entries` are even instantiated if `NR_PMP_ENTRIES == 0`. See here: https://github.com/openhwgroup/cva6/blob/97172398ad5d3be22c127e4fa45d7b63731bbab4/src/pmp/src/pmp.sv#L31

masgia:
You are obviously right.
The extra check in pmp.sv was introduced because I misinterpreted the specs: I assumed they meant "configured" with "implemented" but that doesn't fit.
And I was using an older version of csr_register.sv, so I didn't notice your latest changes.

But I still think there is a small bug in the update of pmpcfg_q: you check 
`if(pmpcfg_q[i].addr_mode != riscv::NA4)`
but I think you should use pmpcfg_d instead.

Moschn:
I must have missed this. Thanks for double-checking and fixing! Would be beneficial to squash the commits in this PR into a single one that contains only the one fix.

Moschn:
LGTM. 

@zarubaf ready to merge from my side

zarubaf:
Thanks. Can you elaborate a bit on why we would want to read the protection value from the `_d` part of the register? Not sure I get the change.

masgia:
PMPCFG.A = NA4 is not supported by CVA6, since it only supports G=1.
So we should prevent from this configuration to be set (NA4 is a value accepted and used by pmp_entry.sv).
The code change tells exactly this: if the new value (_d) is not supported, don't update the register.
With the old code setting NA4 would have been legal, and overwriting it would have been impossible.

zarubaf:
Okay, that makes sense. Thanks!

PR #713, opened 2021-08-27
mmu: Prevent elaboration of incompatible MMU version

niwis:
Fix the VCS flow for 64-bit CVA6 by guarding incompatible MMU implementations. For instance, `mmu_sv32` assumes `PLEN` > `VLEN`, which is not the case for RV64.

Example:
https://github.com/openhwgroup/cva6/blob/720b59e2ca6e002b9b3a2c47ed4ad85b847978a3/src/mmu_sv32/cva6_mmu_sv32.sv#L212
causes for 64-bit CVA6:
```
(vlog-8607) Negative replication multiplier (-8)
```


niwis:
Sorry for the radio silence. This should of course work (and would probably be the more elegant solution). How do you suggest to proceed? Adding the parameter on `master` and updating the Makefile or waiting for `cva6_reorg` to be merged?

zarubaf (approved):
Great, looks much better now :-)

PR #714, opened 2021-08-27
Cv32a6 genesys

sjthales:
Hi,
cv32a6 tested on genesys 2 board via baremetal application
cv64a6 tested on genesys 2 with baremetal application and linux running.


zarubaf:
@sjthales Do you think you can add a portion to the `README` on how to build the 32-bit version. A bit of advertisement and guide can't hurt. Thanks.

PR #716, opened 2021-08-31
Fix typos: 'funct3' in utype_t should be 'imm'.

lingscale:


zarubaf (approved):
Indeed. Thanks for catching this!

PR #723, opened 2021-09-07
updating bootrom for 32b

sjthales:
Verified with Linux 64 bits and Linux 32 bits


zarubaf (approved):
LGTM! @Jbalkind had some concerns regarding OP, are they addressed?

sjthales:
Actually, All I modified regards the FPGA implémentation on Genesys board. According that changes nothing for OP.

Jbalkind:
That's not true. This breaks our build because files were moved. openpiton/bootrom/baremetal has multiple symlinks into bootrom/

Issue #720, opened 2021-09-03
Some illegal instructions are capable of being executed


zarubaf:
Location:  cva6/src/decoder.sv (https://github.com/openhwgroup/cva6/blob/master/src/decoder.sv), line 153 to 162.

Details: The if statement in the default case is missing else. Hence the illegal_instr will not be set even for instructions not meeting the if condition. 

Thus any illegal instruction of the format 64’hxxx00073 (‘x’ is dont-care) will not be detected as illegal by Ariane.

jquevremont:
The input and log files that were forwarded by the team who discovered the issue.
[bug_2.zip](https://github.com/openhwgroup/cva6/files/7114416/bug_2.zip)

Issue #719, opened 2021-09-03
Incorrect implementation of the FENCE.I instruction detection logic


zarubaf:
Location:  cva6/src/decoder.sv (https://github.com/openhwgroup/cva6/blob/master/src/decoder.sv), line 220 to 240.

Details: The RISC-V specification states that the hardware implementations should ignore the imm, rs1, and rs2 fields of the FENCE.I instruction. However, the cv6 implementation expects a hardcoded value of zero in all these fields. Hence, it is flagging valid FENCE.I instructions with non-zero values in these fields as illegal instructions. 

For example, the SPIKE successfully executes the FENCE.I instructions with non-zero imm field but cv6 does not.

jquevremont:
The inputs files and log files provided by the team who sent us the bug notice.
[bug_1.zip](https://github.com/openhwgroup/cva6/files/7114409/bug_1.zip)

PR #724, opened 2021-09-07
Fixes for issues #719, #720

OttG:
As stated in the issue https://github.com/openhwgroup/cva6/issues/719 the immediate field should be ignored for FENCE.I which isn't in the master (the spec also states that standard software should zero out those fields so that's probably why it went unnoticed) (line 234,235).
For issue https://github.com/openhwgroup/cva6/issues/720 I just added the else on the default case which will result into an illegal instruction if no match in the cases are found (line 163, 164).


zarubaf (approved):
Great! Thanks, @OttG. LGTM, let's wait for the CI to finish, and then we can merge.

PR #737, opened 2021-09-27
Fix build after mmu reorg

ASintzoff:
Fix Flist.cv64a6_imafdc_sv39 by removing files related to mmu as they are now in another place.
Without this fix, any build based on such flist fails.


zarubaf (approved):
👍

PR #742, opened 2021-10-01
ariane_testharness: Fix ariane port list

niwis:
Update the port list of ariane's instantiation according to its definition:
https://github.com/openhwgroup/cva6/blob/3ddf797e95923fd11113c8e443046105dfbf8843/core/ariane.sv#L45-L49


zarubaf (approved):
Thanks!

PR #746, opened 2021-10-01
riscv-dbg: update to v0.4.1 to support 32-bit CVA6 debug

ASintzoff:
It seems that this update was not part of the cva6_reorg merge.


zarubaf (approved):
Perfect. Thanks!

PR #750, opened 2021-10-05
remove ariane_wrapper module and ariane_soc package needs in Flist.cv*

ASintzoff:


corev_apu/tb/ariane_soc_pkg.sv
```diff
@@ -11,6 +11,8 @@
 // Author: Florian Zaruba, ETH Zurich
 // Description: Contains SoC information as constants
 package ariane_soc;
+  localparam NR_COMMIT_PORTS = 2;
```
zarubaf, line 14: That isn't entirely clear to me, `NR_COMMIT_PORTS` should be in `ariane_pkg`, right?

corev_apu/tb/ariane_soc_pkg.sv
```diff
@@ -69,7 +71,37 @@ package ariane_soc;
...
+    int unsigned                      NrCachedRegionRules;   // Number of regions which have cached property
+    logic [NrMaxRules-1:0][63:0]      CachedRegionAddrBase;  // base which needs to match
+    logic [NrMaxRules-1:0][63:0]      CachedRegionLength;    // bit mask which bits to consider when matching the rule
+    // cache config
+    bit                               Axi64BitCompliant;     // set to 1 when using in conjunction with 64bit AXI bus adapter
+    bit                               SwapEndianess;         // set to 1 to swap endianess inside L1.5 openpiton adapter
+    //
+    logic [63:0]                      DmBaseAddress;         // offset of the debug module
+    int unsigned                      NrPMPEntries;          // Number of PMP entries
+  } ariane_cfg_t;
```
zarubaf, line 102: Same here. I think we want to keep one single place where we define this!

zarubaf (changes requested):
I think I need some clarification on the re-definition. Otherwise LGTM

PR #751, opened 2021-10-06
Fixed performance counters to properly register the second commit port

OttG:
Fixed `perf_counters.sv` to properly register the second commit port.

Fixed the checks for Calls and Returns. For Calls we check for both `jal` and `jalr` and wether the linking register is x1 (RA) or x5 (the alternative for the return address). For what is concerced Returns, fixed the check on `rd` to be 'd0 instead of 'd1.
I tested these changes on `matmult-int` test from `embench` where on the current version it registers 1/10 of the actual branches (they most commit on port 2).

I added the possibility for increasing the counters by more than one per cycle (up to `NR_COMMIT_PORTS`), although this makes it a bit cluttery. Also, this might be very rare, maybe with store and loads it should be possible to commit 2 at a time but I'm not sure about branch/calls/returns. If you think is too messy I could revert that part as it was before.


OttG:
I agree, this looks much better. I tested on the same matmult I used as example and looks like is working. Let's wait for CI to be sure (btw I squashed the last commit let me know if is ok).

PR #757, opened 2021-10-11
Makefile, ariane_testharness.sv, rvfi_tracer.sv: configure SIM_FINISH, InclSimDTM

ASintzoff:
For instance, thanks to DTM disable, possible to run longer simulations


zarubaf:
Sorry, saw the pr and forgot about it again. I am okay with the idea of the changes. But we are trying to avoid `ifdefs`. Usually the simulator allows you to set parameters of the top-level. In questa it is `-g`, I do not know about Verilator and others but maybe you can double check that. Would in my opinion be the much more elegant solution.

PR #760, opened 2021-10-13
update make fpga variant=32 to support linux 32b

sjthales:
Following @nathpranay issue, I updated release 4.3 to ease cv32a6 release.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,7 +74,7 @@ endif
 # Sources
 # Package files -> compile first
 ifeq ($(findstring 32, $(variant)),32)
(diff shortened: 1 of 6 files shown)


zarubaf:
My impression was that this needs to be updated now that we've merged #769, right? Please ping me once that is ready or in case I didn't understand it correctly.

sjthales:
Yes, master branch and this release needs to be the same. However I think master branch also may need to be adapted. I will come back to you as soon as possible.

nathpranay (approved):

zarubaf (approved):

Issue #321, opened 2019-09-19
Exit from WFI


suppamax:
Hi,

in csr_regfile.sv I see this condition is used to exit from WFI
```        
if (|mip_q || debug_req_i || irq_i[1]) begin
            wfi_d = 1'b0;
```
I don't understand why the core should be waken up in case of a pending interrupt, even if it has not been enabled via MIE, or in case of an external irq for supervisor mode, again unmasked.
In case an interrupt is received, the code will continue its execution but might not enter the ISR.

zarubaf:
As far as I remember the spec once said that any pending interrupt (even when not enabled) should un-stall the core. In the newer version the specification is much clearer on the behavior. Still, you can't place any assumptions on the behavior of the `wfi` it could be implemented as a `nop`. Agreeing that this should be fixed in the next version.

MikeOpenHWGroup:
Hi @suppamax , this _looks_ like it was resolved, can you confirm?

suppamax:
:+1:

PR #765, opened 2021-11-17
Mask non-enabled interrupts in WFI control

colluca:
This PR addresses issue #321.

The current revision of the RISC-V Priviliged ISA specification states:
> The operation of WFI must be unaffected by the global interrupt bits in mstatus (MIE and SIE) [...], but should honor the individual interrupt enables (e.g, MTIE) (i.e., implementations should avoid resuming the hart if the interrupt is pending but
not individually enabled).

This behaviour is not mandatory but it is suggested (could we say also intuitive?).
Furthermore it would have positive implications on power.

Suppose the core is stalled on a WFI. Suppose timer interrupts are disabled, as we are expecting a SW interrupt to resume the hart. By specification, a timer interrupt will eventually become pending, and if this condition occurs before the SW interrupt arrives, the hart will resume from the WFI and start looping until the SW interrupt pending bit is set or the timer interrupt bit is cleared. In both cases this results in the core doing more work than with the implementation suggested by the spec.

### DIFF
--- a/core/csr_regfile.sv
+++ b/core/csr_regfile.sv
@@ -966,9 +966,9 @@ module csr_regfile import ariane_pkg::*; #(
     always_comb begin : wfi_ctrl
         // wait for interrupt register
         wfi_d = wfi_q;
-        // if there is any interrupt pending un-stall the core
+        // if there is any (enabled) interrupt pending un-stall the core
         // also un-stall if we want to enter debug mode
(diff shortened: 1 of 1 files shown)


zarubaf:
Great, fully agree! Thanks!

zarubaf (approved):
👍

Issue #514, opened 2020-08-28
Lack of return instruction in read_section function defined in tb/dpi/elfloader.cc


pawelkudlakaldec:
'read_section' should return a value.
Currently it returns nothing.
See code below:

extern "C" char read_section (long long address, const svOpenArrayHandle buffer) {
    // get actual poitner
    void* buf = svGetArrayPtr(buffer);
    // check that the address points to a section
    assert(mems.count(address) > 0);
    // copy array
    int i = 0;
    for (auto &datum : mems.find(address)->second) {
      *((char *) buf + i) = datum;
      i++;
    }
}

It may cause compilation error if gcc gets -Werror=return-type option.
See fragment of console below:

tb/dpi/elfloader.cc: In function ‘char read_section(long long int, svOpenArrayHandle)’:
tb/dpi/elfloader.cc:63:1: error: control reaches end of non-void function [-Werror=return-type]
   63 | }
      | ^
cc1plus: some warnings being treated as errors

MikeOpenHWGroup:
Hi @pawelkudlakaldec, thanks for your interest in CVA6.  This issue has not been updated in more than a year, and there seem to be merged pull-requests associated with it, so I will assume it is resolved and will close this issue.  If you are still having trouble, please feel free to open another.

PR #767, opened 2021-11-22
elfloader.cc: change read_section prototype

Gchauvon:
As read_section does not return any value, return type is void.

VCS generates segmentation fault when preloading elf without this fix. Related to issue #514.


zarubaf (approved):
Indeed. Thanks for fixing!

PR #771, opened 2021-11-30
Explicit void type should be used for read_section function.

pawelkudlakaldec:
It is related to #767 pullreq. The prototype of 'read_section' function should return explicitly void in SV side.
What is more issue #514 should be closed.


zarubaf (approved):
:+1:

zarubaf:
Thanks a lot!

PR #772, opened 2021-12-02
Add Verdi option and memory preloading for VCS simulator

Gchauvon:
Hello,

I added some support for vcs :
- moved vcs work files to a work directory.
- added debug options to vcs compilation to use VERDI.
- enable preloading for vcs and disable debug because it was reloading memory.
- made explicit the connection for critical_word in miss_handler, it was raising an error for vcs when using 32bit configuration.

Memory preloading makes simulation faster.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,8 @@
 library        ?= work
 # verilator lib
 ver-library    ?= work-ver
+# vcs lib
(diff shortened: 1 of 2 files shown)


zarubaf (approved):
Great. Agree with @JeanRochCoulon

PR #776, opened 2021-12-06
tb_*cache: Update source list

colluca:
Testbench broke after the changes to the directory structure.

This PR updates the source lists according to the new directory structure, such that the TB runs again.


niwis:
Thanks @colluca. Would it make sense to do the same for [`tb_cva6_icache/tb.list`](https://github.com/openhwgroup/cva6/blob/master/corev_apu/tb/tb_cva6_icache/tb.list) and [`tb_wt_dcache/tb.list`](https://github.com/openhwgroup/cva6/blob/master/corev_apu/tb/tb_wt_dcache/tb.list) while we're at it?

Cheers, Nils

colluca:
I guess you're right, will look into that.

I also had to [hardcode](https://github.com/colluca/cva6/blob/07836409314446bb61a6b9985c1bfbaad01971da/corev_apu/tb/tb_wb_dcache/tb.list#L1) the TB to use the 64-bit version of Ariane.
I was using the 32-bit version at the beginning and the TB failed. Could it be worthwhile to create an issue on this?

zarubaf (approved):
Great, thanks @colluca. Is this ready to go from your side? Although you are covered with the Eclipse committer agreement it would be extra nice if you could sign-up (https://accounts.eclipse.org/user/login?destination=user/eca) and sign off the commits so that the one CI check passes. Let me know if you need some more pointers. @niwis should also have this set-up for CVA6. Thanks

colluca:
Done. 
I fixed also the TB of WT D$ and I$, so it's ready to go.

zarubaf:
Thanks!

PR #780, opened 2021-12-17
Add CoreV-X-Interface, top hierarchy modification and rs3 for global purpose register file.

Gchauvon:
Hello,

I propose a first implementation of the CoreV-X-Interface protocol for the cva6. You'll find the protocol here: [CoreV-X-Interface](https://docs.openhwgroup.org/projects/openhw-group-core-v-xif/x_ext.html).

My implementation supports all mandatory interfaces from the protocol: 
- Issue interface
- Commit interface
- Result interface

Memory Interface and memory result interface used to connect coprocessor to CPU memory are not supported yet.
Compressed Interface may not be necessary as cva6 already support Compressed extension.

### CoreV-X-Interface Implementation
CoreV-X-IF is enabled via the `CVXIF_PRESENT` flag in `ariane_pkg.sv`. 
All illegal instructions are decoded under the R4-format. A new functional unit called `cvxif_fu.sv` is implemented in the `ex_stage.sv` and will proceed to create request on the X-Interface to offload illegal instructions. This module handle responses from the coprocessor aswell. 
In case an instruction is rejected ( illegal for CVA6 and coprocessor ), an `ILLEGAL_INSTR` exception is sent via the writeback bus. If the instruction is accepted, the coprocessor will give the result of the instruction so it can be writeback to the register file via the writeback bus aswell.

### Hierarchy modification
This PR contains a modification of the hierarchy for top modules. 
A new module called `cva6.sv` is basically `ariane.sv` with cvxif ports.
`ariane.sv` is now a wrapper containing `cva6.sv` and an example coprocessor when CVXIF is enable. `ariane.sv` keeps the same inputs/outputs as before so people that already have instantiated ariane can still use the repository.
Testharness now looks like this: 
hierarchy

### 3rd read ports for ariane_regfile
I also added the possibility to have a 3rd read port for the global purpose register ( `ariane_regfile` ). The CoreV-X-Interface protocol allows the CPU to transmit 3 sources registers to the coprocessor and it was my use case during development. I made it configurable via the `ariane_pkg::NR_RGPR_PORTS` because adding this 3rd port has an impact in area.


JeanRochCoulon (approved):

core/cvxif_fu.sv
```diff
@@ -0,0 +1,72 @@
...
+      cvxif_req_o.x_result_ready = 1'b1;
+      x_ready_o = cvxif_resp_i.x_issue_ready;
+      if (x_valid_i) begin
+        cvxif_req_o.x_issue_valid          = x_valid_i;
+        cvxif_req_o.x_issue_req.instr      = x_off_instr_i;
+        cvxif_req_o.x_issue_req.id         = fu_data_i.trans_id;
+        cvxif_req_o.x_issue_req.rs[0]      = fu_data_i.operand_a;
+        cvxif_req_o.x_issue_req.rs[1]      = fu_data_i.operand_b;
+        if (cvxif_pkg::X_NUM_RS == 3)
+          cvxif_req_o.x_issue_req.rs[2]    = fu_data_i.imm;
```
zarubaf, line 43:
```suggestion
          cvxif_req_o.x_issue_req.rs[2]    = fu_data_i.imm;
        end
```

core/cvxif_fu.sv
```diff
@@ -0,0 +1,72 @@
...
+      if (~cvxif_resp_i.x_issue_resp.accept && cvxif_req_o.x_issue_valid && cvxif_resp_i.x_issue_ready) begin
+          x_trans_id_o          = cvxif_req_o.x_issue_req.id;
+          x_result_o            = 0;
+          x_valid_o             = 1;
+          x_exception_o.cause   = riscv::ILLEGAL_INSTR;
+          x_exception_o.valid   = 1;
+          x_exception_o.tval    = cvxif_req_o.x_issue_req.instr;
+          x_we_o                = '0;
+      end
+      else begin
```
zarubaf, line 61:
```suggestion
      end else begin
```

core/decoder.sv
```diff
@@ -1075,7 +1085,7 @@ module decoder import ariane_pkg::*; (
             // check here if we decoded an invalid instruction or if the compressed decoder already decoded
             // a invalid instruction
             if (illegal_instr || is_illegal_i) begin
-                instruction_o.ex.valid = 1'b1;
+                if (~CVXIF_PRESENT) instruction_o.ex.valid = 1'b1;
```
zarubaf, line 1088:
```suggestion
                if (!CVXIF_PRESENT) instruction_o.ex.valid = 1'b1;
```

core/ex_stage.sv
```diff
@@ -319,6 +330,33 @@ module ex_stage import ariane_pkg::*; #(
         .pmpaddr_i
     );
 
+    generate
```
zarubaf, line 333: Same as above
```suggestion
```

core/ex_stage.sv
```diff
@@ -319,6 +330,33 @@ module ex_stage import ariane_pkg::*; #(
...
+        if (CVXIF_PRESENT) begin : cvxif_gen
```
zarubaf, line 334:
```suggestion
        if (CVXIF_PRESENT) begin : gen_cvxif
```

core/ex_stage.sv
```diff
@@ -319,6 +330,33 @@ module ex_stage import ariane_pkg::*; #(
...
+                .x_off_instr_i,
+                .x_trans_id_o,
+                .x_exception_o,
+                .x_result_o,
+                .x_valid_o,
+                .x_we_o,
+                .cvxif_req_o,
+                .cvxif_resp_i
+            );
+        end else begin : no_cvxif_gen
```
zarubaf, line 352:
```suggestion
        end else begin : gen_no_cvxif
```

core/include/cvxif_pkg.sv
```diff
@@ -0,0 +1,110 @@
...
+    logic  [X_RFW_WIDTH-1:0]  data;
+    logic  [4:0]              rd;
+    logic                     we;
+    logic                     exc;
+    logic  [5:0]              exccode;
+  } x_result_t ;
+
+  typedef struct packed {
+      logic                 x_compressed_valid;
+      x_compressed_req_t    x_compressed_req;
```
zarubaf, line 87: The indentation is somehow different than the other structs above.

core/issue_read_operands.sv
```diff
@@ -296,6 +312,33 @@ module issue_read_operands import ariane_pkg::*; #(
       end
     end
 
+    generate
```
zarubaf, line 315:
```suggestion
```

core/issue_read_operands.sv
```diff
@@ -296,6 +312,33 @@ module issue_read_operands import ariane_pkg::*; #(
...
+              cvxif_valid_q  <= 1'b0;
+              cvxif_off_instr_q  <= 32'b0;
+          end
+          if (!issue_instr_i.ex.valid && issue_instr_valid_i && issue_ack_o) begin
+              case (issue_instr_i.fu)
+              CVXIF: begin
+                  cvxif_valid_q       <= 1'b1;
+                  cvxif_off_instr_q   <= orig_instr;
+              end
+              default:;
```
zarubaf, line 331: I think case items should be 4 spaces intended.
```suggestion
              CVXIF: begin
                  cvxif_valid_q       <= 1'b1;
                  cvxif_off_instr_q   <= orig_instr;
              end
              default:;
```

core/issue_read_operands.sv
```diff
@@ -296,6 +312,33 @@ module issue_read_operands import ariane_pkg::*; #(
...
+      if (CVXIF_PRESENT) begin
+        always_ff @(posedge clk_i or negedge rst_ni) begin
+          if (!rst_ni) begin
+              cvxif_valid_q  <= 1'b0;
+              cvxif_off_instr_q  <= 32'b0;
+          end else begin
+              cvxif_valid_q  <= 1'b0;
+              cvxif_off_instr_q  <= 32'b0;
+          end
+          if (!issue_instr_i.ex.valid && issue_instr_valid_i && issue_ack_o) begin
```
zarubaf, line 325: There is something wrong with the logic here. I'd suggest you split this in a proper comb and sequential block. 

I don't think that this is the logic you've intended, together with the reset this will give a strange priority.

core/scoreboard.sv
```diff
@@ -344,12 +348,13 @@ module scoreboard #(
...
     .data_i  ( rs_data     ),
     .gnt_i   ( 1'b1        ),
-    .req_o   ( rs3_valid_o ),
+    .req_o   ( rs3_valid   ),
     .data_o  ( rs3         ),
     .idx_o   (             )
   );
 
-  assign rs3_o = rs3[ariane_pkg::FLEN-1:0];
+  assign rs3_o = rs3[(ariane_pkg::NR_RGPR_PORTS == 3 ? riscv::XLEN : ariane_pkg::FLEN)-1:0];
```
zarubaf, line 356: I think this block would be more readable if written like this:
```suggestion
  if (ariane_pkg::NR_RGPR_PORTS == 3) begin : gen_gp_three_port
      assign rs3_o = rs3[riscv::XLEN-1:0];
  else begin : gen_fp_three_port
      assign rs3_o = rs3[ariane_pkg::FLEN-1:0];
  end
```

corev_apu/cvxif_example/cvxif_example_coprocessor.sv
```diff
@@ -0,0 +1,153 @@
...
+
+  fifo_v3 #(
+      .FALL_THROUGH  ( 1         ), //data_o ready and pop in the same cycle
+      .DATA_WIDTH    ( 64        ),
+      .DEPTH         ( 8         ),
+      .dtype         ( x_issue_t )
+    ) fifo_commit_i (
+    .clk_i           ( clk_i      ),
+    .rst_ni          ( rst_ni     ),
+    .flush_i         ( 1'b0       ),
```
zarubaf, line 117: Indentation is off somehow

corev_apu/cvxif_example/instr_decoder.sv
```diff
@@ -0,0 +1,49 @@
...
+
+module instr_decoder import cvxif_pkg::*; #(
+   parameter int                                  NbInstr             = 1,
+   parameter cvxif_instr_pkg::copro_issue_resp_t  CoproInstr[NbInstr] = {0}
+)
+(
+    input   logic              clk_i,
+    input   x_issue_req_t      x_issue_req_i,
+    output  x_issue_resp_t     x_issue_resp_o
+
```
zarubaf, line 18:
```suggestion
```

zarubaf (changes requested):
Thanks this looks really good already. I've left a couple of nits and one block of logic I don't think is as you want it to behave IMHO :-) 

One additional thing I'd recommend is to check that the `NR_RGPR_PORTS` parameter has a sane value (i.e, max three and three only if CVX is enabled? you know it best). And check this with a proper `initial` assertion somewhere.

Gchauvon:
I added 2 commits to be more align on style guideline. 
I also modified the block in `issue_read_operands.sv` which was indeed wrong. 
I added an initial assertion into `issue_read_operands.sv` to check the value for `NR_RGPR_PORTS` which is 2 when CVXIF is disable and can be either 2 or 3 when it is enable.

zarubaf (approved):
Thanks a lot! LGTM

PR #785, opened 2022-01-11
 This commit to address issue421 for DIV and REM operation of serial …

nathpranay:
This PR is addressed to solve issue#421 : Erronous results for -1/1 while performing DIV and REM operation.


zarubaf:
I needed to re-base on the new directory structure so I've opened a new PR. Sorry that the authorship got lost in the process. New PR in #796.

PR #791, opened 2022-01-14
AXI and common_cells upgrade

micprog:
- Bump `common_cells` to `v1.23.0`
- Bump `axi` to `v0.31.0`
- Remove `axi_node` dependency, replace with `axi_xbar` from `axi` repository
- Slightly bump `register_interface` for `axi` compatibility


zarubaf (approved):
That is great @micprog - thanks for keeping the changes so clean and minimal, that is really appreciated.

PR #795, opened 2022-01-18
Fix flists after commit #791

Gchauvon:
Hello, 
Just a minor fix.
A file was missing in the Flists used by core-v-verif repository.


zarubaf (approved):
LGTM

PR #798, opened 2022-01-21
Flist.cv64a6_imafdc_sv39_gate, update LIB_VERILOG path

yanicasa:
Modification on the flist gate "Flist.cv64a6_imafdc_sv39_gate" to clean post synthesis simulation jobs.
Most of the time the tech files are in a fixed path on the server and not relative to the project. With the current prefix this forced us to do a workaround.


zarubaf (approved):
👍 thanks for splitting this up

PR #799, opened 2022-01-21
ariane_xilinx: Fix xbar clock

niwis:
**Issue**
On FPGA, Ariane deadlocks immediately after loading the bitstream while attempting to fetch from bootrom.

**Methodology**
Hundreds of ILA probes in and around the crossbar.

**Solution**
Fix the name of the clock net provided to the crossbar.


zarubaf:
😅 thanks, when did that happen? In the AXI XBar refactoring?

PR #801, opened 2022-01-24
Makefile: add missing incdir for VCS (fix #791)

ASintzoff:
Signed-off-by: André Sintzoff <andre.sintzoff@thalesgroup.com>


zarubaf (approved):
👍

PR #802, opened 2022-01-24
Fixed issue counter in order to leverage the full scoreboad length.

OttG:
The scoreboard issue counter `issue_cnt_q` and `issue_cnt_n` have length equal to `log2(NUM_ENTRIES) - 1` where `NUM_ENTRIES` is the number of scoreboard entries. In the standard case where `NUM_ENTRIES = 8` the counters are 3 bits long meaning that they can count from 0 to 7 making it impossible to use the full scoreboard (always leaving one spot free).


zarubaf (approved):
Ui, nice catch! 🙏

PR #806, opened 2022-01-26
Remove debug module from `devicetree`

noah95:
This changes two things in the devicetree:

1. Remove the debug-module. It causes an `hwirq 0xffff is too large` kernel panic in the mainline kernel (fault occurred with `v5.14.0`)
2. Change the compatible string of the uart to one that activates the FiFo. Tested with linux-v5.16.0 in the occamy system (see `8250_port.c` in kernel sources)


niwis:
Cheers @noah95! Could you also regenerate cva6/corev_apu/fpga/src/bootrom/bootrom*.[h|sv]? (`make all` in cva6/corev_apu/fpga/src/bootrom/). Then I can give this a quick try on a Genesys2.

zarubaf (approved):
LGTM and thanks. @niwis if you could ping here whether the boot still succeeds on the `core-v-apu` platform, then we can merge.

P.S.: @noah95 do you think you can fix the failing Eclipse Committer agreement check. Requires that your Github user name is associated with a signed Eclipse Committer Agreement (ETH has on in place, but that needs to be associated with your user). Also all commits need to be "signed" using `commit -s` (might require rebasing and signing).

noah95:
@zarubaf might need to re-approve running workflows. Commits should be signed and authored with the correct mail

niwis:
Looks good to me, Linux boots on Genesys2

PR #813, opened 2022-02-04
ariane_testharness/ariane_xilinx: Fix AXI ID width

niwis:
- Fix the AXI ID width for the CLINT (previously `4`, now `5`)
- Parametrise the CLINT's AXI types
- Deprecate `axi_[master|slave]_connect` and move to AXI assign macros, as they allow arbitrary AXI types


zarubaf (approved):
Thanks for this nice clean-up. LGTM!

PR #818, opened 2022-02-08
Add missing sources for questa simulation

andreaskuster:
Simulating the current master with Questasim leads to errors due to `cvxif_example` being enabled, but the sources are missing in the Makefile. This PR fixes that.


zarubaf (approved):
Thanks!

PR #820, opened 2022-02-10
Fix minor PMP uncompliance

flaviens:
Hi!

The latest RISC-V privileged spec says, regarding the PMP configuration registers:
> The R, W, and X fields form a collective WARL field for which the combinations with R=0 and W=1 are reserved.

It seems that Ariane does not comply with this rule: I tested writing a PMP config register with W=1 and R=0 on Ariane, Ibex and BOOM, and only Ariane seems to ignore this spec.
This PR aims at fixing this uncompliance.

Thank you for your time!


zarubaf:
@Moschn probably knows more about it :-)

Moschn:
Maybe I do not understand the word `reserved` correctly, but here is how I interpret it: A configuration that is reserved is not allowed from a software standpoint. So the HW can do as it likes if such a configuration is presented. Obviously reserved also means that his configuration should not be used in a custom extension.

If I now think about it, I might be wrong since it is supposed to be WARL. So reading a reserved configuration should never happen as it cannot be legal.

If my interpretation of `reserved` is correct, then the current implementation should be compliant. If I am incorrect then you are right, and we should fix this.

However, you should probably not implement this using a bitwise AND (`&`). Instead, why don't you just add the check to the existing if? I would assume this also breaks the CI.

Moschn:
How about you change it like this:
```diff
diff --git a/core/csr_regfile.sv b/core/csr_regfile.sv
index c27528b..d2af08b 100644
--- a/core/csr_regfile.sv
+++ b/core/csr_regfile.sv
@@ -1156,7 +1156,7 @@ module csr_regfile import ariane_pkg::*; #(
             for(int i = 0; i < 16; i++) begin
                 if(i < NrPMPEntries) begin
                     // We only support >=8-byte granularity, NA4 is disabled
-                    if(pmpcfg_d[i].addr_mode != riscv::NA4) 
+                    if(pmpcfg_d[i].addr_mode != riscv::NA4 && !(pmpcfg_d[i].access_type.r == '0 && pmpcfg_d[i].access_type.w == '1) 
                         pmpcfg_q[i] <= pmpcfg_d[i];
                     else
                         pmpcfg_q[i] <= pmpcfg_q[I];
```

I think it would be more readable than what you proposed. Also maybe add some comments to describe why this is there :)

flaviens:
Good catch! Then let's do what you proposed!

Moschn:
Ups, I also just opened a new PR proposing this exact change. See #823

PR #821, opened 2022-02-10
tb/cache: Fix regressions

niwis:
Due to recent changes, the modular cache test benches are unable to run. Fix them by
- updating the parameter interface of `std_nbdcache`,
- updating the dependencies and include directories,
- preventing an overflow during `range_check`.


zarubaf (approved):
Awesome! Thanks

PR #823, opened 2022-02-10
Fix reserved CSR write

Moschn:
Fix illegal write to PMPCFG. Reported by Flavien Solt (@flaviens)

Supersedes #820 

I also found a way to simplify and clarify the code for the CSR read and write of PMPADDR. Should be functionally the same as before.


zarubaf (approved):
Thanks @Moschn. Do you think you could fix the failing Eclipse committer check (I think you have an account there, right?). For sure you are covered as part of ETH.

Moschn:
I committed with a different email address. Re-signed the eclipse thing with the new email

PR #835, opened 2022-03-09
Remove an unused argument in bootrom functions

atkarim:
The argument for the boot partition size is unused, since we use the GPT partition table to deduced it now.
This is a remnant of an older version of the code : https://github.com/openhwgroup/cva6/blob/v4.2.0/fpga/src/bootrom/src/gpt.c


zarubaf:
Awesome thanks! Can you please re-generate the bootroms and check them in as part of this PR as well?

zarubaf (approved):
👍

PR #838, opened 2022-03-11
:bug: wb_dcache: Fix unaligned SC return data bug

colluca:
When a store-conditional fails due to a missing reservation, by specification it must return a non-zero value. In addition, the non-zero value must be a 1, other non-zero values being reserved. Previously a value of `64'b1` was returned by the `axi_adapter`. However, when an atop's address is non-64bit-aligned, the return data has to be realigned. This realignment causes the lower half-word being lost, and the return datum taking a value of `'0`.

This commit fixes the bug by returning 1 from the `axi_adapter` when the request is 64-bit and placing a 1 in both upper and lower halfwords when the request is 32-bit. Externally, the correct halfword is realigned/masked thus returning a value of 1 in all cases. Unfortunately, this requires registering the `size` of the request.


zarubaf (approved):
LGTM

PR #843, opened 2022-03-25
updated csr_regfile to support mandatory mcycle and minstret counter also added support to make perf_counters optional…

nathpranay:
…as per specification for XLEN=32, also made perf_counters optional 
@zarubaf 

Signed-off-by: nathpranay <pranay.kantinath@thalesgroup.com>


core/cva6.sv
```diff
@@ -603,6 +605,11 @@ module cva6 import ariane_pkg::*; #(
     .eret_i            ( eret                   ),
     .resolved_branch_i ( resolved_branch        )
   );
+ end
```
zarubaf, line 608: Remove new line

core/cva6.sv
```diff
@@ -603,6 +605,11 @@ module cva6 import ariane_pkg::*; #(
...
+ else begin 
+    assign data_perf_csr    = '0;
+ end
+ endgenerate
```
zarubaf, line 612:
```suggestion
```

core/csr_regfile.sv
```diff
@@ -242,11 +242,15 @@ module csr_regfile import ariane_pkg::*; #(
...
-                riscv::CSR_MCYCLE:             csr_rdata = cycle_q;
-                riscv::CSR_MINSTRET:           csr_rdata = instret_q;
+                riscv::CSR_MCYCLE:             csr_rdata = riscv::IS_XLEN32 ? cycle_q[31:0] : cycle_q;
+                riscv::CSR_MCYCLE_H:           csr_rdata = riscv::IS_XLEN32 ? cycle_q[63:32]: '0;
+                riscv::CSR_MINSTRET:           csr_rdata = riscv::IS_XLEN32 ? instret_q[31:0] : instret_q;
+                riscv::CSR_MINSTRET_H:         csr_rdata = riscv::IS_XLEN32 ? instret_q[63:32]:'0;
                 // Counters and Timers
-                riscv::CSR_CYCLE:              csr_rdata = cycle_q;
-                riscv::CSR_INSTRET:            csr_rdata = instret_q;
+                riscv::CSR_CYCLE:              csr_rdata = riscv::IS_XLEN32 ? cycle_q[31:0] : cycle_q;
```
zarubaf, line 250: What does the standard say about `CYCLE_H` In 64 bit mode? Should it throw an illegal instruction exception?

zarubaf (changes requested):

core/csr_regfile.sv
```diff
@@ -564,8 +568,10 @@ module csr_regfile import ariane_pkg::*; #(
                     mip_d = (mip_q & ~mask) | (csr_wdata & mask);
                 end
                 // performance counters
-                riscv::CSR_MCYCLE:             cycle_d     = csr_wdata;
-                riscv::CSR_MINSTRET:           instret     = csr_wdata;
+                riscv::CSR_MCYCLE:             cycle_d[31:0]  = csr_wdata;
```
Moschn, line 571: I do not typically leave any comments, but I was just wondering what would happen on a 64bit configuration if the write value is larger than 32 bits. To my understanding, only the lower 32 bits would be overwritten, the upper bits would just remain. 

Most probably I am wrong though so feel free to ignore this comment.

core/csr_regfile.sv
nathpranay: As per riscv-privileged-v1.10.pdf [ Chapter 2] : " Attempts to access a non-existent CSR raise an illegal instruction exception. Attempts to access a CSR without appropriate privilege level or to write a read-only register also raise illegal instruction exceptions. A read/write register might also contain some bits that are read-only, in which case
writes to the read-only bits are ignored." As per spec, all this _H CSR are RV32 only. In that way, in 64 Bit mode, it should throw illegal instruction.

core/csr_regfile.sv
nathpranay: Hi,
As per spec:
0xB00 MRW mcycle Machine cycle counter.
0xB02 MRW minstret Machine instructions-retired counter.
0xB03 MRW mhpmcounter3 Machine performance-monitoring counter.
0xB04 MRW mhpmcounter4 Machine performance-monitoring counter.
.
.
.
0xB1F MRW mhpmcounter31 Machine performance-monitoring counter.
0xB80 MRW mcycleh Upper 32 bits of mcycle, RV32I only.
0xB82 MRW minstreth Upper 32 bits of minstret, RV32I only.
0xB83 MRW mhpmcounter3h Upper 32 bits of mhpmcounter3, RV32I only.
0xB84 MRW mhpmcounter4h Upper 32 bits of mhpmcounter4, RV32I only.
.
.
.
0xB9F MRW mhpmcounter31h Upper 32 bits of mhpmcounter31, RV32I only

When XLEN=32, 0xB00(mcycle) and 0xB80 (mcycleh) together makes 64bit mcycle counter. Reading 64 bit mcycle counter will be done by reading 2 sepearate 32 bit (0xB00 and 0xB80). Writing also shall be done similarly.
For XLEN=32 or 64 or 128 the performance counters width is 64 bit.

core/csr_regfile.sv
zarubaf: So I think we need to make sure we throw an illegal instruction when trying to access the 32 only regs. Right now they would read zero.

core/csr_regfile.sv
Moschn: I am not questioning the size of `mcycle`. I am just wondering why only the lower 32 bits are writable. I understand that for XLEN=32, one can write to the upper 32bits by writing to `mcycleh`. However, in XLEN=64 the `mcycleh` is not available.   
We should be able to write to all 64bits of the cycle register for both XLEN=32 and XLEN=64 right?

core/csr_regfile.sv
nathpranay: Yes, i agree with you. For pmpcfg1 and pmpcfg3 access while XLEN=64, also should throw an illegal instruction?

0x3A0 MRW pmpcfg0 Physical memory protection configuration.
0x3A1 MRW pmpcfg1 Physical memory protection configuration, RV32 only.
0x3A2 MRW pmpcfg2 Physical memory protection configuration.
0x3A3 MRW pmpcfg3 Physical memory protection configuration, RV32 only

core/csr_regfile.sv
```diff
@@ -564,8 +568,10 @@ module csr_regfile import ariane_pkg::*; #(
                     mip_d = (mip_q & ~mask) | (csr_wdata & mask);
                 end
                 // performance counters
-                riscv::CSR_MCYCLE:             cycle_d     = csr_wdata;
-                riscv::CSR_MINSTRET:           instret     = csr_wdata;
+                riscv::CSR_MCYCLE:             cycle_d[31:0]  = csr_wdata; 
+                riscv::CSR_MCYCLE_H:           cycle_d[63:32] = riscv::IS_XLEN32 ? csr_wdata : '0 ; 
+                riscv::CSR_MINSTRET:           instret[31:0]  = csr_wdata;
```
Gchauvon, line 573: Shouldn't it be `instret_d` instead of `instret` ? 
Otherwise the register will not be updated because `instret_q` takes the value of `instret_d` which is not updated with `instret` after those lines.

```suggestion
                riscv::CSR_MINSTRET:           instret_d[31:0]  = csr_wdata; 
```

core/csr_regfile.sv
```diff
@@ -564,8 +568,10 @@ module csr_regfile import ariane_pkg::*; #(
                     mip_d = (mip_q & ~mask) | (csr_wdata & mask);
                 end
                 // performance counters
-                riscv::CSR_MCYCLE:             cycle_d     = csr_wdata;
-                riscv::CSR_MINSTRET:           instret     = csr_wdata;
+                riscv::CSR_MCYCLE:             cycle_d[31:0]  = csr_wdata; 
+                riscv::CSR_MCYCLE_H:           cycle_d[63:32] = riscv::IS_XLEN32 ? csr_wdata : '0 ; 
+                riscv::CSR_MINSTRET:           instret[31:0]  = csr_wdata; 
+                riscv::CSR_MINSTRET_H:         instret[63:32] = riscv::IS_XLEN32 ? csr_wdata : '0 ;
```
Gchauvon, line 574: Same reason as previous line
```suggestion
                riscv::CSR_MINSTRET_H:         instret_d[63:32] = riscv::IS_XLEN32 ? csr_wdata : '0 ;
```

zarubaf:
I think the same functionality has been merged as part of #847

PR #847, opened 2022-03-30
Add support for "high" counter CSRs in 32-bit mode

spersvold:
In 32bit mode MCYCLEH, MINSTRETH, CYCLEH, TIMEH and INSTRETH are
used to return the most significant 32-bit value of the counters
which are now always 64-bit wide.

Signed-off-by: Steffen Persvold <spersvold@gmail.com>

### DIFF
--- a/core/csr_regfile.sv
+++ b/core/csr_regfile.sv
@@ -136,8 +136,8 @@ module csr_regfile import ariane_pkg::*; #(
 
     logic        wfi_d,       wfi_q;
 
-    riscv::xlen_t cycle_q,     cycle_d;
-    riscv::xlen_t instret_q,   instret_d;
+    logic [63:0] cycle_q,     cycle_d;
+    logic [63:0] instret_q,   instret_d;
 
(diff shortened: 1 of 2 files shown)


zarubaf (changes requested):
Looks good to me in general, I only think the write CSR portion is missing :/ Could you add that as well?

zarubaf:
Ah, and can you make sure the ECA is passing? I think you need to sign up with Eclipse to make this pass :-) Let me know if you need more advice on that.

spersvold:
> Looks good to me in general, I only think the write CSR portion is missing :/ Could you add that as well?

Indeed! I Forgot the MCYCLE/MINSTRET registers are writable (the non-M variants are not). I've added this now. Thanks :)

spersvold:
> Ah, and can you make sure the ECA is passing? I think you need to sign up with Eclipse to make this pass :-) Let me know if you need more advice on that.

Should be fixed now.

zarubaf (approved):
lgtm

PR #849, opened 2022-04-06
Dev dcache 32bits

sjthales:
This is wt dcache adaptation in 32b, it hasn't been done yet.

That allows to reduce the amount logic resources.

It has been tested on Linux in 32b and 64b version.


zarubaf (approved):
Only nits, lgtm in general. Very nice fixing of the hardcoded path, thanks.

JeanRochCoulon (approved):
I agree with Florian's comment. Other than that the modification seems ok to me. Thanks

sjthales:
Modifications performed tested on linux 32 and 64 b.

PR #870, opened 2022-04-29
wt_dcache_wbuffer.sv: remove init for user

Gchauvon:
Remove init for user field of the register.

@JeanRochCoulon can you look at that ?

Guillaume


Gchauvon:
Yes, I have.

PR #908, opened 2022-06-09
Fix exception type on PMP check during PTW

### LINKED ISSUE #906: [Bug Report] Incorrect exception type of PMP violation during address translation
Our co-simulation framework found that the exception type of address translation PMP violation is incorrect.

In the following test case, we let a non-leaf (level 2) PTE out of the PMP range.
Next, we will perform a store operation with a special virtual address, during translation processor will try to access the PTE that does not match any PMP.
cva6 throws a load access fault, while spike throws a store access fault.

```
[cva6]  Exception @    119000, PC: 000000008000036c, Cause: Load Access Fault,
[cva6]                                 tval: 0000000080102008
[spike] core   0: 0x000000008000036c (0x00b52023) sw      a1, 0(a0)
[spike] core   0: exception trap_store_access_fault, epc 0x000000008000036c
[spike] core   0:           tval 0x0000000040201010
[spike] core   0: 0x0000000080000004 (0x34302f73) csrr    t5, mtval
[error] WDATA SIM 0000000040201010, DUT 0000000080102008
[error] check board clear 30 error
```

According to riscv-privileged specification:
> If accessing pte violates a PMA or PMP check, raise an access-fault exception corresponding to the original access type.

Hence, cva6 should throw a store access fault.

[cva6-9.zip](https://github.com/openhwgroup/cva6/files/8862145/cva6-9.zip)

> @ProjectDimlight helps reproduce the problem

Moschn:
Fixes #906

According to the spec:
> If accessing pte violates a PMA or PMP check, raise an access-fault
> exception corresponding to the original access type.

Found by @Phantom1003 and @ProjectDimlight

Signed-off-by: Moritz Schneider <moritz.schneider@inf.ethz.ch>

### DIFF
--- a/core/mmu_sv39/mmu.sv
+++ b/core/mmu_sv39/mmu.sv
@@ -397,8 +397,12 @@ module mmu import ariane_pkg::*; #(
                 if (ptw_access_exception) begin
                     // an error makes the translation valid
                     lsu_valid_o = 1'b1;
-                    // the page table walker can only throw page faults
-                    lsu_exception_o = {riscv::LD_ACCESS_FAULT, {{riscv::XLEN-riscv::PLEN{1'b0}}, ptw_bad_paddr}, 1'b1};
+                    // Any fault of the page table walk should be based of the original access type
+                    if (lsu_is_store_q) begin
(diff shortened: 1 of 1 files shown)


zarubaf (approved):
Thanks, good catch!

JeanRochCoulon (approved):

Issue #876, opened 2022-05-15
[Bug Report] incorrect SFENCE.VMA decoder


Phantom1003:
Hi, we are currently doing some co-simulation with cva6 and spike, and we found the decoder has an incorrect behavior when meeting a sfence.vma with non-zero rd field.

According to the ISA Specification (Volume II: RISC-V Privileged Architectures V20211203 Page 152) for the SFENCE.VMA format, instr[11:7] should be 5'b00000. 

When modified instr[11:7] to 5'b00001. cva6 treats this instruction as SFENCE.VMA as well. No exception occurred. The implementation is missing a check for this field. 

https://github.com/openhwgroup/cva6/blob/44a89b9cd408b869fb4ed2c0cec35e1561588e37/core/decoder.sv#L152-L165

In the following test case, there is an invalid sfence.vma at 0x80000190, whose rd field is 1, cva6 execute it as normal instruction, while spike throws an excaption.
```
  [cva6] 532890ns    26637 M 000000008000018c 0 12000073 sfence.vma 
  [spike] core   0: 0x000000008000018c (0x12000073) sfence.vma zero, zero                         
  [cva6] 534950ns    26740 M 0000000080000190 0 120000f3 sfence.vma    
  [spike] core   0: 0x0000000080000190 (0x120000f3) unknown
  [spike] core   0: exception trap_illegal_instruction, epc 0x0000000080000190
  [spike] core   0:           tval 0x0000000000000000
```

[cva6-0.zip](https://github.com/openhwgroup/cva6/files/8694534/cva6-0.zip)

> @LuminaDCIX helps reproduce the problem

Phantom1003:
cc to @zarubaf

PR #921, opened 2022-06-23
fix sfence.vma decoder

Phantom1003:
from https://github.com/openhwgroup/cva6/issues/876, this patch passed the testcase in the report.
The original design reset the illegal_instr during privilege check, use the previous value instead 0.


zarubaf (approved):
Great, thanks soo much for finding and fixing those! Greatly appreciated!

JeanRochCoulon:
LGTM

Issue #899, opened 2022-06-03
[Bug Report] incorrect DRET decoder


Phantom1003:
Our co-simulation framework found the decoder has an incorrect behavior when execute a `dret` with non-zero rd field.

According to the Debug Specification Version 1.0.0:
> To resume execution, the debug module sets a flag which causes the hart to execute a `dret`. `dret` is an instruction that only has meaning while in Debug Mode and not executing from the Program Buffer. Its recommended encoding is 0x7b200073.

When modified instr[11:7] to 5'b00001. cva6 treats this invalid `dret` as normal `dret` as well. No exception occurred. The implementation is missing a check for this field.

In the following test case, there is an invalid `dret` at 0x8000019c, whose rd field is 1, cva6 execute it as normal instruction, while spike throws an excaption.

```
[spike] core   0: 0x000000008000019c (0x7b200273) unknown
[spike] core   0: exception trap_illegal_instruction, epc 0x000000008000019c
[spike] core   0:           tval 0x000000007b200273
[spike] core   0: 0x0000000080000004 (0x00000f17) auipc   t5, 0x0
[error] PC SIM 0000000080000004, DUT 000000008000019c
[error] INSN SIM 00000f17, DUT 7b200273
[CJ] Commit Failed
[cva6]     786ns      771 D 000000008000019c 0 7b200273 INVALID   // CVA6 continues to execute 
[cva6]    1068ns     1053 S 000000008000017c 0 0ff0000f fence         // subsequent instructions
```

We believe is the debug mode checking reset the correct signal:
https://github.com/openhwgroup/cva6/blob/909d85a56cc5ace65765a63d7ed56b7ac2026f99/core/decoder.sv#L134-L135

[cva6-3.zip](https://github.com/openhwgroup/cva6/files/8833048/cva6-3.zip)
> @LuminaDCIX helps reproduce the problem

zarubaf:
Ah nice catch. Would you mind preparing a PR for that? This would be greatly appreciated. Thanks!

PR #922, opened 2022-06-23
fix dret decoder

Phantom1003:
from https://github.com/openhwgroup/cva6/issues/899, this patch passed the testcase in the report.
debug mode check reset the illegal_instr signal, use its previous value instead of 0.


zarubaf (approved):
Awesome, thanks!

JeanRochCoulon:
Thanks !

PR #930, opened 2022-07-04
Adding a Bug issue template 

eyssartk:
This template provides an automatic title and label for bug issues.


zarubaf (approved):
Thanks! Much needed imho!

PR #943, opened 2022-08-18
perf_counters: Explicit typecast for comparison

niwis:
Comparing a logic vector to an enum causes some tools (e.g. Questasim) to throw a mismatch error. Fix this by explicitly typecasting.


JeanRochCoulon:
@Gchauvon or @yanicasa, could you approve ?

zarubaf (approved):
👍

PR #945, opened 2022-08-25
rvfi_tracer.sv: Fix log format of FP regs written by compressed insns.

zchamski:
Previously, the names of FP registers written by compressed instructions were printed as integer register names.
This change applies the FP-specific printing format to destination regs of both uncompressed *and* compressed
instructions that set an FP register.
 
NOTE: CVA6 supports only XLEN==FLEN for compressed FP load insns, unlike RV ISA spec v2.2 that allows FLEN
to be 64 bits while XLEN==32.  This means that on CVA6, C.FLD/C.FLDSP insns are recognized only on CV64A6
and C.FLW/C.FLWSP only on CV32A6.

Files changed:

* corev_apu/tb/rvfi_tracer.sv: Recognize compressed insns that set FP regs
  so that all FP register commits are correctly formatted.

Signed-off-by: Zbigniew Chamski <zbigniew.chamski@thalesgroup.com>

### DIFF
--- a/corev_apu/tb/rvfi_tracer.sv
+++ b/corev_apu/tb/rvfi_tracer.sv
@@ -41,7 +41,8 @@ module rvfi_tracer #(
         // Destination register information
         $fwrite(f, "%h 0x%h (0x%h)",
           rvfi_i[i].mode, pc64, rvfi_i[i].insn);
-        // Decode instruction to know if destination register is FP register
+        // Decode instruction to know if destination register is FP register.
+        // Handle both uncompressed and compressed instructions.
         if ( rvfi_i[i].insn[6:0] == 7'b1001111 ||
              rvfi_i[i].insn[6:0] == 7'b1001011 ||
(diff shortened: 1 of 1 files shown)


JeanRochCoulon:
@zchamski    We do not used to add so much comments in RTL from the beginning. I suggest to remove the comments from the RTL. The comments can be added to the PR discussion and to the design specification (when it will exist!).

zchamski:
Hi @JeanRochCoulon , the comments in RTL have been shrunk as requested.  I updated also the description of the PR, splitting it into chapters and adding a reference to the test set that was submitted as PR against https://github.com/openhwgroup/core-v-verif/cva6/dev.

zchamski:
Just as a confirmation, @spidugu444 's testing confirms the issue is fixed in the configuration where it was first identified
PR945-discussion

spidugu444:
@zchamski with your fix in CVA6 RTL  64-bit floating point ABI issue has been resolved.

PR #959, opened 2022-09-18
Cvvdev/dev/rvfi

JeanRochCoulon:


JeanRochCoulon:
With the help of this present PR RVFI reports information on memory accesses and rs1/rs2 data values, could you confirm me RVFI works for you use cases, that is to say:
- Test termination - @zchamski 
- ISACOV - @AyoubJalali
Wait for your approval for merging.
Thanks

PS: @ASintzoff

AyoubJalali:
@JeanRochCoulon I checked what you did, I get the values of RS1 & RS2 but are not the right values, for example :
1- for some instructions like **addi** (it has no source register 2) and I get the value of rs2 instead of rs1 (in addi instruction no rs2 register).
2- **auipc** instruction don't use rs1 or rs2 but still i get values for the rs1 & rs2 (actually i notice that i get the value of the pc in rs1).
3- For the memory signals i get values, but i didn't check the correctness of the values.

AyoubJalali:
@JeanRochCoulon I verified the memory signals, I compare what i get with spike trace, and it seems good for.

zchamski:
@JeanRochCoulon , the RVFI from commit [4f2b1b4](https://github.com/openhwgroup/cva6/pull/959/commits/4f2b1b42f32825b9bfb156d61c7ce1f219d6addc) works with my candidate "orderly test termination" changeset.  Correct operation was confirmed on Verilator-generated waveforms in both 64b and 32b.   Please note that the orderly termination conidition relies only on the values of `mem_addr`, `mem_wdata` and `mem_wmask`, with `pc_rdata` used as an informative guide during waveform debugging.

AyoubJalali:
> @JeanRochCoulon I checked what you did, I get the values of RS1 & RS2 but are not the right values, for example : 1- for some instructions like **addi** (it has no source register 2) and I get the value of rs2 instead of rs1 (in addi instruction no rs2 register). 2- **auipc** instruction don't use rs1 or rs2 but still i get values for the rs1 & rs2 (actually i notice that i get the value of the pc in rs1). 3- For the memory signals i get values, but i didn't check the correctness of the values.

@JeanRochCoulon I checked a second time and it seems good for me. I made a mistake while decoding these instructions.
For me RVFI seems good for me.

PR #967, opened 2022-09-29
Adding Access Region exception feature

### LINKED ISSUE #879: [Bug Report] Issue with memory access
Hi,
We found that cva6 allows LOAD instructions to load value from an address that is outside the DRAM space. According to RISC-V Unprivileged ISA V20190608, page 7, "if an instruction attempts to access memory at an inaccessible address, an exception is raised for the instruction. Vacant locations in the address space are never accessible". However, cva6 did not throw an exception for this case.
We have the triggering test case (.mem file), the RTL trace log, and the spike trace log attached below. We can see that we are trying to load from an address outside DRAM space to register x10 (lbu  a0, -302(sp)). CVA6 is not throwing any exception for this instruction whereas spike is throwing load access fault exception.

Is this a bug in cva6?
[bug_report.zip](https://github.com/openhwgroup/cva6/files/8749872/bug_report.zip)

Gidha-Iftikhar:
This PR closes #879. 

It adds the unsupported feature of access exception i.e when if an instruction attempts to access memory at an inaccessible address, an exception is raised for the instruction. Vacant locations in the address space are never accessible.

A function is_inside_access_region has been added in ariane_pkg.sv, which checks the address is in access range or not. The function is called in mmu unit, where physical address is passed to function and then raise exception if function check fails. 

The changes can be verified by the [access test](https://github.com/riscv-software-src/riscv-tests/blob/master/isa/rv64mi/access.S) which changes the last bit of the address to 1, making it outside the access region, and try to jump at that address. If the feature is enabled, accessing that address should raise instruction access fault.

### DIFF
--- a/core/include/ariane_pkg.sv
+++ b/core/include/ariane_pkg.sv
@@ -53,6 +53,7 @@ package ariane_pkg;
       //
       logic [63:0]                      DmBaseAddress;         // offset of the debug module
       int unsigned                      NrPMPEntries;          // Number of PMP entries
+      logic [63:0]                      AccessRegionLength;
     } ariane_cfg_t;
 
     localparam ariane_cfg_t ArianeDefaultConfig = '{
@@ -76,7 +77,9 @@ package ariane_pkg;
       SwapEndianess:          1'b0,
       // debug
       DmBaseAddress:          64'h0,
-      NrPMPEntries:           8
+      NrPMPEntries:           8,
+      //Access Region
+      AccessRegionLength:     64'hC000_0000
     };
 
     // Function being called to check parameters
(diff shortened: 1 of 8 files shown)


fatimasaleem:
Hi @JeanRochCoulon Can you please take a look at these changes? Thanks.

JeanRochCoulon:
Thank you for your CVA6 interest. Teh PR has been reviewed by @asintzoff @gchauvon

You define a big region from 0 to 0xC0000000. The region is defined as input parameter of cva6 top module, so we can fine tune the value according to the platform.

Why did not take the memory mapping as reference ?  In that way, you will include more unimplemented addresses, also the unimplemented addresses between two memories.
To my mind the memory mapping defined in ariane_soc_pkg.sv should be used as input of the exception generator, and not only a big region.

JeanRochCoulon:
As this PR does not updates from a while, close it

PR #968, opened 2022-10-02
feat: Bump CVA6 for OpenPiton, fix mmu issues

tianrui-wei:
This PR does the following

1. Bump the filelist for OpenPiton for new directory layout
2. Remove AXI Interface for OpenPiton in the top level
3. Fix several issues in MMU discovered during address translation last
   year, the changes in core/mmu_sv39/mmu.sv are a joint effort
   between Jbalkind minho-pulp zarubaf niwis acostillado tianrui-wei 
   (removing github @ to avoid spamming everyone)
4. disable bitmanip by default
5. separate an ariane config package for openpiton synthesis. Some of
   the previous changes makes ariane too big for openpiton, so we need
   to revert these changes
6. Don't increase number of writeback ports (NR_WB_PORTS) when cvxif is
   not enabled

Signed-off-by: Tianrui Wei <tianrui@tianruiwei.com>


zarubaf:
Cool! Is it booting again in OpenPiton with those changes? Would you mind also applying those fixes to the 32-bit MMU. Sorry for the code duplication, that is something that we are trying to fix with templating, @JeanRochCoulon.

zarubaf (approved):
👍

tianrui-wei:
Thanks Florian! OpenPiton is booting Linux correctly, and I've ported the changes to sv32 in [`8b5cbff` (#968)](https://github.com/openhwgroup/cva6/pull/968/commits/8b5cbff74fc2b03c443ea74fb7948e287f18ce0b).

JeanRochCoulon:
Hello @tianrui-wei     The Thales Continuous Integration is not PASS. I will provide you information as soon as it will finish.

tianrui-wei:
Hi @JeanRochCoulon thanks. I have a valid VCS license at Berkeley. Let me know what else you need. You can also reach me at my email if you want to talk in private :)

JeanRochCoulon:
Hello, you can follow the Thales CI status at https://riscv-ci.pages.thales-invia.fr/dashboard/dashboard_cva6.html.
Do you have already used core-v-verif repository? To reproduce the current error, you have to execute the smoke test: "source cva6/regress/smoke-tests.sh" from core-v-verif repository.

tianrui-wei:
Hi @JeanRochCoulon thanks for the pointer. May I ask what version of the riscv compiler you use? I'm getting the following error, which I assume to be compiler version and ISA string mismatch
```
/scratch/riscv-tests/isa/rv32ui/../rv64ui/add.S:14: Error: unrecognized opcode `csrw mepc,t0'
```

JeanRochCoulon:
@zchamski Could you help @tianrui-wei on GCC side ?!   Thanks

zchamski:
> Hi @JeanRochCoulon thanks for the pointer. May I ask what version of the riscv compiler you use? I'm getting the following error, which I assume to be compiler version and ISA string mismatch
> 
> ```
> /scratch/riscv-tests/isa/rv32ui/../rv64ui/add.S:14: Error: unrecognized opcode `csrw mepc,t0'
> ```

Hi @tianrui-wei , the problem you are facing comes from GCC evolution: in January 2022 GCC changed the default spec version of the RISC-V ISA.  Until GCC 11.2 inclusive, the CSR-related instructions were considered part of the base ISA.  Starting with GCC 11.3, the CSR-related instructions are assumed to belong in a dedicated extension (`Zicsr`), so they are not recognized by default.

There are three ways to handle this change without rebuilding the compiler:
* use an older compiler (this is currently the case for CVA6 regression suites: they use GCC 10.2)
* use GCC 11.3 or later and select the `Zicsr` extension on the compiler command line (add `_zicsr` suffix to the ISA name in the`-march=` option.)
* use GCC 11.3 or later and enable the "old" behavior (GCC 11.2, GCC 10...) by adding `-misa-spec=2.2` to the compiler command line. 

See https://github.com/seL4/seL4/issues/749 for a discussion of the reasoning behind this change, including the link to the original mailing list message.

tianrui-wei:
Hi @zchamski @JeanRochCoulon thanks for the explanation. @Jbalkind proposed a fix but didn't solve the thales CI despite passing the smoke tests. Could you check how the error could be reproduced? Thanks!

JeanRochCoulon:
Maybe you did not set the right value in DV_SIMULATORS variable?    
"export DV_SIMULATORS=vcs-uvm,spike" should be executed before smoke tests. It should make smoke tests fail.
Tell me !

Jbalkind:
I worked on this more last night. When the CVXIF_ENABLE change occurred, NR_WB_PORTS was also changed to 5, which significantly increases the size of the scoreboard. We set it back to 4 (as used in OpenPiton already and the default for a long time) when CVXIF_ENABLE is not set. For some reason, this breaks the 32 bit config (test times out). We're thinking to just add a separate config to replace NR_WB_PORTS and set it to 4 for OpenPiton for now, but this is probably worth looking into as it seems that the addition of CVXIF broke this configuration for 32 bit.

JeanRochCoulon:
@Gchauvon, the submitted PR does not work in 32bits config when CV-X-IF is enabled. As CV-X-IF designer, have any idea ?

Gchauvon:
I did look into this. 
It appears that writeback bus assignements are broken when `NB_WB_PORTS==4` 
I have a fix to condition those assignment to the CVXIF_PRESENT parameter. I did a pull request to @tianrui-wei branch complete this pull request.

tianrui-wei:
@Gchauvon Thanks for submitting the PR, I already merged it.
@JeanRochCoulon Can you manually retrigger the CI flow from inside Thales? It seems mergint the PR doesn't trigger the CI.

Gchauvon:
As I was afraid. Synthesis tools didn't like the condtion syntax in port connection. So I did a new PR and create intermediate signals (https://github.com/tianrui-wei/cva6/pull/2). I did a synthesis, it passed. Continuous integration should be ok after this.

tianrui-wei:
Thanks @Gchauvon for submitting the fix. I've already merged your changes into my branch and pushed it here. Hopefully the ci will go through 🤞

tianrui-wei:
@JeanRochCoulon Can you check the thales CI if the newest commit passes? It seems it's not visible to public despite I merged it in cli and pushed it to github

JeanRochCoulon (approved):
Hello, the dashboard seems broken for 3 days, but the regression is green !!! 
The bit manip extension support has been disabled, this is ok for me.
A new cva6 configuration package has been defined. This was not planned, as only the configurations specified in the CVA6 specifications were planned. I see no problem defining a new configuration, but since no regression is running on it, a future PR might cause it to fail. You (and your colleagues) are responsible for maintaining it.
Thank you for the PR, and your MMU bug fix.

tianrui-wei:
Thanks! Appreciate your review and guidance!

Issue #842, opened 2022-03-23
Segmentation fault using QuestaSim


davidmallasen:
Hello,

The latest version of the repo (8d893fb647d758bed5876307c64caa6adbc89f97) triggers a segmentation fault when trying to simulate using QuestaSim. This does not happen with commit 00236be3d8552f93a0bebda8f9820ec54b64a000 . The steps I have followed are:
~~~
git clone https://github.com/openhwgroup/cva6.git
cd cva6
git submodule update --init --recursive

mkdir tmp

echo '
#include <stdio.h>

int main(int argc, char const *argv[]) {
    printf("Hello CVA6!\\n");
    return 0;
}' > hello.c

riscv64-unknown-elf-gcc hello.c -o hello.elf

make sim elf-bin=$RISCV/riscv64-unknown-elf/bin/pk target-options=hello.elf batch-mode=1
~~~

This outputs the following:
~~~
Makefile:151: XCELIUM_HOME not set which is necessary for compiling DPIs when using XCELIUM
vlib work
[...]

Issue #800, opened 2022-01-22
Questions regarding simulation with Questa


KeithoHimself:
I am working to get the simulations described in the README running under Questa and I am running into issues.  Before digging deeper, I'd like to know the following:

1) Is the flow in the ./ci directory currently running under Questa?  Is this flow run nightly?

2) What version of Questa is being used?
    - I have found that modern version of Questa throw a sig 11 (i.e. they crash) when loading the simulation
    - Questa 2017.3 will load and run the sim, but when running "hello world", I do not see "Hello World!" in the transcript.
 
3) When using Questa, the sim tries to load libfesvr.so.  Where is that library being built?
   - The ./ci/install-fesvr recipe does not appear to build and install libfesvr.so, only libfesvr.a
   - A manual change to riscv-isa-sim/fesv/esvr.mk.in to add "fesvr_install_shared_lib = yes" appears to be required

rickoco:
Hey @KeithoHimself thanks for your comments.  Within the OpenHW Group compute platform on AWS, we're in the process of setting up and running various CI flows using Cadence and Synopsys tools and OpenHW community members also have Siemens / Mentor based flows.  It would be good to learn more about your environment.  Email me at rickoco@openhwgroup.org so we can setup a time to chat.
Rick

christian-lanius:
@KeithoHimself 
I am having the same issue, i.e. I am unable to simulate the design at this point in time using Questasim.
I have tried two versions of questasim: 
2018-19 Europractice version results in an error 
```
Error: {PATH}/cva6/corev_apu/clint/axi_lite_interface.sv(55): Internal error: unhandled IDL type ../../src/vlog/vsymtab.c(5443) <net_type>. Please contact Questa support at http://supportnet.mentor.com/
```
If it helps, I can provide the full list of these errors. 

Running the 2020-21 EP version results in the same sig 11 error you got. As the issue occurs in the opt step, I can circumvent it by skipping this step: Commenting out `$(VOPT) $(compile_flag) -work $(library)  $(top_level) -o $(top_level)_optimized +acc -check_synthesis` (line 344), replacing the name in the sim command in the Makefile with the unoptimized version and launching simulation with `QUESTASIM_FLAGS="-novopt -suppress 12110"` makes the simulation start. I also have to set the MODELSIM_HOME variable. However, the simulation then still crashes because of an infinite loop detected. 

So, even getting rid of the sig 11 error is not enough to get the simulation to run for me.

[...]

PR #988, opened 2022-11-02
Flist: Split Flists for SV32 and SV39 to prevent Questasim elaboration errors

niwis:
Fixes the questasim simulation flow. In particular:
~~1. Port width mismatch in `ariane_testharness`. Therefore, propagate the `AXI_USER_WIDTH` from `ariane_testharness` to `ariane_peripherals`~~ :arrow_right: #1043 
~~2. `riscv-tests` do not signal success or failure after execution in questasim. This is due to the `rvfi_tracer` intercepting `ecall`s and terminating the simulation prematurely (they are needed by the proxy kernel for communication). Therefore, disable `RVFI_TRACE` for regular simulations per default.~~ :arrow_right: #1141 
~~3. Some benchmarks time out in Questasim simulation. For instance, `rsort` requires over 4M cycles on the WB cache, which is well above the current default timeout of 2M cycles (not sure if this is due to a performance regression). Therefore, set the default timeout to 6M cycles which is a safe value for all riscv tests and benchmarks.~~ :wastebasket:
~~4. Fix the compilation order for Questasim (#1008).~~ :arrow_right: #1043 
~~5. Increase the testbench's memory size (#1014).~~ :wastebasket:
6. Questasim will error when elaborating the SV32 MMU for 64-bit CVA6. We had this issue a while again and filtered out the incompatible MMU version depending on the target. This was broken again with the change to file lists. Hence, split the MMU out of the common file list into a separate one that can be chosen from the Makefile. I believe using bender would be a more elegant and maintainable solution at this point. Furthermore, why the file name inversion (`Flist.<target>`)?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -198,11 +198,12 @@ src :=  core/include/$(target)_config_pkg.sv                                    
         corev_apu/tb/common/SimDTM.sv                                                \
         corev_apu/tb/common/SimJTAG.sv
 
+flists := ${CVA6_REPO_DIR}/core/Flist.cva6
 # SV32 MMU for CV32, SV39 MMU for CV64
 ifeq ($(findstring 32, $(target)),32)
-    src += $(wildcard core/mmu_sv32/*.sv)
+    flists += ${CVA6_REPO_DIR}/core/Flist.mmu-sv32
 else
-    src += $(wildcard core/mmu_sv39/*.sv)
+    flists += ${CVA6_REPO_DIR}/core/Flist.mmu-sv39
 endif
 
 src := $(addprefix $(root-dir), $(src))
@@ -283,7 +284,7 @@ endif
 vcs_build: $(dpi-library)/ariane_dpi.so
 	mkdir -p $(vcs-library)
 	cd $(vcs-library) &&\
-	vlogan $(if $(VERDI), -kdb,) -full64 -nc -sverilog +define+$(defines) -assert svaext -f ../core/Flist.cva6 &&\
+	vlogan $(if $(VERDI), -kdb,) -full64 -nc -sverilog +define+$(defines) -assert svaext $(addprefix -f , ${flists}) &&\
 	vlogan $(if $(VERDI), -kdb,) -full64 -nc -sverilog +define+$(defines) $(filter %.sv,$(ariane_pkg)) +incdir+core/include/+$(VCS_HOME)/etc/uvm-1.2/dpi &&\
 	vhdlan $(if $(VERDI), -kdb,) -full64 -nc $(filter %.vhd,$(uart_src)) &&\
 	vlogan $(if $(VERDI), -kdb,) -full64 -nc -sverilog -assert svaext +define+$(defines) $(filter %.sv,$(src)) +incdir+../vendor/pulp-platform/common_cells/include/+../vendor/pulp-platform/axi/include/+../corev_apu/register_interface/include/ &&\
@@ -301,7 +302,7 @@ build: $(library) $(library)/.build-srcs $(library)/.build-tb $(dpi-library)/ari
 
 # src files
 $(library)/.build-srcs: $(library)
-	$(VLOG) $(compile_flag) -timescale "1ns / 1ns" -work $(library) -pedanticerrors -f core/Flist.cva6 $(list_incdir) -suppress 2583
+	$(VLOG) $(compile_flag) -timescale "1ns / 1ns" -work $(library) -pedanticerrors $(addprefix -f , ${flists}) $(list_incdir) -suppress 2583
 	$(VLOG) $(compile_flag) -work $(library) $(filter %.sv,$(ariane_pkg)) $(list_incdir) -suppress 2583
 	# Suppress message that always_latch may not be checked thoroughly by QuestaSim.
 	$(VCOM) $(compile_flag_vhd) -work $(library) -pedanticerrors $(filter %.vhd,$(uart_src))
@@ -436,7 +437,7 @@ XRUN_COMP = $(XRUN_COMP_FLAGS)		\
 	$(filter %.sv, $(ariane_pkg)) 	\
 	$(filter %.vhd, $(uart_src))  	\
 	$(filter %.sv, $(src))	      	\
-	-f ../core/Flist.cva6    	    \
+	$(addprefix -f , ${flists})                 \
 	$(filter %.sv, $(XRUN_TB))	\
 
 XRUN_RUN = $(XRUN_RUN_FLAGS) 		\
@@ -540,7 +541,7 @@ xrun-ci: xrun-asm-tests xrun-amo-tests xrun-mul-tests xrun-fp-tests xrun-benchma
 
 # verilator-specific
 verilate_command := $(verilator)                                                                                 \
-                    -f core/Flist.cva6                                                                           \
+                    $(addprefix -f , ${flists})                                                                                  \
                     $(filter-out %.vhd, $(ariane_pkg))                                                           \
                     $(filter-out core/fpu_wrap.sv, $(filter-out %.vhd, $(src)))                                  \
                     +define+$(defines)$(if $(TRACE_FAST),+VM_TRACE)$(if $(TRACE_COMPACT),+VM_TRACE+VM_TRACE_FST) \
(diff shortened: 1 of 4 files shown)


christian-lanius:
Did you observe the issues discussed in #842 and #800 when you worked on this pull request?
So far, I am unable to simulate Ariane on our setup with Questasim, as I still get the segmentation fault during optimization. I tried to merge your changes into my local branch, but this did not help as far as I can tell.

christian-lanius:
Dear Nils,
That is disappointing to hear. I tried the latest master branch with your proposed changes and unfortunately I still get the same segmentation fault. I guess the seg fault is then unrelated to the missing ".AxiUserWidth". Even with questa 2020.4, which is the closest version that I have to the one you use.

Given that the repo works fine for you, there must be some mismatch between the simulation setups.

davidmallasen:
I'm also facing the same issue as @christian-lanius and getting a segfault following the steps in that issue from a clean repo. I also have questasim 2020.4.

niwis:
The segfault is not related to this PR. Let's continue this discussion in https://github.com/openhwgroup/cva6/issues/842

christian-lanius:
I don't have a simple way to submit pull requests and as this is related to this topic, maybe @niwis you can extend this pull request with the changes discussed in #1008 (the current master does not simulate in questasim, as the packages don't get built first and the path to the flist file is wrong) and in #1014 (increase the main memory parameter in the testbench back to 2**24, so that it is aligned with the verilator testbench). This fixes the pk example from the README, which is currently broken in questasim/vcs.

Unrelated, vcs is additionally broken in that the target-options don't propagate to the simulator, but that is a different topic.

niwis:
@christian-lanius done. @JeanRochCoulon could you have a look at this PR? The questasim regressions are piling up (OP up-to-date). I agree that it would probably make sense to add questasim to the CI and would be happy to help with that.

JeanRochCoulon:
Hello @niwis, FOr a while I have followed your current PR. As I will be out of office until Jan 2, I plan to review it begin of Jan.

Makefile
```diff
@@ -29,7 +29,7 @@ verilator      ?= verilator
 # traget option
 target-options ?=
 # additional definess
-defines        ?= WT_DCACHE+RVFI_TRACE
```
JeanRochCoulon, line 32: Unfortunately removing RFVI_TRACE will make all the Thales CI fail

Makefile
```diff
@@ -197,11 +197,12 @@ src :=  corev_apu/tb/axi_adapter.sv
         corev_apu/tb/common/SimDTM.sv                                                \
         corev_apu/tb/common/SimJTAG.sv
 
+flists := ${CVA6_REPO_DIR}/core/Flist.cva6
 # SV32 MMU for CV32, SV39 MMU for CV64
 ifeq ($(findstring 32, $(target)),32)
-    src += $(wildcard core/mmu_sv32/*.sv)
```
JeanRochCoulon, line 202: A recent PR has factorize the different flist into a single one. I do not understand why questa cannot manage it like VCS ?

corev_apu/tb/rvfi_tracer.sv
```diff
@@ -22,7 +22,7 @@ module rvfi_tracer #(
   int unsigned SIM_FINISH;
   initial begin
     f = $fopen($sformatf("trace_rvfi_hart_%h.dasm", HART_ID), "w");
-    if (!$value$plusargs("time_out=%d", SIM_FINISH)) SIM_FINISH = 2000000;
```
JeanRochCoulon, line 25: The SIM_FINISH must remain 2000000 to avoid to full fill the Thales server in case of problem. Bu you can extend the timeout by using the option --issrun_opts="+time_out=60000000" when launching your simulation with cva6.py

corev_apu/tb/ariane_tb.sv
```diff
@@ -34,7 +34,7 @@ module ariane_tb;
     // toggle with RTC period
     localparam int unsigned RTC_CLOCK_PERIOD = 30.517us;
 
-    localparam NUM_WORDS = 2**16;
+    localparam NUM_WORDS = 2**24;
```
JeanRochCoulon, line 37: Hello world is a small test, why do you need so much memory space ?

JeanRochCoulon (changes requested):
Hello @nils,    Thank you for the PR and I apologize for the delay.  As you can see the four commits (timeout, hello_word, flist, RVFI) must be discussed.
Can you submit a new PR without them to be able to merge.
See you

niwis:
Hello @JeanRochCoulon, I have opened a PR with 1. and 4. in #1043 while we discuss the remaining fixes here.

Makefile
niwis: Why? It looks to me like `RVFI_TRACE` is defined via the `isscomp_opts` variables when it's actually needed:
https://github.com/openhwgroup/core-v-verif/blob/521dbba95a408adb95965451dceec696e877f094/cva6/sim/cva6.py#L757-L758
I feel like the common users of this script do not want `RVFI_TRACE`, especially when executing software that contains `ecall` instructions such as the riscv-tests. I don't think that terminating on an `ecall` should be the default behaviour of the CVA6 simulation. It took us quite some git-bisecting to find out why the riscv-tests weren't passing anymore, and the same might happen to others.

JeanRochCoulon: You are right. The ultimate solution is to pass RVFI_TRACE through the isscomp_opts option. As it is not the case Today, all teh Thales CI will fail. If I gettime, I will try to fix it.

Makefile
niwis: I assume you mean #1012? That PR reintroduces the problem from #713 and is exactly what I am trying to solve here. The underlying problem is that Questasim elaborates all source files, even if the modules are not instantiated. When elaborating `mmu_sv32` for the CV64A6 configuration, it encounters a `Negative replication multiplier` error. The proposed solution is to prevent the elaboration of an incompatible MMU. More details in #713

corev_apu/tb/rvfi_tracer.sv
niwis: That works for me. I'll discard this change.

corev_apu/tb/ariane_tb.sv
niwis: I would have to investigate this further. But anyways, I also prefer the solution that was proposed last week by using a dynamic simulation memory. I will discard this change.

Makefile
christian-lanius: Is it too ugly to suppress the error? This will fix the issue, but might lead to actual errors being hidden. I am unaware about the details, but maybe this can be done on a per-file basis?

Makefile
JeanRochCoulon: Today we change our mind. We will modify the end of test to stop on HTIF condition (write a special data in to_host address) and no more on ecall. I let you know.

JeanRochCoulon:
Hello @nils,   Tests do not anymore terminate on ecall instruction but on a write access in to_host address when bit0=1. Detection is done by RVFI. To my mind, your PR can be simplified by removing the point number 2 of your list. Can you try it terminates on your side?  Thanks very much to have pointed this test termination limitation. Right now, the termination is aligned with what is implemented in Spike. Cheers

MarioOpenHWGroup:
Could you please rebase the branch ?

niwis:
Done. There should only be one remaining item (see top), which is splitting the Flist for sv32 and sv39. I have not tested this yet. But I should have bandwidth next week.

MarioOpenHWGroup:
> Done. There should only be one remaining item (see top), which is splitting the Flist for sv32 and sv39. I have not tested this yet. But I should have bandwidth next week.

I will look if the changes work on the newest commit as I have some fixes for the core-v-verif repo :+1: What Questasim version are you using?

niwis:
Great! I'm using 2020.1

JeanRochCoulon:
Hello, 
Unfornately the proposed modification (split Flist) will make the ASIC synthesis failed. The current flow takes the Flist to list the files to be read by the synthesis tool. Moreover, having a verilog not enough robust to support different VLEN/PLEN configurations is not good. That's why I woudl prefer the verilog fix. #1136  has implemented it for the no MMU mode, this might be applied on the sv32 and sv39 modes too.

Makefile
JeanRochCoulon: Hello, thanks @MarioOpenHWGroup to propose multi solutions. The suppress 8607 seems good to me and I would suggest to open an github issue to log the workaround.

niwis:
Thanks!

PR #993, opened 2022-11-09
Fifo v3 fpga

sjthales:
Adding of fifo_v3_fpga.sv which is an FPGA-optimized version of the fifo_v3.sv module.

In the original version of fifo_v3.sv, the stacked data is saved in D flip-flops.

This FPGA-optimized version aims to replace the use of D flip-flops with an asynchronous read memory available in Xilinx and Microchip FPGA.


JeanRochCoulon:
@sjthales @jquevremont 
In the last meetings, we decided to avoid ifdef commands and file duplication. I think the fifo_v3 module duplication could be avoided by conditioning the body description in relation to the FPGA_EN localparam configuration.

sjthales:
fifo_v3 is a pulp-platform module, that means we don't have control over this module.

What I propose is to rewrite fifo_v3 in common/local/util by adding FPGA_EN. --> The goal is to have control over fifo_v3

What do you think?

gurkaynak:
It is not good form to replace the original headers. This code is from ETH/Unibo author Florian Zaruba. It contains additions from sjthales (or someone else). Original headers have to remain. Feel free to add your information to the header, but you can not remove the original contribution if you keep the majority of the code the same.

JeanRochCoulon:
fifo_v3 could be replaced by fifo_v3_wrapper which uses on fifo_v3 module when FPGA_EN is zero and fifo_v3_fpga module when FPGA_EN is one. fifo_v3_wrapper.sv describing fifo_v3_wrapper module and fifo_v3_fpga.sv describing fifo_v3_fpga module are to be added to CVA6 repository. Tell me if it makes sense to you.

sjthales:
That makes sense but I will have to modify each instantiation of fifo_v3 to fifo_v3_wrapper. Are you ok?

sjthales:
In term of behavior, it is the same thing. The difference is that I instantiated an asynchronous read memory to use LUTRAM resources rather than D fip-fop.

sjthales:
I'm not sure I understand. What exactly do you suggest I do?
Make a PR to common_cells repository? How to manage the AsyncSpRam module? How to manage the fact that common_cells used in cva6 is not up to date.

zarubaf:
1. Add the change locally, not only in the fpga folder. You need to discuss with your colleagues as well on how to manage patches to common things. (See also related discussion on Mattermost)
2. Fix the asynch mem module so that it adheres to the style guide. Fix attributions in the fifo (see @gurkaynak comment).
3. Propose upstream.
4. Update common cells.

PR #995, opened 2022-11-11
[skip ci] Add hierarchy to the CVA6 documentation

MikeOpenHWGroup:
This PR _looks_ big, but it is not. Very few files have been changed.  However, it is a significant reorganisation that adds hierarchy to what is published on ReadTheDocs.  The document is now in three independent chapters:
1. **OpenHW Group CVA6 User Manual**: this is the original documentation authored by @zarubaf et al.
2. **CVA6 Requirements Specification**: this is the specification doc generated by @jquevremont et al.
3. **CV32A6 Design Document**: this is the design spec currently under development by @JeanRochCoulon and his team.

So the biggest change is merely organisational, with very little content change.  The most significant change is to replace "step 1" with "v0.1.0".

The "skip ci" prefix has been added to the title since this pull-request does not change any code.


JeanRochCoulon:
This re-organization is nice. Great to propose it. The old organization allowed to generate the doc under readthedocs. Is it again the case with the new orgnization ?  By the way, we should think about moving the doc under the openhwgroup account if possible.

jquevremont (approved):
LGTM

MikeOpenHWGroup:
Thanks for the reviews, I will merge this in.

To answer @JeanRochCoulon's questions:
> The old organization allowed to generate the doc under readthedocs. Is it again the case with the new orgnization ? By the way, we should think about moving the doc under the openhwgroup account if possible.

Yes, as soon as this is on the master branch I will generate a RTD project for it using the OpenHW Group account.  We should disable the existing non-OpenHW Group RTD project.  I will discuss this with you off-line.

PR #1004, opened 2022-12-02
Frontend btb fpga

sjthales:
- Optimized BTB for FPGA targets.

- Prediction target addresses are stored in a BRAM.

- For now flushing is not supported because the frontend module's flush signal in cva6 is not connected.


sjthales:
Any comment?

JeanRochCoulon:
I would be pleased to merge this PR, but file duplication must be avoided. FPGA_EN variable can condition the FPGA optimization inside btb.sv file. In that case the modification in frontend.sv would be smaller.

PR #1005, opened 2022-12-05
Use unique Flist for testharness and UVM

JeanRochCoulon:
Modification has been verified for VCS (Synopsys) simulation, Verilator simulation and FPGA build.

Modifications have been also done for XRUN and Questa, but without verification because I do not own this type of licence. @MikeOpenHWGroup do you know someone able to try to simulate with such tools ?


JeanRochCoulon:
@MikeOpenHWGroup the GitAction and Thales CI are clean with VCS and Verilator tools. Do not hesitate to test it with other tools. Regards

PR #1006, opened 2022-12-05
[DO NOT MERGE, superseded by #1007] Vendorize core submodules

zchamski:
This is currently a preview of vendorization, applied to `common/submodules/common_cells`.

The process consists in four steps (the first step being done just once for all repos):
* Add vendorization script.
* Add configuration script for PULP `common_cells` submodule.
* Import PULP `common_cells` module.
* Update core and APU configuration files to use new import.


zarubaf:
That looks very good! One small hint from my side would be that you can exclude files from the vendorization process. I usually did not take any files other then `LICENSE` and the source files (so really just what is needed), to avoid cluttering the repo. Maybe that is something you want to do as well?

zarubaf:
Nit: the flexfloat code is not used in CVA6, so maybe we can leave it out from vendorizing?

zchamski:
> That looks very good! One small hint from my side would be that you can exclude files from the vendorization process. I usually did not take any files other then `LICENSE` and the source files (so really just what is needed), to avoid cluttering the repo. Maybe that is something you want to do as well?

Hi @zarubaf , I actually excluded quite some files already (docs, CI configuration files etc.).  The exclude lists can surely be refined further, maybe based on the actual content of Flists from #1005.

JeanRochCoulon:
The https://github.com/openhwgroup/cva6/pull/1006 PR has been merged. Please rebase your PR to allow the merge into master.

PR #1023, opened 2022-12-23
Remove unused submodules

niwis:
Some submodules were vendored in #1015. Remove those submodules, as they are no longer used.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,9 +4,6 @@
 [submodule "common/submodules/common_cells"]
 	path = common/submodules/common_cells
 	url = https://github.com/pulp-platform/common_cells.git
-[submodule "corev_apu/axi"]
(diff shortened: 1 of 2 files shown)


zarubaf (approved):
Perfect, thanks!

zchamski:
Hi @niwis , vendorization was applied to all modules referenced from `core`, but only the references in `core` have been switched over to use the vendorized paths.  Because of this, the removal of the AXI submodule in PR #1023 seems to have broken the FPGA synthesis jobs in CI.  Specifically, the FPGA code in `corev_apu` refers to the original submodule location of AXI submodule and synthesis breaks on missing AXI definitions.

niwis:
Thanks for noticing! Should be fixed now.

PR #1027, opened 2023-01-03
rvfi: fix mem_wdata

VarunKoyyalagunta:
Was looking at the wrong WB port, `load_result` instead of `store_result`.
Broken from 4c01614f83cd9dd91d115aa190bd4fc6a284b59e


JeanRochCoulon:
Hello @VarunKoyyalagunta   Thanks for your PR. But looking at the current scoreboard.sv, the error seems fixed. Maybe someone else saw the error in parallel. 
Can I ask you to approve that the bug was fixed and close your PR if it is the case ?

Issue #683, opened 2021-06-17
Performance Counters width


nathpranay:
Hi Team,

As per riscv-privileged spec "All of these counters have 64-bit precision on RV32, RV64, and RV128." In CVA6 performance counter width is tied to XLEN value. According to the spec, the perf counters width should be 64 bits.

-Pranay

MikeOpenHWGroup:
Hi @ASintzoff, can you respond to @nathpranay's question?  Thanks!

ASintzoff:
In PR #1034 modifications are proposed to be compliant with the 64-bit width.

ASintzoff:
Performance counter section in the User Manual refers to 64-bit width.

PR #1034, opened 2023-01-20
Implementation of 64bit perf_counters

TulikaSi:
This PR is meant to make the performance counters XLEN independent and make it as generic 64bit counters (as specified in the RISC-V specifications)

### DIFF
--- a/core/cache_subsystem/cache_ctrl.sv
+++ b/core/cache_subsystem/cache_ctrl.sv
@@ -115,6 +115,7 @@ module cache_ctrl import ariane_pkg::*; import std_cache_pkg::*; #(
         req_port_o.data_gnt    = 1'b0;
         req_port_o.data_rvalid = 1'b0;
         req_port_o.data_rdata  = '0;
+        req_port_o.data_rid    = '0;
         miss_req_o    = '0;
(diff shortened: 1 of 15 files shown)


sorear:
Which spec requirement is this going off of?  I see a 64-bit requirement in the ratified unprivileged spec 20191213 §10.1 for cycle/time/instret only but §10.2 which describes hpmcounterN has `The number and width of these additional counters, and the set of events they count is platform-specific.`

TulikaSi:
Hi @sorear 
The section 10.2 of Volume 1, Unprivileged Spec v. 20191213 mentions about the **29 additional unprivileged [hpmcounter3–hpmcounter31]** which are **64 bit**.
The statement which you pointed out from the specification i.e. ( The number and width of
these additional counters, and the set of events they count is platform-specific.) refers to the additional 29 **event selector CSRs, mhpmevent3–mhpmevent31**, that controls the selection of the event.
Section 3.1.10 of  Volume 2, Privileged Spec v. 20211203 describes the counters in more detail.

https://github.com/openhwgroup/cva6/issues/683#issue-923509007---This issue speaks about the performance counter

PR #1040, opened 2023-01-31
Add number of pipe registers in configuration

numero-744:


core/include/cv32a60x_config_pkg.sv
```diff
@@ -37,4 +37,7 @@ package cva6_config_pkg;
 
     localparam CVA6ConfigFPGAEn = 0;
 
+    localparam CVA6ConfigNrLoadPipeRegs = 0;
```
zarubaf, line 40: Can we keep it the same as the old? Otherwise this might break synthesis for people.
```suggestion
    localparam CVA6ConfigNrLoadPipeRegs = 1;
```

numero-744: Yes I agree with you, I initially wanted to keep the same current values to keep compatibility with current version.

I have removed the last commit.

zarubaf (approved):
👍

PR #1042, opened 2023-02-01
Makefile: Propagate CFLAGS from env

niwis:
Propagate the CFLAGS and CXXFLAGS specified in the environment executing the Makefile for DPI compilation. By default, the current CFLAGS are extended by the CFLAGS from the environment.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -120,7 +120,7 @@ endif
 
 dpi_hdr := $(wildcard corev_apu/tb/dpi/*.h)
 dpi_hdr := $(addprefix $(root-dir), $(dpi_hdr))
-CFLAGS := -I$(QUESTASIM_HOME)/include         \
+CFLAGS += -I$(QUESTASIM_HOME)/include         \
           -I$(VCS_HOME)/include \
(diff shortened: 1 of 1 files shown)


zarubaf:
Shouldn’t it be enough to do the first `CFLAGS` assignment with `+=` instead of the `:=`?

niwis:
That would absolutely work for me as well! The idea here was to make it a bit more generic, e.g. to allow to unset flags by defining `DPI_CFLAGS`, and not mixing `CLAGS` with `CXX` (afaik it is convention to pass `CFLAGS` to `CC` and `CXXFLAGS` to `CXX`).
But I would also be happy to just change `:=` to `+=` if this is preferred.

JeanRochCoulon:
As th CI is OK, tell me if merging right now or if you prefer to implement += instead of :=

zarubaf:
I think I would prefer the `+=` change. Less intermediate variables need to be tracked throughout the code. If you don't mind.

niwis:
Sure! Done.

PR #1064, opened 2023-02-14
[WIP - do not merge YET] rvfi_tracer.sv: Terminate tests on write-into-tohost iso. ECALL instruction.

zchamski:
This PR changes the simulation termination condition from "execute an ECALL instruction" to "write `0b........1` at `tohost` memory address". 
* The value written into `tohost` must have its least-significant bit set.
* The bits above the LSbit contain the return value of the test.
* The execution of an `ecall` instruction does not terminate the simulation anymore. Instead, the execution of the `ecall` instruction will be handled as implemented, including the side effects in the environment.


zchamski:
Closing: superseded by https://github.com/openhwgroup/cva6/pull/1075

PR #1066, opened 2023-02-15
fix: improve stability of config_pkg_generator

cathales:
It was printing "setting xxx to yyy" but not modifying the line in the file for some values. Now it works properly.

I reworked the code until I found the issue, so it is possible to reduce the "diff to fix" if needed. Also I do not really understand why it was working for some lines but not all.


JeanRochCoulon:
@Gchauvon can you approve the modification ?

Gchauvon:
LGTM

PR #1077, opened 2023-02-25
[skip ci] Interfaces in a different file

jquevremont:
According to https://docs.google.com/document/d/1W3gBZPk0tNWquXo5CkcLWU07CdxaaT0gzsnhQ_dpiIo, Interfaces is a different file.

### DIFF
--- a/docs/01_cva6_user/Interfaces.rst
+++ b/docs/01_cva6_user/Interfaces.rst
@@ -1,0 +1,49 @@
+..
+   Copyright (c) 2023 OpenHW Group
+   Copyright (c) 2023 Thales
+
+   SPDX-License-Identifier: Apache-2.0 WITH SHL-2.1
+
+.. Level 1
+   =======
+
+   Level 2
+   -------
+
+   Level 3
+   ~~~~~~~
+
+   Level 4
+   ^^^^^^^
+
+.. _cva6_interfaces:
+
+Interfaces
+==========
+
+AXI Interface
+-------------
+Need for step1 verification. Already written by MU Electronics.
+Focus on the features used by the CVA6 and refer to ARM documentation for the AXI specification (e.g. do not draw the standard chronogram).
+Features:
+* See requirement specification
+* Atomic transactions
+* “USER” bus width extension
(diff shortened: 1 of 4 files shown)


docs/01_cva6_user/Interfaces.rst
```diff
@@ -0,0 +1,49 @@
...
+
+   Level 4
+   ^^^^^^^
+
+.. _cva6_interfaces:
+
+Interfaces
+==========
+
+AXI Interface
```
JeanRochCoulon, line 24: AXI interface chapter is big, can you create a dedicated file for it ?  When done, I will push the file

docs/01_cva6_user/Interfaces.rst
```diff
@@ -0,0 +1,49 @@
...
+
+   Level 3
+   ~~~~~~~
+
+   Level 4
+   ^^^^^^^
+
+.. _cva6_interfaces:
+
+Interfaces
```
JeanRochCoulon, line 21: Why not insert CVXIF in _Interfaces_ chapter ?   CVXIF chapter is big, would be helpful to dedicated a special file.

JeanRochCoulon (changes requested):

docs/01_cva6_user/Interfaces.rst
jquevremont: In this AXI section, I plan to point on the specificities of the interface (USER, atomic support), so I think it will be quite compact.
The list of signals will be on the prior section.
For the AXI specification (chronograms, transactions...), we'll point to ARM's references.
If a chapter gets too big, we can always decide to cut it later.

docs/01_cva6_user/Interfaces.rst
jquevremont: There's more in the CV-X-IF chapter than for other interfaces, like "how to design a coprocessor", "how to program it". I realize that we lost some CV-X-IF sections when moving from GDocs to GitHub.

JeanRochCoulon (approved):

PR #1078, opened 2023-02-27
Add mem_paddr information into RVFI interface

JeanRochCoulon:


JeanRochCoulon:
Hello @zchamski   This pull request can be useful to implement "end of test". Can you try it to confirm it is ok for you. In this case I would merge it.

zchamski:
Hi @JeanRochCoulon , my local CI of test termination went OK for both physical and virtual modes (tested with smoke-tests and dv-riscv-tests).

PR #1099, opened 2023-03-07
Specify CVA6 parameters and relative CV32a6_v5.0.0 configuration

JeanRochCoulon:


jquevremont:
For the sake of the reader, may I suggest the following:
- In the first table, add a column to explain what the values mean (e.g. when CExtEn=1, the RVC compact instructions are supported; when 0 these instructions are not supported), so that the user do not have to read the code to understand how to configure CVA6.
- In the first table, add a column to say where the parameter is set (`ifdef, ariane_pkg, top parameters...)
- In the second table, explain what the value implies for this configuration (e.g. "RVC support" on CExtEn row).

If you do not have time to fill these columns now, at least they could be added and left temporarily blank, so that the user knows the information will come some day.

JeanRochCoulon:
Thank you Jérôme for the review. I think that adding information will ease the user reading. But as the list of parameters and how they are declared will likely change in the coming weeks, following the ARA/Hypervisor/Cache merges, I would prefer if you merge as it is Today, and revisit it when the strategy versus configuration will be more stable.

PR #1101, opened 2023-03-07
Update `Bender.yml`

niwis:
This is required (in some form) by #1024.

Most PULP systems (like Ara) use bender for dependency and source file management.

Problem: the current `Bender.yml` directly includes the vendored IPs (e.g. common cells). However, if common cells is also listed as a dependency of the system, this causes redefinitions of the same module.

Solution: do not include the vendored dependencies in `Bender.yml`, but add them as a bender dependency. We track the vendored revision in `Bender.lock` to stay in sync.

I know it's not a perfect solution, but it should be non-breaking since afaik `Bender.yml` is not used anywhere within this repo.

Adding @joennlae who authored the changes and validated them in [Cheshire](https://github.com/pulp-platform/cheshire).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,7 +16,6 @@ src/moore.sv
 build/*
 /tmp*
 *.dasm
-/Bender.lock
 /Bender.local
 build/
 *.vcd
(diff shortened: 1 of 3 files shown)


zarubaf (approved):
Cool, Bender isn't used by anyone so that should be okay! Thanks

PR #1103, opened 2023-03-07
ariane.sv: Propagate AXI parameters

niwis:
Required by #1024.

(Optionally) expose the AXI parameters of `cva6` from the `ariane` wrapper.

### DIFF
--- a/core/ariane.sv
+++ b/core/ariane.sv
@@ -66,6 +66,7 @@ module ariane import ariane_pkg::*; #(
     .AxiIdWidth ( AxiIdWidth ),
     .axi_ar_chan_t (axi_ar_chan_t),
     .axi_aw_chan_t (axi_aw_chan_t),
+    .axi_w_chan_t (axi_w_chan_t),
     .axi_req_t (axi_req_t),
     .axi_rsp_t (axi_rsp_t)
   ) i_cva6 (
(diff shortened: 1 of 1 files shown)


core/ariane.sv
```diff
@@ -14,7 +14,15 @@
...
-  parameter ariane_pkg::ariane_cfg_t ArianeCfg     = ariane_pkg::ArianeDefaultConfig
+  parameter ariane_pkg::ariane_cfg_t ArianeCfg     = ariane_pkg::ArianeDefaultConfig,
+  parameter int unsigned AxiAddrWidth = ariane_axi::AddrWidth,
+  parameter int unsigned AxiDataWidth = ariane_axi::DataWidth,
+  parameter int unsigned AxiIdWidth   = ariane_axi::IdWidth,
+  parameter type axi_ar_chan_t = ariane_axi::ar_chan_t,
+  parameter type axi_aw_chan_t = ariane_axi::aw_chan_t,
+  parameter type axi_w_chan_t  = ariane_axi::w_chan_t,
+  parameter type axi_req_t = ariane_axi::req_t,
+  parameter type axi_rsp_t = ariane_axi::resp_t
```
zarubaf, line 25: You also probably want to use the parameter in the port-list, right?

niwis: Good catch, thanks!

zarubaf: Did you add it? I can't see the change in the ports, I think just added `axi_w_chan_t`, but the ports still reference the axi struct from the package. I would also expect modifications in `cva6`. As well as LSU etc. I guess this change only makes sense if it is propagated, right?

zarubaf: Of course, you are right, thanks!

zarubaf (approved):
LGTM

PR #1104, opened 2023-03-07
ariane_pkg: Parametrise cache layout

niwis:
Required by #1024.

With this change, different CVA6 configurations can define different cache layouts in their respective `cva6_config_pkg`.


zarubaf (approved):
Terrific! Thanks

JeanRochCoulon (approved):
Great @niwis !!

PR #1105, opened 2023-03-07
ariane_pkg: Set AXI_USER_WIDTH to 1 if disabled

niwis:
Required by #1024 

In the current implementation, `AXI_USER_WIDTH` equals 2 (`FETCH_USER_WIDTH*2`) if `AXI_USER_EN` is false. This causes unexpected port width mismatches when the user signal is disabled.


zarubaf (approved):
Classic system verilog! Thanks

JeanRochCoulon:
Test failed in vcs-uvm... (again)

PR #1107, opened 2023-03-08
Fix definition of the cache size (data and instructions)

cfuguet:
The cache size is a given number of bytes.

Currently it was computed as a given number of lines but then the width of indexes was bad.

With this fix:
- CVA6ConfigIcacheLines and CVA6ConfigDcacheLines define the total number of lines in the cache.
- The cache size is computed as CVA6Config[ID]cacheLines * number of bytes per line
- The configuration packages were modified to keep the original capacities: 16KB for Icache and 32KB for Dcache.

There is an exception for the cv32a6_ima_sv32_fpga_config_pkg where the capacities are 8KB for both Icache and Dcache.

### DIFF
--- a/core/include/ariane_pkg.sv
+++ b/core/include/ariane_pkg.sv
@@ -287,11 +287,10 @@ package ariane_pkg;
 
     localparam FETCH_USER_WIDTH = (cva6_config_pkg::CVA6ConfigFetchUserEn == 0) ? 1: cva6_config_pkg::CVA6ConfigFetchUserWidth;  // Possible cases: between 1 and 64
     localparam DATA_USER_WIDTH = (cva6_config_pkg::CVA6ConfigDataUserEn == 0) ? 1: cva6_config_pkg::CVA6ConfigDataUserWidth;    // Possible cases: between 1 and 64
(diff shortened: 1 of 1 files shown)


JeanRochCoulon:
@niwis, can you approve this PR ?  Thanks

niwis (commented):
Previously, `CVA6ConfigXcacheLines` was one way's capacity (in bytes). I admit that this was a bad name. But maybe instead of converting total size to number of lines and back to the size, would it be more readable to define a parameter `CVA6ConfigXcacheByteSize` and use this directly?

cfuguet:
Yes, I agree. I did not want to change the semantics you introduced recently with the number of lines. But I also find more convenient to define the parameters of the cache by:
- Total capacity (size) in bytes => CVA6ConfigXcacheByteSize 
- Its number of ways => CVA6ConfigXcacheSetAssoc
- Size of cache-cache lines in bytes => CVA6ConfigIcacheLineWidth/8

Do you want me to do the modification ?

cfuguet:
Yes. I will do the modification

cfuguet:
Replaced by pull request #1110

PR #1108, opened 2023-03-08
added shared TLB

sjthales:
Level 2 shared TLB N-way associative , it allows to reduce the number of ITLB and DTLB entries without degrading processor performance.


JeanRochCoulon:
Hello @sjthales 
This PR does not offer the opportunity to choose between the old or new TLB architecture. Merging it will break the ASIC design flow because the RAM instantiated in new TLB is not yet handled. To be discussed.

zarubaf (approved):
I think this is a very good change and makes sense!

Maybe I would wish for a bit more documentation for example on the user guide to mention that there should be SRAMs generated for it.

I leave it to you @JeanRochCoulon to decide for that.

PR #1113, opened 2023-03-12
pd/synth/Makefile: Enable RVFI_MEM support at gate level.

zchamski:
This PR enables the synthesis of RVFI_MEM support which was missing in gate synthesis configuration.  As a consequence, the RVFI tracer module could not see memory-related events when attached to gate-level models.


JeanRochCoulon:
Good catch @zchamski !!!!

PR #1119, opened 2023-03-16
sram: Tie unused user signal to `0`

niwis:
this avoids undriven user signals in the AXI interconnect if disabled.


zarubaf (approved):

JeanRochCoulon:
Hello @niwis 
The DROMAJO has been removed from the RTL by #1204. This removes RTL lines and fix the issue you have identified. If you think the issue is solved, you can close the PR. Thanks !

niwis:
Hi @JeanRochCoulon, indeed, the first block has been resolved by removing DROMAJO, and the second block was fixed identically in #1159. So this is not needed anymore

PR #1121, opened 2023-03-16
Add id signal to the request/response interface towards the data cache

cfuguet:


sjthales:
hello, when you modify MMU sv39, you need also to modify MMU Sv32 to be compliant in 32bit version.

cfuguet:
Thank you @sjthales, you are right. I forgot about the 32 bits MMU. Now the work is done for it too.

Do you plan to have a test in the CI for the CV32A6 version ? It will be useful to be sure that new contributions do not forget about it as I did.

Thank you,

César

JeanRochCoulon:
Hello @cfuguet    I understand you add an ID to the data access, and Today it is tied to zero. Does it mean that in the futur you plan to implement a ldst which will be able to launch an memory access while the previous one is not completed ?  Cheers

cfuguet:
Hello @JeanRochCoulon, yes that is the idea. A PR will follow to support multiple inflight loads. This is in preparation of the integration of the HPDcache 👍

JeanRochCoulon:
interesting !!!

PR #1124, opened 2023-03-17
Add parameter to define the width of the memory transaction ID

cfuguet:
This new parameter in the CVA6 configuration packages defines the width of the transaction ID between the I/Dcaches and the interconnection interface.

It should be defined in global CVA6 packages and not in any specific implementation of a cache to ease the integration of different cache subsystems.

### DIFF
--- a/core/cache_subsystem/cva6_icache.sv
+++ b/core/cache_subsystem/cva6_icache.sv
@@ -26,7 +26,7 @@
 
 
 module cva6_icache import ariane_pkg::*; import wt_cache_pkg::*; #(
-  parameter logic [CACHE_ID_WIDTH-1:0]  RdTxId             = 0,                                  // ID to be used for read transactions
+  parameter logic [MEM_TID_WIDTH-1:0]   RdTxId             = 0,                                  // ID to be used for read transactions
   parameter ariane_pkg::ariane_cfg_t    ArianeCfg          = ariane_pkg::ArianeDefaultConfig     // contains cacheable regions
 ) (
   input  logic                      clk_i,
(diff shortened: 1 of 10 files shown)


JeanRochCoulon:
Thanks for the PR. Unfortunately the smoke tests fail in CI !   
https://riscv-ci.pages.thales-invia.fr/dashboard/dashboard_cva6_0.html

cfuguet:
Hi @JeanRochCoulon ! Is it possible for me to see why it failed ? How can I locally run those tests ?

Thanks

JeanRochCoulon:
You can locally execute the cva6/regress/smoke_tests.sh file. Before do not forget to configure your terminal env by following README instructions.

cfuguet:
Ok thank you. It will be the occasion to play for the first time with the core-v-verif repository 👌

cfuguet:
Actually this ID is between caches and the NoC. It is not the one between the core and the caches.

It was already 2 bits by default. I did not change that. The issue was more that this parameter was a constant in the WT cache package. With this modification it is now defined in ariane_pkg.

cfuguet:
I just fixed some conflicts that appeared after some recent merged pull requests.

To complete my response before, the current WT cache needed multiple ID bits because it supports multiple outstanding writes (when the number of entries in the write buffer is bigger than 1). This is why the default value was 2. I kept that.

The commit in this PR makes the number of bits for the ID towards the NoC configurable. Its value is defined in the target cv* configuration package.

cfuguet:
Hello @JeanRochCoulon, I am not sure to understand why the checks are failing now. It looks like an authentication problem with Opensuse servers?

In the Thales dashboard checks are passing.

Best regards

JeanRochCoulon:
Hello @cfuguet 
I see you are fighting since 2 days to have this GitHub Actions CI pass !!!
I do not understand the root cause, but I advise you to submit a new PR.
Regards

Issue #1122, opened 2023-03-17
[BUG] std_cache_subsystem (write-back cache) cannot be currently used


cfuguet:
### Is there an existing CVA6 bug for this?

- [X] I have searched the existing bug issues

### Bug Description

Hello,

I've noticed that currently the write-back cache cannot be used with the current compilation makefiles and list of files for the CVA6.

There are two issues:
1. In the Flist.cva6 there is a '+define+WT_DCACHE' that forces the use of the write-through cache. This define shall be removed.
2. std_cache_subsystem files are not listed.

In the cva6.sv file, there is a '#ifdef WT_DCACHE .... #else '. The only way to use the write-back cache is to not define the WT_DCACHE macro. However, as mentioned above, it is not currently possible.

By the way, this issue means that the CI, that currently supposes to execute tests on both WT and WB caches, do not actually do it.

JeanRochCoulon:
Define are so dangerous !!!   We must remove them !

JeanRochCoulon:
@cfuguet  #1127 should fix the bug. Can you double check ?

cfuguet:
Yes, that is an option. The drawback is that we will have code replication in every config file. But not that significant I guess.

cfuguet:
This was already fixed

Issue #603, opened 2021-02-01
Describe CVA6 for write-through and write-back cache versions


MikeOpenHWGroup:
In #602, there was a conversation like this:

> Is the list of files in the manifest complete?
>> Looks okay for the write-through cache version. Of course, the write-back cache needs different files.

AFAICT, the ability to compile CVA6 for either  write-through and write-back cache versions is not described in the User Manual.  The User Manual should be updated to explain:
1. How to compile the CVA6 for  write-through and write-back cache versions.
2. Theory of operation for the write-through cache controller.
3. Theory of operation for the write-back cache controller.

j-silv:
> 1. How to compile the CVA6 for  write-through and write-back cache versions.

For an FPGA emulation (say for example, using the [Genesys 2 board](https://reference.digilentinc.com/reference/programmable-logic/genesys-2/reference-manual)) can we enable the write-back cache by simply commenting out the `` `define WT_DCACHE`` line in [the genesysii.svh file?](https://github.com/openhwgroup/cva6/blob/3ddf797e95923fd11113c8e443046105dfbf8843/corev_apu/fpga/src/genesysii.svh)

It seems as though the `WB_DCACHE` define is not mentioned anywhere else in the source code, so I'm assuming we don't need to uncomment that line.

suppamax:
@j-silv yes, by default the write-back cache is selected; you have compile the design with `WT_DCACHE` defined if you want to use a write-through cache

suppamax:
We've just [published](https://planv.tech/2022/11/08/cva6s-instruction-and-writeback-data-cache/) our notes about the WB cache.
And here's an interesting [paper](https://arxiv.org/pdf/2202.03749.pdf) about the WT cache.

Hope it helps

MikeOpenHWGroup:
Hi @JeanRochCoulon, this is an old issue with new information.  Let's get this into the User Manual.

JeanRochCoulon:
I agree. The cache type should be addded in the CVA6 parameter chapter of user manual. I take the point !

JeanRochCoulon:
Hi, I have just merged the #1127 to replace the define by a localparam into the config pkg. On my side, the issue is solved. Mike, up to you to take the decidsion to close the issue.

MikeOpenHWGroup:
Thanks everyone!  Closing.

Issue #483, opened 2020-07-24
Avoid using System Verilog defines in RTL code


Silabs-ArjanB:
The CVA6 uses the following defines (used in ifdef/ifndef statements):

- DROMAJO
- FIRESIM_TRACE
- FORMAL
- INSTR_TRACER_IF_SV
- PITON_ARIANE
- SYNTHESIS
- VERILATOR
- WT_DCACHE

Similar to the request made in https://github.com/openhwgroup/cv32e40p/issues/301 all such System Verilog defines should (ideally) be completely avoided in RTL code (and for example be replaced by localparam or parameter, avoided by a cleaner split between RTL and non-RTL code (in separate files), etc.). On CV32E40P all defines were avoided by rewrites performed in https://github.com/openhwgroup/cv32e40p/pull/389 (actually the define related to assertions is still there, but will soon be removed as well as assertions will move into seperate files outside of the RTL files). If for some reason defines will still be used, it would be cleaner if they use a uniform prefix (e.g. CVA6_)  to reduce the chance of conflicts with other RTL code.

Jbalkind:
PITON_ARIANE is already prefixed with PITON_ as it is related to OpenPiton. That is our prefix. Please do not remove these macros.

MikeOpenHWGroup:
Thank you for your comments @Silabs-ArjanB, and FWIW, I completely agree with you.

@JeanRochCoulon, let's solve this one-and-for-all.  Let's first tackle the non `PITON_ARIANE` macros.  Once we have a solution in place, we can raise the topic with @Jbalkind.

JeanRochCoulon:
@MikeOpenHWGroup @Jbalkind 
I have started to remove the defines from the RTL. #1127 is a PR to remove WT_DCACHE.

After the modification, you can find below the remaining defines in the CVA6 RTL:
grep -r ifdef |grep .sv|grep core/
core/include/wt_cache_pkg.sv:`ifdef PITON_ARIANE
core/include/wt_cache_pkg.sv:`ifdef PITON_ARIANE
core/include/ariane_pkg.sv:`ifdef PITON_ARIANE
core/include/ariane_pkg.sv:`ifdef PITON_ARIANE
core/include/ariane_pkg.sv:`ifdef SPIKE_TANDEM
core/include/ariane_pkg.sv:`ifdef PITON_ARIANE
core/include/ariane_pkg.sv:`ifdef RVFI_MEM
[...]

PR #1127, opened 2023-03-21
Replace WT_DCACHE define by CVA6ConfigCacheType localparam

JeanRochCoulon:
Cache WT or WB was configured by a define.
#603 requests more description of it
#483 tells to not use define
#1122 reports a bug: WB is never used/verified

To solve these issues, WT_DCACHE define has been replaced by CVA6ConfigCacheType localparam

In GitHub Actions CI, cv32a6_imafc_sv32 has been added, this configuration implements WB.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       matrix:
         testcase: [asm-tests, mul, amo, fp, benchmarks]
-        cache: [WB_DCACHE, WT_DCACHE]
+        target: [cv64a6_imafdc_sv39, cv32a60x]
     env:
       RISCV: /riscv
     steps:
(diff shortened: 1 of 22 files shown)


cfuguet:
Hi @JeanRochCoulon, I made an eye-review (no simulation), and it looks good.

However, I have one remark: currently you suppose a binary choice, either the WB or the WT cache. As you know, there is another cache subsystem that is showing the tip of the nose :).

Should not we have some kind of enumeration type that defines legal cache subsystem types:
enum {
WB = 0,
WT = 1,
// HPDcache = 2
} cache_type_t ;

This values can be used in the configuration package to select a given cache subsystem. We can then add new types to it if new cache subsystems are supported.

The question is now where to define this enumeration type. It cannot be in ariane_pkg because we will have a circular dependency between config packages and ariane_pkg. We could create an additional file: ariane_cache_pkg.sv. Constants from it can be imported from the config packages.

Just an idea...

PR #1128, opened 2023-03-21
Config WT dcache wbuff depth 

sjthales:
Configure wbuffer size for FPGA target.


JeanRochCoulon:
Thales CI is failed. Can you rebase your PR? I think it will solve the problem.

PR #1134, opened 2023-03-24
Add parameter on config file to define the width of the memory transaction ID

cfuguet:
This new parameter in the CVA6 configuration packages defines the width of the transaction ID between the I/Dcaches and the interconnection interface.

It should be defined in global CVA6 packages and not in any specific implementation of a cache to ease the integration of different cache subsystems.

Current default value (before this PR) is 2 bits. I kept this value, then no increase in the area is expected.

@JeanRochCoulon, can you review these changes?

Thank you

### DIFF
--- a/core/cache_subsystem/cva6_icache.sv
+++ b/core/cache_subsystem/cva6_icache.sv
@@ -26,7 +26,7 @@
 
 
 module cva6_icache import ariane_pkg::*; import wt_cache_pkg::*; #(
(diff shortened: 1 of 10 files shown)


JeanRochCoulon:
Aaah !!  CI works better :-)

cfuguet:
Thanks @JeanRochCoulon, I'm more close to being able to integrate the HPDcache :)

PR #1137, opened 2023-03-24
Fix enum value comparison: it needs explicit cast to int

cfuguet:
QuestSim requires to explicitly cast to integer an enumeration variable when comparing against an integer constant.


cfuguet:
Hello @JeanRochCoulon,

Latest modifications to select among different cache subsystems with an enumeration value, do not compile with QuestaSim.

This PR shall fix the issue.

Regards,

JeanRochCoulon:
Thanks @cfuguet for your watching

PR #1141, opened 2023-03-27
remove RVFI_TRACE/RVFI_MEM ifdef verilog directive

JeanRochCoulon:
Remove the defines to solve #483

Optional input/output are managed by parameter type definition


JeanRochCoulon:
Of course the CVA6 instantiation will be impacted
@Jbalkind and @sjthales can you approve ?
Thanks

Jbalkind:
I'm not clear on how to use the new parameter in a way that's equivalent to not defining RVFI_MEM. Can you explain? It looks like at present there are going to be a bunch of extra wires added to the design unconditionally.

JeanRochCoulon:
@Jbalkind I changed strategy. I replaced the parameter trick by wrapper trick, this is surely less impacting for project instantiating cva6.

This PR has been replaced by #1153

JeanRochCoulon:
This PR has been updated to 
- remove RVFI_TRACE and RVFI_MEM inside cva6 core, this should not impact the cva6 instantation.
- remove the FIRESIM IO interface. As this interface was with a bug, it could not be activated. That's why the modification should not impact.

I plan to merge on Tomorrow end of (my) day. @Jbalkind, @sjthales send me an alert if you see something wrong.

Jbalkind:
I don't think I see any issues here

JeanRochCoulon:
@Jbalkind The ultimate commit adds the definition of a define directive (RVFI_PORT) into the cva6_config_pkg. This define is not used by cva6 but by the top level module which instantiates cva6 (ariane.sv and ariane_testharness). By this way, no other define directives are needed at top level. Without any opposition, the plan is to merge on next Tuesday.

Jbalkind:
Interesting, ok. No problems from me. Perhaps we can adopt something similar for PITON_ARIANE.

PR #1153, opened 2023-03-31
Remove idef directives from IO cva6 module

JeanRochCoulon:


JeanRochCoulon:
Also if the impact of interface modification has been minimized, @Jbalkind, can you approve this PR ?

Jbalkind:
I'm not quite sure I follow here. It's ok to still have the macros in ariane.sv because it's in corev_apu now?

Are we also looking to remove the macro from elsewhere in the hierarchy? I see that wt_cache_subsystem.sv still has the PITON_ARIANE macro on its IO ports.

JeanRochCoulon:
The objective is to remove all the directives from the CVA6 core, APU is another discussion which I do not open. As in CVA6 core itself it is a huge modification, I split the modification in several PRs. This current PR correponds to the IO interface modification, which could impact your project (also if I think/hope it will not). #1141 will remove the directive RVFI_* inside the cva6 core. After I will submit DROMAJO directive removal, then FIRESIM, ... Currently I have not yet addressed the PITON_* directive, we should discuss how to proceed.

Jbalkind:
Got it. Then I don't think I have any issues and am happy to approve! If you could modify Flist.ariane to also move the ariane.sv file that would be nice but otherwise we can handle it later.

Jbalkind (approved):

sjthales:
Ok for me. Approved on my side.

PR #1169, opened 2023-04-04
Support multiple outstanding load operations to the Dcache

cfuguet:
The ID in the request from the load/store unit must be mirrored by the Dcache in the response. This allows to match a given response to its corresponding request. Responses can be given (by the Dcache) in a different order that the one of requests.

This modification introduces a pending load table that tracks outstanding load operations to the Dcache. The depth of this table is a parameter in the target configuration package.


JeanRochCoulon:
@cfuguet    As the parametrization work is completed (AXI and other), I think we are ready to work on this PR. Can you rebase it ?

cfuguet:
Hello @JeanRochCoulon,

Yes, I will push a newer version in a few days.

Thanks

cfuguet:
This PR is now reopened into #1348

PR #1172, opened 2023-04-06
ci: Propagate target to Makefile

niwis:
The Github ci Matrix defines different CVA6 targets. However, these are currently not propagated to the Makefile, meaning that the Makefile always falls back to the default target (`cv64a6_imafdc_sv39`)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,4 +23,4 @@ jobs:
     - name: Prepare
       run: ci/setup.sh
     - name: run tests
-      run: make run-${{ matrix.testcase}}-verilator defines=${{ matrix.cache }}
+      run: make run-${{ matrix.testcase}}-verilator target=${{ matrix.target }}


JeanRochCoulon:
Good catch, I approve and will merge as soon as the CI passes

JeanRochCoulon:
The Thales CI is green, but not the GitHub Actions CI... Can you relaunch it ?

niwis:
I think this is an actual failure of `rv64uf-v-fadd` for `cv32a60x`

JeanRochCoulon:
Can you rebase to relaunch the CI ?

niwis:
Looks like a [timeout](https://github.com/openhwgroup/cva6/actions/runs/4715647808/jobs/8362771438?pr=1172#step:4:4227) of `rv64um-v-div` (target `cv32a60x`)

niwis:
That's possible. I'm not the one trying to add cv32a6 to the github CI, though. I'm just trying to make it do what it already suggests. Executing the same regression tests three times for the same configuration on every commit and PR (which it currently does) is a waste of energy, IMO. So I strongly recommend fixing this (which this PR attempts) or removing cv32a6 as a target from the github CI.

JeanRochCoulon:
To my mind passing the 32 bits tests is not so easy. To setup it on THales CI it was not quite long. Can I ask you to submit a GitHub issue to log the enhancement and to modify the PR by removing the Cache define which should not be needed since #1127 has been merged.

PR #1184, opened 2023-04-13
ptw: Wait for rvalid on flush

niwis:
When the PTW is flushed, we need any pending transactions to complete before returning to the `IDLE` state. Currently, the PTW returns to`IDLE` after one cycle. Remain in `WAIT_RVALID` until we have actually received `rvalid`.


JeanRochCoulon:
@zarubaf    As MMU is not a very known module from my side, can you approve the modification ?

zarubaf (approved):
Thanks and indeed - very strange we never saw an error with that. How did you find this?

niwis:
@hs-yun and @morenes found this while testing a new formal tool we've been working on (more on that to come :)) I guess we never saw an error because you'd need a multi-cycle `flush_i` or multiple `flush_i` shortly after one another, which doesn't seem to happen too often.

Issue #1180, opened 2023-04-11
[BUG] remove ifndef VERILATOR directives


JeanRochCoulon:
### Is there an existing CVA6 bug for this?

- [X] I have searched the existing bug issues

### Bug Description

Hello,

In CVA6 core, plenty of ifndef VERILATOR conditions RTL. Are asserts supported by new Verilator 5 version ? If it is the case, we could remove them from RTL.  

core/include/instr_tracer_pkg.sv:`ifndef VERILATOR
core/include/ariane_pkg.sv:      `ifndef VERILATOR
core/csr_regfile.sv:    `ifndef VERILATOR
core/mmu_sv39/tlb.sv:    `ifndef VERILATOR
core/mmu_sv32/cva6_tlb_sv32.sv:    `ifndef VERILATOR
core/load_unit.sv:`ifndef VERILATOR
core/store_buffer.sv:    `ifndef VERILATOR
core/issue_read_operands.sv:    `ifndef VERILATOR
core/cache_subsystem/cva6_icache.sv:`ifndef VERILATOR
core/cache_subsystem/wt_dcache_missunit.sv:`ifndef VERILATOR
core/cache_subsystem/cache_ctrl.sv:    `ifndef VERILATOR
core/cache_subsystem/wt_dcache_wbuffer.sv:`ifndef VERILATOR
core/cache_subsystem/wt_axi_adapter.sv:`ifndef VERILATOR
core/cache_subsystem/tag_cmp.sv:        `ifndef VERILATOR
core/cache_subsystem/std_cache_subsystem.sv:`ifndef VERILATOR
core/cache_subsystem/wt_dcache_mem.sv:`ifndef VERILATOR
core/cache_subsystem/wt_l15_adapter.sv:`ifndef VERILATOR
core/cache_subsystem/wt_cache_subsystem.sv:`ifndef VERILATOR
core/cache_subsystem/miss_handler.sv:    `ifndef VERILATOR
core/cache_subsystem/miss_handler.sv:    `ifndef VERILATOR
core/cache_subsystem/wt_dcache.sv:`ifndef VERILATOR
core/cache_subsystem/wt_dcache_ctrl.sv:`ifndef VERILATOR
core/axi_shim.sv:`ifndef VERILATOR
core/scoreboard.sv:  `ifndef VERILATOR
core/frontend/instr_queue.sv:  `ifndef VERILATOR
core/frontend/frontend.sv:    `ifndef VERILATOR
[...]

PR #1185, opened 2023-04-14
Enable assertions in Verilator after migrating to Verilator v5, part 1: core/*.sv

zchamski:
Goal: remove all `ifndef VERILATOR` directives that disabled Verilog assertions when using Verilator v4 and below (cf. issue https://github.com/openhwgroup/cva6/issues/1180).

### DIFF
--- a/core/axi_shim.sv
+++ b/core/axi_shim.sv
@@ -293,14 +293,12 @@ module axi_shim #(
 // ----------------
 
 //pragma translate_off
-`ifndef VERILATOR
-   initial begin
-    assert (AxiNumWords >= 1) else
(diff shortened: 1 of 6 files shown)


JeanRochCoulon:
Zbigniew, I have converted into DRAFT version to avoid merging it when it is "Work In Process". When you are ready to merge, please remove the Draft version

zchamski:
This is the first installment of removal of `ifndef VERILATOR`, limited to files `core/*.sv`.  Changes to files located in subdirectories and vendorized modules will follow in separate PRs.

zchamski:
Force-pushed the rebased commit as Github does not like local merges that much...

PR #84, opened 2018-08-23
Merge ariane_next into master

zarubaf:
- Add official debug module
- Flatten tb submodules
- Remove unnecessary AXI logic
- Unify questa and verilator testbench
- CI enhancements (adding riscv benchmarks)

PR #1188, opened 2023-04-15
Remove FORMAL directive from pmp module, as Verilator 5.0 supports assert

JeanRochCoulon:


JeanRochCoulon:
@Moschn    As Verilator 5.x supports assert, I propose to remove FORMAL directive from PMP module. As it, teh asserts will be always enabled.  Can you approve the PR?

Moschn:
LGTM

christian-lanius:
I think that the last assertion [#84](https://github.com/openhwgroup/cva6/blob/master/core/pmp/src/pmp.sv#L84) is wrongly triggered if simulated using questasim. It gets triggered at time 0, so while the system is going into reset.

PR #1189, opened 2023-04-17
Fix reference to ariane.sv in Bender.yml

flaviens:
Hi there!

It looks like Bender.yml references the wrong path to ariane.sv.
EDIT: Maybe you want to move it to the (commented) test section.

Thanks!
Flavien


zarubaf (approved):
LGTM

PR #1190, opened 2023-04-17
Bender.yml: Correct path and location of rvfi_pkg

flaviens:
Hi there!

This PR intends to fix 2 things
a) Fix the path to the rvfi package sv file.
b) Enforce the correct dependency order between packages.

Thanks!
Flavien

### DIFF
--- a/Bender.yml
+++ b/Bender.yml
@@ -25,6 +25,7 @@ sources:
         files:
           - core/include/cv64a6_imafdc_sv39_config_pkg.sv
           - core/include/riscv_pkg.sv
+          - common/local/rvfi/rvfi_pkg.sv
           - core/include/ariane_dm_pkg.sv
           - core/include/ariane_pkg.sv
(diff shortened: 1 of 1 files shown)


flaviens:
Hi @JeanRochCoulon, it's a pleasure if I can make people's lives a bit easier :D and thank you for coming back quickly.

A not super elegant solution would be to write a simple flow `bender`  -> `verilator`.
I think in such a flow:
- `bender` will validate you the syntactic integrity and the dependencies, and will notify dep conflicts.
- `verilator` will notify a lot of issues such as redefinitions and missing modules (you probably want to ignore some of what it says).

It will detect most of the problems (but not check equivalence with the design you intended, as long as you get a design Verilator is happy with).

I usually use morty in between, not sure it's useful.

In any case I'm far from an expert in any of these things.
Some people from Prof. Benini's team may have far more informed suggestions.
Hope it helps!

zarubaf (approved):
👍

PR #1199, opened 2023-04-18
Fix satp spelling in documentation

flaviens:
Title self-explaining


zarubaf:
@flaviens thanks a lot. Do you think you could get the eclipse check to pass for future contributions, that would be amazing!

PR #1204, opened 2023-04-24
Remove DROMAJO

JeanRochCoulon:


zarubaf (approved):
Would be eventually nice to (eventually) have something similar with a more streamlined implementation. But agree that we should not keep unmaintained things! Thanks for the clean up

PR #1215, opened 2023-04-28
Pass AXI parameters uniquely through the CVA6 parameters

cfuguet:
These modifications remove redundant definitions of AXI parameters in the CVA6 core. AXI parameters are only passed through the top parameters and propagated.

Actual definition of AXI parameters must be performed at SoC level and pass during the CVA6 instantiation. This allows different instances of CVA6 in the same SoC to have different AXI parameters.


zarubaf (approved):
This is great! Thanks so much for the (much needed clean-up). Now things are way more organized and nicer!

zarubaf:
From my PoV we just need to fix Thales CI and then we are good to merge!

cfuguet:
@JeanRochCoulon can you point me out which tests are not succeeding in Thales CI ?

Thanks,

JeanRochCoulon:
Thank you for this useful PR. Two things make the CI fail
- The CVA6 is inserted in testharness testbench but also in UVM testbench which belongs to the core-v-verif repository. As the interface port has changed, the UVM simulations crash (in dashboard 0/0 means no simulation has been done because crash). the core-v-verif need to be modified too.
- By default AXI is configured with logic typedef. CVA6 is synthetized with default configuration. And when simulating in gate, the CVA6 is inserted in UVM testbench which is not in default configuration, the interface port is wrong and the simulation fails.
I hope this can help

JeanRochCoulon:
@cfuguet  If needed, we can schedule a 30 minutes discussion

cfuguet:
Hi @JeanRochCoulon,

Yes, let's follow-up by mail to setup a meeting.

Thanks

cfuguet:
This PR is replaced by PR #1223.

The new PR follows branch name guidelines to make sure that the core-v-verif and CVA6 run both with the correct branches that contain modifications in both repositories.

PR #1240, opened 2023-05-17
Fix Questasim execution with rvfi_tracer

MarioOpenHWGroup:
The Questa execution was not working:
  - The binary was not loaded in memory.
  - The $(tb) files were not compiled, those enable the RVFI functionality.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -284,11 +284,11 @@ build: $(library) $(library)/.build-srcs $(library)/.build-tb $(dpi-library)/ari
 
 # src files
 $(library)/.build-srcs: $(library)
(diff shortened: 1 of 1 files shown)


JeanRochCoulon:
Great @MarioOpenHWGroup !!    Thank you for the contribution. It would be useful to add some Questasim jobs into core-v-verif thales CI too. Do you think it is feasible ?

MarioOpenHWGroup:
> Great @MarioOpenHWGroup !! Thank you for the contribution. It would be useful to add some Questasim jobs into core-v-verif thales CI too. Do you think it is feasible ?

Yes sure, we can add a simple test to ensure that it does not break again :+1:

PR #1262, opened 2023-06-12
ZiCond Extension

asimahsan1990:
This PR is for  implementation of [Zicond](https://github.com/riscv/riscv-zicond/releases/tag/v1.0-rc2)  ratified v1.0 (conditional integer operations) extension. As for the verification, I have just verified at the unit level.  Currently, Zicond is in progress on the GCC (a patch is available). On LLVM, it is available in the current version(17).

The "Conditional" operations extension provides a simple solution that provides most of the benefits and flexibility one would desire to support conditional arithmetic and conditional-select/move operations while remaining true to the RISC-V design philosophy. The instructions follow the format for R-type instructions with three operands (i.e., two source operands and one destination operand). Using these instructions, branch-less sequences can be implemented (typically in two-instruction sequences) without the need for instruction fusion, special provisions during the decoding of architectural instructions, or other micro-architectural provisions.


asimahsan1990:
@JeanRochCoulon

JeanRochCoulon:
Great contribution @asimahsan1990 !!

JeanRochCoulon:
As discussed during CVA6 verification meeting, I propose to enable Zicond for 64 bit configuration, the objective is to be able to use it on at least one CVA6 configuration.
BTW, which tests could use Zicond, are there some tests from arch-test which can be reused ?

asimahsan1990:
> As discussed during CVA6 verification meeting, I propose to enable Zicond for 64 bit configuration, the objective is to be able to use it on at least one CVA6 configuration. BTW, which tests could use Zicond, are there some tests from arch-test which can be reused ?

On Arch-test, there is [PR](https://github.com/riscv-non-isa/riscv-arch-test/tree/main/riscv-test-suite) open; those tests be available soon. Currently, there aren't any tests available that we can use.

fatimasaleem:
Thank you @asimahsan1990 for making the required changes. It will be good to run the available tests (https://github.com/riscv-non-isa/riscv-arch-test/pull/321) against spike as the reference model (https://github.com/riscv-software-src/riscv-isa-sim/pull/1241) and mention the test status in the PR description as well.

AyoubJalali:
@JeanRochCoulon what we should do about the code coverage.

PR #1263, opened 2023-06-14
serdiv - Reduced worst-case latency 

Seek64:
## Description
In the current version, the serdiv module takes between 1 and WIDTH+2 cycles from the incoming request (_in_vld_i_) until the assertion of the valid signal (_out_vld_o_).
The worst case scenario is a division by 0 or -1, as the counter gets set to WIDTH (see line 108).
By introducing fast paths for these two cases (similar to _div_res_zero_), the maximum latency can be reduced to WIDTH+1 and divisions by 0 or -1 only take a single cycle now.
We verified the functionality of both the original and proposed design with a set of formal properties.
We have not checked if these changes affect the critical path in any way.

## Changes
- Added a new register _op_b_neg_one_q_ which indicates if b is -1. 
  The _lzc_b_no_one_ signal is now used to determine if b is 0 or -1.
- Introduced a fast path for 0 and -1 by adjusting the FSM and _out_mux_ accordingly.
- Reduced the counter width by 1, as its maximum value changed from WIDTH to WIDTH-1.

### DIFF
--- a/core/serdiv.sv
+++ b/core/serdiv.sv
@@ -47,6 +47,7 @@ module serdiv import ariane_pkg::*; #(
   logic [WIDTH-1:0]       op_b_q, op_b_d;
   logic                   op_a_sign, op_b_sign;
   logic                   op_b_zero, op_b_zero_q, op_b_zero_d;
+  logic                   op_b_neg_one, op_b_neg_one_q, op_b_neg_one_d;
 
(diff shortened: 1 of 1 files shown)


JeanRochCoulon:
Great !!   Thank you for your clear explanation of your improvement. I have checked the non regression of frequency. Goog job.

PR #1265, opened 2023-06-19
Remove instr_tracer and keep RVFI tracer

JeanRochCoulon:


MikeOpenHWGroup (approved):
LGTM @JeanRochCoulon.  I am approving this PR, but please do not merge based only on my approval - I am just the verification geek and we should obtain an approval from more qualified team members.

Jbalkind:
I'm fairly sure that we use this in openpiton?

JeanRochCoulon:
@Jbalkind   Of course, in relation to habits, some users/designers use the old tracer. But RVFI is the common tracer used on OpenHW cores and we duplicate the number of tracer: time consuming to support all of them. I am sure they can switch to RVFI tracer without any problem. Please warm me if it would not be the case. The plan is to merge on next Monday. Cheers

zarubaf:
I'd be happy to get rid of this ugly beast, but can you enlighten me as to what the alternative is?

Jbalkind:
Can you please not merge this with just two days' notice? We need time to figure out a solution. I don't really see the urgency here

JeanRochCoulon:
Hello @Jbalkind   There is more urgency on #1267. Thanks

Jbalkind:
I really don't see a strong justification for getting rid of this. There are a few files which are barely touched (3, 4 months, 2 years, 2 years ago in terms of changes based on the files I look at). There is virtually no maintenance effort exerted here. As a downstream user who does not want to require users to have spike installed to get a useful instruction trace, I really do not want to see this removed. It's not even adding ports like some other things do.

PR #1267, opened 2023-06-20
merge axi and l15 ports into noc port

JeanRochCoulon:

### DIFF
--- a/core/cache_subsystem/wt_cache_subsystem.sv
+++ b/core/cache_subsystem/wt_cache_subsystem.sv
@@ -25,8 +25,8 @@ module wt_cache_subsystem import ariane_pkg::*; import wt_cache_pkg::*; #(
   parameter int unsigned AxiAddrWidth = 0,
   parameter int unsigned AxiDataWidth = 0,
   parameter int unsigned AxiIdWidth   = 0,
-  parameter type axi_req_t = ariane_axi::req_t,
-  parameter type axi_rsp_t = ariane_axi::resp_t
+  parameter type noc_req_t = ariane_axi::req_t,
+  parameter type noc_resp_t = ariane_axi::resp_t
 ) (
   input logic                            clk_i,
   input logic                            rst_ni,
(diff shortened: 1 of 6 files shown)


MikeOpenHWGroup (approved):
LGTM @JeanRochCoulon. I am approving this PR, but please do not merge based only on my approval - I am just the verification geek and we should obtain an approval from more qualified team members.

JeanRochCoulon:
@cathales do have any idea ?

JeanRochCoulon:
@Jbalkind  As I have no way to test the modification in openpiton/ariane_verilog_wrap.sv, could you have a look and maybe try it? Thanks

zarubaf (approved):
looks very good!

Jbalkind:
I just cleaned up some syntax errors but the design will not build and pass tests. I will have to spend some time to dig into this and figure out what's broken. The core makes its first request and receives the response but then it gets stuck.

jquevremont:
CVA6 does not come with a NoC (i.e. a complex interconnect). I suggest to use the "socket" word instead to qualify this interface, as it can be used to plug CVA6 on a bus or a NoC.

core/cache_subsystem/wt_cache_subsystem.sv
```diff
@@ -153,16 +147,16 @@ module wt_cache_subsystem import ariane_pkg::*; import wt_cache_pkg::*; #(
     .dcache_data_i      ( dcache_adapter          ),
     .dcache_rtrn_vld_o  ( adapter_dcache_rtrn_vld ),
     .dcache_rtrn_o      ( adapter_dcache          ),
-    .l15_req_o          ( l15_req_o               ),
-    .l15_rtrn_i         ( l15_rtrn_i              )
+    .l15_req_o          ( noc_req_o               ),
+    .l15_rtrn_i         ( noc_rtrn_i              )
```
Jbalkind, line 151: This is a typo - should be noc_resp_i instead of noc_rtrn_i. If you fix this, our simulation seems to be passing once again

Jbalkind:
I just made a PR to the ThalesSiliconSecurity/cva6 fork for fixes that make this PR work with openpiton. I don't think any of the fixes involve anything used outside of openpiton so it should be trivial to merge

Jbalkind:
> CVA6 does not come with a NoC (i.e. a complex interconnect). I suggest to use the "socket" word instead to qualify this interface, as it can be used to plug CVA6 on a bus or a NoC.

We discussed this when we had the meeting proposing this change and agreed on noc. I'm not a huge fan of socket as it's used for other purposes in other cva6 users' projects (e.g. ESP, DECADES). Perhaps "memintf" or "l1intf" or similar?

PR #1272, opened 2023-06-20
Removing bootrom sv files

eyssartk:
The top makefile as it is written does not regenerate these files as they are already present. Unfortunately, they have not been updated with the rest of the bootrom source code. Especially with the device tree update from #806. Deleting them force the makefile to recreate them.


JeanRochCoulon:
Hi @eyssartk     The CI is failed. I copied the log:
image
Have a good day !

cathales:
To make the CI pass, maybe we need to generate the bootrom via its makefile here? https://github.com/openhwgroup/cva6/blob/8d9e0c4738c1b8d2c4b2821f121118448dacac66/Makefile#L656

```makefile
make -C corev_apu/fpga/src/bootrom/
```

cathales:
Hello, any news about this PR? What do you think about my suggestion?

eyssartk:
Hello,
Yes this would solve the problem. 
Some explanation about the reason I opened this PR:

- The bootrom_32.sv is not aligned with the cv32a6.dts. 
#806  does not update the sv file.
- If you add bootrom or dts modifications, they are not taken into account. That creates confusions.

Of course, I agree that it would require all "hardware people" to have a functioning RISC-V toolchain.

JeanRochCoulon:
This PR need to be updated to fix the bug. The issue is tracked by #1485. I close the issue hoping somebody will fix it in the coming weeks.

PR #1276, opened 2023-06-21
Makefile/questa: Pass elf-bin to fesvr

niwis:
`elf-bin` is propagated to fesvr via the plusplus arg. Since #1240, running
```
make sim elf-bin=path/to/elf
```
returns
```
terminate called after throwing an instance of 'std::runtime_error'
  what():  could not open
+PRELOAD=path/to/elf (did you misspell it? If VCS, did you forget +permissive/+permissive-off?)
```

As per the README, if preloading is needed, `make sim preload=path/to/elf` should be used instead (https://github.com/openhwgroup/cva6/tree/master#memory-preloading), which sets `+PRELOAD=path/to/elf`:
https://github.com/openhwgroup/cva6/blob/6c89fda0da14e188c43c818be3af15123c82fc69/Makefile#L249-L253.


zarubaf (approved):
Top, thanks! 👍

PR #1277, opened 2023-06-21
icache_axi_wrapper: Cast paddr to AxiAddr

niwis:
Prevents a negative repetition multiplier if `paddr` width > `AxiAddrWidth`.


zarubaf (approved):
Makes sense! Let's wait on the Thales CI 👍

PR #1278, opened 2023-06-22
declare cva6_cfg_t to pass the configuration through the hierarchy

JeanRochCoulon:
this PR prepares the parametrization PRs

### DIFF
--- a/core/alu.sv
+++ b/core/alu.sv
@@ -18,7 +18,9 @@
 // Description: Ariane ALU based on RI5CY's ALU
 
 
-module alu import ariane_pkg::*;(
+module alu import ariane_pkg::*; #(
+    parameter ariane_pkg::cva6_cfg_t cva6_cfg = 0
+) (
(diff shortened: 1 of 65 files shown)


core/alu.sv
```diff
@@ -18,7 +18,9 @@
 // Description: Ariane ALU based on RI5CY's ALU
 
 
-module alu import ariane_pkg::*;(
+module alu import ariane_pkg::*; #(
+    parameter ariane_pkg::cva6_cfg_t cva6_cfg = 0
```
zarubaf, line 22: I think you need to assign all zeros to the `packed struct`. This just works because we have one int unsigned in the dummy struct at the moment.
```suggestion
    parameter ariane_pkg::cva6_cfg_t cva6_cfg = '0
```

JeanRochCoulon: Yes, of course, you are right. I will fix it.

PR #1284, opened 2023-06-28
More information during the interfaces modification.

jquevremont:
Provide more information to CVA6 users.


jquevremont:
README.md modification only. I guess the CI failure is unrelated and can be safely ignored.

PR #1288, opened 2023-06-30
Add coverage pragmas to exclude BITMANIP code

JeanRochCoulon:
To increase the code coverage of CVA6, pragmas are inserted to disable some code lines. Pragmas are related to VCS tool, that is not so nice, but other solutions have been studied wihtout any success (macro approach is complex,  function is not supported by all tools)

### DIFF
--- a/core/alu.sv
+++ b/core/alu.sv
@@ -73,6 +73,7 @@ module alu import ariane_pkg::*;(
     always_comb begin
       operand_a_bitmanip = fu_data_i.operand_a;
 
+      //VCS coverage off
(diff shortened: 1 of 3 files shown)


JeanRochCoulon:
@mike, can you approve it ?

MikeOpenHWGroup (approved):
I am approving/merging this PR in order to allow forward progress.  Having said that, it would be good to investigate other ways to create code-coverage exclusions without the use of pragmas (which are difficult to maintain and may cause future projects to miss coverage they would otherwise target).

See task #1292 for more information.

Issue #1202, opened 2023-04-20
CVA6 Custom CSRs Address Configuration Issue 


spidugu444:
### Is there an existing CVA6 bug for this?

- [X] I have searched the existing bug issues

### Bug Description

Hi,

While verifying access modes for CVA6 CSRs, I found an issue on custom CSRs(ICACHE and DCACHE) address configuration as mentioned below.

As per riscv priviliged specification address for RW custom CSRs will be in range of 0x7C0-0x7FF or 0xBC0-0xBFF, but in CVA6 core address for custom CSRs are in range of **0x700-0x77F** which is for Standard read/write.

please find below snaps for better understanding. 

CVA6 Custom CSRs address range as per [CVA6 user manual](https://github.com/openhwgroup/cva6/issues/new?assignees=&labels=Type%3ABug&template=bug.yaml&title=%5BBUG%5D+%3Ctitle%3E) is as shown below.

![image](https://user-images.githubusercontent.com/103561542/233372060-22c1fede-edd0-4d04-83d1-0a32bfa19e26.png)

As per riscv priviliged specification address range will be as shown below.
![image](https://user-images.githubusercontent.com/103561542/233371257-3c7aa35a-a669-4d30-bb37-2fd6c9d6a628.png)

Thanks,

spidugu444:
@RanjanThales

frikhaAziz:
I pushed a commit to solve this issue. Please review it and close it if there is no problem. @spidugu444

spidugu444:
sure @frikhaAziz, I will let you know.

spidugu444:
@frikhaAziz, good day. Your RTL tweaks have allowed me to access the ICACHE and DCACHE CSRs at a custom read/write address range at 0x7C0-0x7C1.

spidugu444:
@JeanRochCoulon, On the spike end, the custom CSRs ICACHE and DCACHE have not been implemnted.

MikeOpenHWGroup:
Hi @spidugu444, it seems that the original problem reported in this issue has been solved by @frikhaAziz.  If you agree please:
- close this issue.
- open a new issue against the Spike model to get the custom CSRs implemented.

spidugu444:
Closing this issue as I raised a new [issue](https://github.com/openhwgroup/cva6/issues/1346) for CVA6 CSRs spike discrepancies.

PR #1290, opened 2023-06-30
Changing Addresses of custom registers ICACHE and DCACHE.

frikhaAziz:
ICACHE and DCACHE have an address in 0x700-0x77F which is a Standard read/write in M-mode. But ICACHE and DCACHE are Custom registers so they need to be Custom read/write. In fact there is two possibilities 0x7C0-0x7FF and 0xBC0-0xBFF. I chose the first one 0x7C0-0x7FF. 
ICACHE: 0x700==>0x7C0. 
DCACHE: 0X701==>0x7C1.  
RTL and Spec are changed to solve #1202 issue.


zarubaf (approved):
Very good point!

PR #1296, opened 2023-07-04
Start populating more CODEOWNERS

zarubaf:
I just figured out that apparently `CODEOWNERS` need to have write permissions to the repository. There are a couple of feature requests on Github to also enable the `triage` group to be listed but that hasn't been implemented yet.

I suggest starting populating it so that we are ready once that feature is implemented. And for all committers @JeanRochCoulon, @jquevremont, and @zarubaf it would already take effect.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,5 +1,16 @@
 # Global Owners
-*  @zarubaf
+*  @JeanRochCoulon @zarubaf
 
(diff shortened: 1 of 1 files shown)


JeanRochCoulon:
@zarubaf   I suggest to identify cache and fpu codeowners.  Do you have contacts with the stakeholders ?

PR #1299, opened 2023-07-05
Added ZiCond Extension Support

fatimasaleem:
This PR is for implementation of [Zicond](https://github.com/riscv/riscv-zicond/releases/tag/v1.0-rc2) ratified v1.0 (conditional integer operations) extension. As for the verification, I have just verified at the unit level. Currently, Zicond is in progress on the GCC (a patch is available). On LLVM, it is available in the current version(17).

The "Conditional" operations extension provides a simple solution that provides most of the benefits and flexibility one would desire to support conditional arithmetic and conditional-select/move operations while remaining true to the RISC-V design philosophy. The instructions follow the format for R-type instructions with three operands (i.e., two source operands and one destination operand). Using these instructions, branch-less sequences can be implemented (typically in two-instruction sequences) without the need for instruction fusion, special provisions during the decoding of architectural instructions, or other micro-architectural provisions.

Authored by: [Asim Ahsan](https://github.com/asimahsan1990)

### DIFF
--- a/core/alu.sv
+++ b/core/alu.sv
@@ -302,6 +302,13 @@ module alu import ariane_pkg::*; #(
                 default: ; // default case to suppress unique warning
             endcase
         end
+
+        if (ariane_pkg::RCONDEXT) begin
+           unique case (fu_data_i.operation)
+           CZERO_EQZ : result_o = (fu_data_i.operand_b) ? (fu_data_i.operand_a) : (0) ; // move zero to rd if rs2 is equal to zero else rs1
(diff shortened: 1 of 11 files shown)


JeanRochCoulon:
Hello @fatimasaleem    Do not warry about the Thales CI status, the server is down. I will warm you when it will be fixed !

fatimasaleem:
[Abdulwadoodd](https://github.com/Abdulwadoodd) Once you confirm the status of [Zicond arch-tests](https://github.com/riscv-non-isa/riscv-arch-test/pull/321), I will mark this PR ready.

PR #1302, opened 2023-07-11
Fix infinite loop in std_cache_pkg::one_hot_to_bin (resolves #1301)

### LINKED ISSUE #1301: [BUG] Infinite loop in std_cache_pkg::one_hot_to_bin()

### Bug Description

If the `in` input to the [std_cache_pkg::one_hot_to_bin()](https://github.com/openhwgroup/cva6/blob/513bb91f823a18a6653a1685acd60408049c9c8f/core/include/std_cache_pkg.sv#L78) function is 0 and  `ariane_pkg::DCACHE_SET_ASSOC` is a power of 2 then this function enters an infinite loop since the loop variable`i` wraps from DCACHE_SET_ASSOC-1 to 0 and the loop condition is always true.

Changing the type of `i` to `int unsigned` resolves this issue.

maxbjurling:
Use `int unsigned` for loop variable `i` to avoid wrapping before loop exit condition is met.

Resolves bug https://github.com/openhwgroup/cva6/issues/1301

### DIFF
--- a/core/include/std_cache_pkg.sv
+++ b/core/include/std_cache_pkg.sv
@@ -78,7 +78,7 @@ package std_cache_pkg;
     function automatic logic [$clog2(ariane_pkg::DCACHE_SET_ASSOC)-1:0] one_hot_to_bin (
         input logic [ariane_pkg::DCACHE_SET_ASSOC-1:0] in
     );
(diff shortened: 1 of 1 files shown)


JeanRochCoulon:
Thank you for this bug fix !!

PR #1305, opened 2023-07-14
Move Cache Invalidation and FFlag Signals into `acc_dispatcher`

domenicw:
This PR moves the data cache invalidation and fflag signals used when `ENABLE_ACCELERATOR = 1` from the top module into the `acc_dispatcher` module. This change is related to PR #1024.

Doing this, projects that do not use the invalidation interface or the fflag signals can just tie them off. This also means that the request and response structs can leave these signals out if they are not needed.


zarubaf (approved):
That looks good to me! Thanks

zarubaf:
FYI @niwis this will affect Ara :-)

PR #1309, opened 2023-07-19
Adding configs for Polara

hossein1387:
This PR includes configs for CVA6 to be used inside Polara project.

### DIFF
--- a/core/include/cv64a6_imafdc_sv39_openpiton_config_pkg.sv
+++ b/core/include/cv64a6_imafdc_sv39_openpiton_config_pkg.sv
@@ -18,7 +18,7 @@ package cva6_config_pkg;
     localparam CVA6ConfigXlen = 64;
 
     localparam CVA6ConfigFpuEn = 1;
-    localparam CVA6ConfigF16En = 1;
+    localparam CVA6ConfigF16En = 0;
     localparam CVA6ConfigF16AltEn = 0;
     localparam CVA6ConfigF8En = 0;
     localparam CVA6ConfigFVecEn = 0;
@@ -27,7 +27,7 @@ package cva6_config_pkg;
     localparam CVA6ConfigCExtEn = 1;
     localparam CVA6ConfigAExtEn = 1;
     localparam CVA6ConfigBExtEn = 0;
-    localparam CVA6ConfigVExtEn = 1;
+    localparam CVA6ConfigVExtEn = 0;
 
     localparam CVA6ConfigAxiIdWidth = 4;
     localparam CVA6ConfigAxiAddrWidth = 64;
--- a/core/include/cv64a6_imafdc_sv39_openpiton_config_pkg.sv
+++ b/core/include/cv64a6_imafdc_sv39_openpiton_config_pkg.sv
@@ -18,7 +18,7 @@ package cva6_config_pkg;
     localparam CVA6ConfigXlen = 64;
 
     localparam CVA6ConfigFpuEn = 1;
-    localparam CVA6ConfigF16En = 0;
+    localparam CVA6ConfigF16En = 1;
     localparam CVA6ConfigF16AltEn = 0;
     localparam CVA6ConfigF8En = 0;
     localparam CVA6ConfigFVecEn = 0;
--- a/core/include/cv64a6_imafdc_sv39_openpiton_config_pkg.sv
+++ b/core/include/cv64a6_imafdc_sv39_openpiton_config_pkg.sv
@@ -27,7 +27,7 @@ package cva6_config_pkg;
     localparam CVA6ConfigCExtEn = 1;
     localparam CVA6ConfigAExtEn = 1;
     localparam CVA6ConfigBExtEn = 0;
-    localparam CVA6ConfigVExtEn = 0;
+    localparam CVA6ConfigVExtEn = 1;
 
     localparam CVA6ConfigAxiIdWidth = 4;
     localparam CVA6ConfigAxiAddrWidth = 64;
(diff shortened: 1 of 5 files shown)


JeanRochCoulon:
Hello @hossein1387   please to meet you.
It seems that your PR has not been rebased. Could you rebase with the newest master branch ?!?  Thanks

Jbalkind:
Also the PR seems to modify the openpiton config - this shouldn't be the case

hossein1387:
@JeanRochCoulon Please to meet you as well. I just rebased my branch. 
@Jbalkind With the new changes, I only have to enable FP16 in CVA6 config. The reason for that is by default, Ara supports half precision and throws an error if CVA6 does not support it. have a look  [here](https://github.com/PolyMTL-Gr2m/ara/blob/ee887bdb20ce396d438639a85bc2640718806fdc/hardware/src/ara.sv#L485). Of course, this may not be the Openpiton's default config and its more of Polara configuration. In that case, should move the configuration to a new file `cv64a6_imafdc_sv39_polara_config_pkg`?

JeanRochCoulon:
The three files you modify are not used by Thales CI.
LGTM. @zarubaf, as I am out of office up to August 11th, do not hesitate to merge PR (when @hossein1387 is ready to merge).

zarubaf:
Would it be possible to maintain `cv64a6_imafdc_sv39_polara_config_pkg` outside of CVA6 or do you see a strong reason why it should live here?

@JeanRochCoulon got it.

JeanRochCoulon:
For info, our long term view is to replace the config package file by the in parameter configuration.

Jbalkind:
Given Polara is an OpenHW supported tapeout, I'm inclined to say we should put a polara config and let it evaporate with the rest when they do. Maintaining it independently while we're in the tapeout process and the configs keep changing sounds like a pain.

hossein1387:
Based on the discussion with @Jbalkind I added a new config file for Polara. After rebasing with master, I noticed that the CVX interface on ariane is not propagated to the interface list. That is required when I need to integrate it with Ara in Openpiton. Please let me know what you think.

corev_apu/src/ariane.sv
```diff
@@ -16,6 +16,8 @@
 module ariane import ariane_pkg::*; #(
   parameter ariane_pkg::cva6_cfg_t CVA6Cfg = cva6_cfg_empty,
   parameter type rvfi_instr_t = logic,
+  parameter type cvxif_req_t = acc_pkg::accelerator_req_t,
+  parameter type cvxif_resp_t = acc_pkg::accelerator_resp_t,
```
Jbalkind, line 20: I think this should probably default to `logic` to fix CI

corev_apu/src/ariane.sv
```diff
@@ -42,24 +44,26 @@ module ariane import ariane_pkg::*; #(
   // RISC-V formal interface port (`rvfi`):
   // Can be left open when formal tracing is not needed.
   output rvfi_instr_t [CVA6Cfg.NrCommitPorts-1:0] rvfi_o,
+  //
+  output cvxif_pkg::cvxif_req_t        cvxif_req_o,
+  input cvxif_pkg::cvxif_resp_t        cvxif_resp_i,
```
Jbalkind, line 49: I think this should be `cvxif_req_t` and `cvxif_resp_t`, NOT with `cvxif_pkg::`

core/include/cv64a6_imafdc_sv39_openpiton_config_pkg.sv
```diff
@@ -18,7 +18,7 @@ package cva6_config_pkg;
     localparam CVA6ConfigXlen = 64;
 
     localparam CVA6ConfigFpuEn = 1;
-    localparam CVA6ConfigF16En = 0;
+    localparam CVA6ConfigF16En = 1;
```
Jbalkind, line 21: Let's not change this

core/include/cv64a6_imafdc_sv39_openpiton_config_pkg.sv
```diff
@@ -27,7 +27,7 @@ package cva6_config_pkg;
     localparam CVA6ConfigCExtEn = 1;
     localparam CVA6ConfigAExtEn = 1;
     localparam CVA6ConfigBExtEn = 0;
-    localparam CVA6ConfigVExtEn = 0;
+    localparam CVA6ConfigVExtEn = 1;
```
Jbalkind, line 30: Let's not change this

Jbalkind (changes requested):
I think these suggestions might fix the broken CI

corev_apu/src/ariane.sv
hossein1387: The [issue with](https://github.com/openhwgroup/cva6/actions/runs/5816727072/job/15770256354?pr=1309#step:4:19) CI is adding the `cvx` interface to the port list: [here]( https://github.com/PolyMTL-Gr2m/cva6/blob/cva6-rebase-openpiton/corev_apu/src/ariane.sv#L48)

corev_apu/src/ariane.sv
hossein1387: Thanks for pointing this out! Yoan found this bug the hard way. Simulation went all fine but in Synthesis, we got port mismatch.

core/include/cv64a6_imafdc_sv39_openpiton_config_pkg.sv
hossein1387: Done.

corev_apu/src/ariane.sv
hossein1387: Given that this is Polara specific change, I removed the `cvx` interface from the port list.

JeanRochCoulon:
@hossein1387    Can you sign the Eclipse agreement ?   This is the OpenHW process... Thks

hossein1387:
> @hossein1387 Can you sign the Eclipse agreement ? This is the OpenHW process... Thks

I already have done that.

JeanRochCoulon:
Ok... Sometimes when the PR is submitted before signing the agreement, the PR need to be re-submitted. Nevermind, I merge as it is, we will look at it for the next PR. Thanks @hossein1387.

PR #1313, opened 2023-07-20
HOTFIX : Add mode info in the CVA6 req to the Co-proce

AyoubJalali:
It's Fix recommended by @t0251133


zarubaf:
LGTM, tests are supposed to fail afair. Merging!

PR #1317, opened 2023-07-24
Add New Stall and Flush Signals to `acc_dispatcher`

domenicw:
This PR adds three more ports to the `acc_dispatcher`:
- `acc_stall_st_pending_o` output to stall the output of the store buffer
- `flush_pipeline_o` output to flush the complete pipeline
- `dcache_req_ports_i` input that shows the requests from the load and store unit to the d$

Also, a new `acc_cfg_t` type parameter, as well as a `AccCfg` parameter is added is added to the module. It can be used to pass on parameters from the outside into the module. It is therefore also necessary to add these two parameters to the CVA6 top module parameter list.

One final change is done to the PC calculation, specifically to point 6. If we have to set the PC to the one currently in the commit stage, but the commit stage halt signal is raised (meaning no instruction is committed in this cycle), we do not increment the PC to the next instruction, but instead keep it at the one currently in the commit stage. This allows to do a complete pipeline flush, including any instruction which is currently ready to be committed.


zarubaf (approved):
LGTM, thanks. Let's wait for the Thales CI in the evening.

zarubaf:
Ah, and I think you need a re-base somehow.

domenicw:
Rebase is done. As long as there are no more major PRs like the last two, we should now stay conflict free.

domenicw:
The same Thales-CI tests fail as in any other PR. So I think this one can be merged, right?

JeanRochCoulon:
The following test need to be fixed before merging:
image

zarubaf:
Can you please post the CI Log for us? Or a way to reproduce? We do have VCS in case that helps, but I guess it’s going to be some syntactic thing.

domenicw:
Looks like the second run has fixed the failing test (thanks @JeanRochCoulon).
Sadly the synthesis test is now red. Though I cannot see what the problem is there. Comparing the output to the run before (which was green), I cannot make out any differences. Any guess, @JeanRochCoulon?

domenicw:
Any update on what went wrong with the synthesis test @JeanRochCoulon?

zarubaf:
Since the `kG` equivalent is the same, I am going to merge this.

JeanRochCoulon:
ASIC synthesis need to be solved !!   When job is failed, checks are not played on futur PR. That's a regression.
@zarubaf   can you enter and assign a github issue please.

zarubaf:
Done, not sure I understand exactly but sure we can resolve this :-)

zarubaf:
It's actually interesting that the CI with 127354 GE passed and the final 127180 GE failed... Something is fishy.

Issue #1237, opened 2023-05-16
[BUG] <Unknown Values on performance counters>


AyoubJalali:
### Is there an existing CVA6 bug for this?

- [X] I have searched the existing bug issues

### Bug Description

Hello,
The problem here is the a got unknown values on the performance counters (**mhpmcounter-h & mhpmevent**). 

_core   0: 0x0000000080000b0c (0xb1e792f3) csrrw   t0, mhpmcounter30, a5
3 0x0000000080000b0c (0xb1e792f3) x 5 0xxxxxxxxx_ (example)

so to reproduce the problem you have to : 

- Add the yaml description of the test in [cva6_base_testlist.yaml](https://github.com/openhwgroup/core-v-verif/blob/master/cva6/sim/cva6_base_testlist.yaml)

- test: riscv_arithmetic_basic_hint_illegal_csr_test
  description: >
    Arithmetic instruction test, no load/store/branch instructions
  gen_opts: >
    +instr_cnt=500
    +num_of_sub_program=5
    +directed_instr_0=riscv_int_numeric_corner_stream,4
    +no_data_page=1
    +no_branch_jump=1
    +boot_mode=m
    +no_csr_instr=0
    +enable_illegal_csr_instruction=1
    +hint_instr_ratio=100
    +tvec_alignment=8
  iterations: 2
  gen_test: riscv_instr_base_test
  rtl_test: core_base_test

and run the commend : 

1. python3 cva6.py --testlist=cva6_base_testlist.yaml --test riscv_arithmetic_basic_hint_illegal_csr_test --iss_yaml cva6.yaml --target cv32a60x --iss=vcs-uvm,spike -i 1 --iss_timeout 300 --seed 3018203365 (bug on mhpmcounter12h)

2. python3 cva6.py --testlist=cva6_base_testlist.yaml --test riscv_arithmetic_basic_hint_illegal_csr_test --iss_yaml cva6.yaml --target cv32a60x --iss=vcs-uvm,spike -i 1 --iss_timeout 300 --seed 2951308182 (bug on mhpmcounter15)

3. python3 cva6.py --testlist=cva6_base_testlist.yaml --test riscv_arithmetic_basic_hint_illegal_csr_test --iss_yaml cva6.yaml --target cv32a60x --iss=vcs-uvm,spike -i 1 --iss_timeout 300 --seed 1147271990 (bug on mhpmevent8)

[...]

PR #1329, opened 2023-08-07
Update for #1237

TulikaSi:
Since the performance counters are implemented as an optional feature, this commit takes into consideration the case when the performance counters are disabled i.e in that case the output data from the performance counter module should read as 0 (which was reading as X's previously **#1237**)


TulikaSi:
@ASintzoff

JeanRochCoulon:
Thank you @TulikaSi for this fix !

PR #1333, opened 2023-08-11
m/sret optimization

ezelioli:
**Problem:**

With the current implementation, when an `sret` retires the pipeline stalls for several cycles (>30) due to i-cache misses that are never resolved (since outstanding cache misses are killed on a pipeline flush). 

**Proposed solution:**

Treat `m/sret` operations as direct jumps to PC + 0 when computing `npc_d`. This avoids polluting the instruction cache with useless instructions, and generating spurious i-cache misses.  
The only software workaround I found is to manually insert a `j .` instruction after each `m/sret`. However, this can simply be implemented as a hardware optimization to make ISR performance more reliable. 
Assuming no other source of stalling (i.e. real i-cache miss, TLB miss), this brings the `sret` penalty the same as a pipeline flush (6 cycles).

### DIFF
--- a/core/frontend/instr_scan.sv
+++ b/core/frontend/instr_scan.sv
@@ -39,16 +39,19 @@ module instr_scan #(
     logic rv32_rvc_jal;
     assign rv32_rvc_jal = (riscv::XLEN == 32) & ((instr_i[15:13] == riscv::OpcodeC1Jal) & is_rvc & (instr_i[1:0] == riscv::OpcodeC1));
 
+    logic is_xret;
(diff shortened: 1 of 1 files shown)


JeanRochCoulon:
Hello @ezelioli    Indeed this fixes a kind of performance bug. Thank for detecting it and fixing it !!
To merge could you sign the Eclipse Agreement ?

ezelioli:
Hello @JeanRochCoulon, sure I just signed it !

JeanRochCoulon:
Hello @ezelioli   You have signed the agreement but this did not solved the github action job. I think the PR need to be resubmitted to take into account the modification. Anyway, I merge.

PR #1337, opened 2023-08-17
fix prediction shift for BHT and BTB

cathales:
To be able to issue two instructions in the same cycle, fetching 64 bits each cycle is required to feed the instruction queue (there are not enough compressed instructions in usual code).

With compressed instructions, fetching 64 bits might return up to 4 instructions, which requires a change in the BTB and BHT shapes.  
4 columns means an index of two bits.

There is a mechanism to shift predictions when the access is not aligned.  
For the first item of the result, **the index was using a hardcoded number of one bit, so the returned prediction was not the good one**. This PR fixes that.  
For the subsequent items the result was good.

Without C extension and a 32-bits fetch, `INSTR_PER_FETCH = 1` so `(INSTR_PER_FETCH) = 0` which would cause an issue so the `if`/`else` block is used to handle this specific case.

This modification has no impact on area for current configurations.

### DIFF
--- a/core/cva6.sv
+++ b/core/cva6.sv
@@ -834,7 +834,7 @@ module cva6 import ariane_pkg::*; #(
   // Cache Subsystem
   // -------------------
 
-  if (DCACHE_TYPE == int'(cva6_config_pkg::WT)) begin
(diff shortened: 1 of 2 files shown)


zarubaf (approved):
Awesome! Thanks for this clean fix!

PR #1344, opened 2023-08-24
Generate illegal opcode when execute LWU with XLEN=32

JeanRochCoulon:
LWU was decoded in 32 and 64 bits. As it is a 64 bit instruction, this modification generates an illegal opcode when configuration is 32bit.


cathales:
Spike random crashes happened in some jobs so I restarted them and they passed.

PR #1355, opened 2023-08-31
Decode AMOXOR_D (and similar instructions) only when XLEN=64bits 

### LINKED ISSUE #1352: Decoding RV64 as a RV32 instruction

### Bug Description

I generate some illegal instruction (illegal opcode **_0x232532af_**), and I found that an instruction is actually a **RV64A**  instr, the problem is the **CV32A60X** decode it, without raising an exception.
So I think it's a BUG in the decoder, if you wanna reproduice the BUG you just need to inject the instruction 0x232532af in your test.S

JeanRochCoulon:
It fixes #1352

### DIFF
--- a/core/decoder.sv
+++ b/core/decoder.sv
@@ -1073,7 +1073,7 @@ module decoder import ariane_pkg::*; #(
                     instruction_o.rd[4:0]  = instr.atype.rd;
                     // TODO(zarubaf): Ordering
                     // words
-                    if (CVA6Cfg.RVA && instr.stype.funct3 == 3'h2) begin
+                  if (riscv::IS_XLEN64 && CVA6Cfg.RVA && instr.stype.funct3 == 3'h2) begin
                         unique case (instr.instr[31:27])
                             5'h0:  instruction_o.op = ariane_pkg::AMO_ADDW;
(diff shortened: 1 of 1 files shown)


core/decoder.sv
```diff
@@ -1073,7 +1073,7 @@ module decoder import ariane_pkg::*; #(
                     instruction_o.rd[4:0]  = instr.atype.rd;
                     // TODO(zarubaf): Ordering
                     // words
-                    if (CVA6Cfg.RVA && instr.stype.funct3 == 3'h2) begin
+                    if (riscv::IS_XLEN64 && CVA6Cfg.RVA && instr.stype.funct3 == 3'h2) begin
```
AyoubJalali, line 1076: Hello @JeanRochCoulon, I think only the double words AMO.D instruction that should be decoded when XLEN = 64,
the AMO.W word instructions it a RV32A instruction.
So I think you should modify the condition for double words not words

JeanRochCoulon:
Thank you @AyoubJalali
Can you check it fixes the bug ?

AyoubJalali (approved):

AyoubJalali:
> Can you check it fixes the bug ?

LGTM

PR #1359, opened 2023-09-04
Remove Rename Block

fatimasaleem:


JeanRochCoulon:
Rename creates a 5% performance increase. Previously rename was done by hardware, by enabling the rename module. From the time we use GCC13, the rename is done by compilation.
Renaming does not work with all configurations. We anticipate collision between renaming and dual issue for instance. @zarubaf can you approve the renaming removal ?

zarubaf:
Thanks for the explanation, that makes sense. Especially if we can do it in GCC. Is this enabled automatically or is there a CPU profile that needs to be picked?

zarubaf (approved):

JeanRochCoulon:
The optimization is enabled by default in GCC (>13). It is not the case with GCC10. @cathales has experienced it on Coremark.

JeanRochCoulon:
Hello @fatimasaleem,   I tried to identify the gate count increase source (+300 gates) without success. It is a little bit surprising, we remove RTL while netlist is bigger ?!?!!
I merged anyway.

PR #1360, opened 2023-09-05
Fix STD Cache AXI `w_valid` Propagation

domenicw:
This PR fixes the propagation of the AXI `w_valid` signal in the `std_cache_subsystem` in the case that `aw_ready` is dependent on `w_valid`. The W-Channel fifo now also operates in fall through mode, removing a cycle of signal delay.


JeanRochCoulon:
Hello @domenicw 
The CI is failed, look at the failing report. I think a rebase is needed.

image

domenicw:
@JeanRochCoulon Thanks for looking into it. The PR is rebased now, let's see if that fixes it.

PR #1361, opened 2023-09-05
Add Direct D$ Access to `acc_dipsatcher`

domenicw:
This PR adds D$ access to the `acc_dispatcher`. For this, both caches (WT and STD) are adapted to handle a fourth D$ request port. This fourth port is used for issuing loads. Stores will go over the existing store request port. It is shared with the `store_buffer`, though the `store_buffer` always has priority access over the `acc_dispatcher`.

This fourth port is useful for accelerators, that want to access the D$ directly. Eventually it is also needed for the correct implementation of the CV-X-IF, which also has an interface into the memory. So this PR is a precursor to the future work needed for the correct CV-X-IF implementation.

### DIFF
--- a/core/acc_dispatcher.sv
+++ b/core/acc_dispatcher.sv
@@ -54,6 +54,8 @@ module acc_dispatcher import ariane_pkg::*; import riscv::*; #(
     input  logic                                  flush_ex_i,
     output logic                                  flush_pipeline_o,
     // Interface with cache subsystem
+    output dcache_req_i_t                   [1:0] acc_dcache_req_ports_o,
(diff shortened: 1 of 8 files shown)


JeanRochCoulon:
With this fourth data cache access, it is the possibility to manage virutalization with MMU ?!  Nice add-on

domenicw:
> With this fourth data cache access, it is the possibility to manage virutalization with MMU ?! Nice add-on

That is not yet possible. This fourth port goes directly into the cache. But the MMU lives in the `load_store_unit`. So some additional signals would need to be added to access it and to do the address translation. But this should be rather easy to do, I see no issue there.

Issue #1346, opened 2023-08-28
CVA6 CSR spike discrepancies list


spidugu444:
HI,

Please find the below list of CSRs which has discrepancies between CVA6 and spike 
<html xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:dt="uuid:C2F41010-65B3-11d1-A29F-00AA00C14882"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=OneNote.File>
<meta name=Generator content="Microsoft OneNote 15">
</head>

<body lang=en-US style='font-family:Calibri;font-size:11.0pt'>
<!--StartFragment-->

<p style='margin:0in;font-family:Calibri;font-size:11.0pt'>&nbsp;</p>

<p style='margin:0in;font-family:Calibri;font-size:11.0pt'>&nbsp;</p>

<div style='direction:ltr'>



Sl no | Name | Address Offset | Width | Access Type | Reset Value | Display Name | STATUS | COMMENT
-- | -- | -- | -- | -- | -- | -- | -- | --
1 | sstatus | 0x100 | 32 | RW | 0x00000000 | Supervisor Status | Spike Dependency | Need to disable FS field   on spike end
2 | mstatus | 0x300 | 32 | RW | 0x00000000 | Machine Status | Spike Dependency | Need to disable FS field and enable ube bit on   spike end
3 | misa | 0x301 | 32 | RW | 0x4014_1105 | Machine ISA | Spike Dependency | On MISA Bits 0, 2, and 12 behave like RW rather than WARL on spike end.
4 | pmpcfg[0..1] | 0x3A0 | 32 | RW | 0x00000000 | Physical Memory Protection Config 0 | spike dependency | Issue with default value reading as 0x01f
5 | pmpaddr[7] | 0x3B0 [+ i*0x1] | 32 | RW | 0x00000000 | Physical Memory Protection Address | spike dependency | Issue   with spike configuration and default values
6 | mvendorid | 0xF11 | 32 | RO | 0x00000602 | Machine Vendor ID | spike dependency | Need to configure reset value as 0x602 on spike end
7 | marchid | 0xF12 | 32 | RO | 0x00000003 | Machine Architecture ID | spike dependency | Need to configure reset value as 0x01 on    spike end
8 | icache | 0x7C0 | 32 | RW | 0x00000001 | Instruction Cache | spike dependency | Need   to configure Custom CSRs on spike end
[...]

PR #1368, opened 2023-09-07
Hardwire the reserved bits of the PMPCFG CSR to 0

Moschn:
Currently, the PMPCFG CSRs are implemented as a full 8-bit register. However, the bits 5 and 6 are reserved according to the privilege spec. This patch hardwires these two bits to 0. This should also fix the discrepancy between spike and CVA6 for the PMPCFG CSRs (#1346 ). It might also lead to lower area/less luts used.


zarubaf (approved):
Save all the bits! Thanks @Moschn

JeanRochCoulon:
@spidugu444 do you confirm this PR fix (partially) #1346 ?

spidugu444:
With this PR PMPCFG CSR bit [6:5] has been hardwired to zero, which fix the discrepancy between spike and CVA6.

PR #1380, opened 2023-09-11
CV32A6 CSR Access Verification PR

spidugu444:
Please find the PR for CVA6 32-bit CSR Access Verification.

**Highlights:**

1.  Yaml files that may be used to generate CVA6 32-bit CSR access tests have been added, which has been categorized into supervisor mode read-write yaml, machine mode read-write yaml and Machine read-only yaml.

2. Added test programs to verify access modes for CVA6 32-bit CSRs, which has been categorized as supervisor mode read-write, machine mode read-write, read-only, and counter CSRs.

3. The CVA6 counter CSRs behavior as well as the scenario when the counter CSR value reaches to maximum value have been verified.

4. Added CVA6 32-bit CSR access failing tests.

Thanks


JeanRochCoulon:
@AnouarZajni  Can you have a look to this PR ?  The objective is to know how we can insert/use these tests into the CI flow.

AnouarZajni:
Hi @JeanRochCoulon,
Those tests are self-checking and may be heavy to maintain because dependent to CVA6 configuration. As CVA6 verification is relying on spike model, will suggest removing self-checks.

MikeOpenHWGroup:
Hi @AnouarZajni, you are correct that these CSR access mode test-programs are self-checking.  That is a deliberate strategy for several reasons:
1. It is relatively straight-forward to develop self-checking test-programs for access mode verification.
2. Some CSRs are volatile and cannot be accurately predicted by a reference mode (Spike in our case).
3. Ultimately, these CSR access mode test-programs will be generated from a yaml description of the CSRs (the same one that we use to generate the documentation).

So we should not remove the self-checking aspects of the access-mode test-programs.

You make a good point that self-checking test-programs create a maintenance burden because they are dependent on CVA6 configuration.  We will face the same maintenance effort with the reference model.

JeanRochCoulon:
Hello @spidugu444 
Could you please remove the dvplan from this PR, as it has been merged by previous PR.
And create a dir hierarchy level to copy your work into tests/custom/CSR/blablabla.
Thanks

PR #1382, opened 2023-09-11
Fix debug module parameterization

zarubaf:
The debug module address was hard-coded which leads to a mismatch with a newer version of the debug module (since this commit https://github.com/pulp-platform/riscv-dbg/commit/55cda4b8043f4bdd609b188860ea115cc14499bb).

To avoid that in the future and to avoid dependencies on the debug module the debug module addresses can be passed as a parameter of `CVA6Cfg` now. I've also moved the `ariane_dm_pkg` constants into `ariane_pkg` to get rid of this module.


zarubaf:
@JeanRochCoulon how can I re-generate all configurations with the new config parameters?

JeanRochCoulon:
Hi Florian,  the configurations are setup in core/include/cvXXXXXXX_config_pkg.sv. You can add the new parameters to the cva6_cfg struct which is defined inside.

zarubaf:
> Hi Florian, the configurations are setup in core/include/cvXXXXXXX_config_pkg.sv. You can add the new parameters to the cva6_cfg struct which is defined inside.

Is that a manual process? I thought I could use the package Generation script 🙈

JeanRochCoulon:
Yes, manual :-(

zarubaf:
@JeanRochCoulon should be ready now for review. Thanks

JeanRochCoulon:
Thanks @zarubaf
Parametrization is not finished but we go ahead !!

PR #1383, opened 2023-09-11
dont use submodules to merge reports

cathales:
The "merge reports" job does not use submodules.
Not cloning/fetching submodules saves bandwidth.


zarubaf (approved):
Cool, totally agree!

Issue #1365, opened 2023-09-07
[TASK] Add Zicond Extension support in CVA6


fatimasaleem:
### Description 

This task aims to modify CVA6 RTL to add the Zicond("Conditional" operations) extension support.

The Zicond extension provides a simple solution that provides most of the benefits and flexibility one would desire to support conditional arithmetic and conditional-select/move operations while remaining true to the RISC-V design philosophy. The instructions follow the format for R-type instructions with three operands (i.e., two source operands and one destination operand). Using these instructions, branch-less sequences can be implemented (typically in two-instruction sequences) without the need for instruction fusion, special provisions during the decoding of architectural instructions, or other micro-architectural provisions.

### Changes

There will be changes in two major blocks of CVA6. 
1. Add opcodes of Zicond instructions in the decoder
2. The execute unit will going to be ALU for Zicond instructions as well

### Current Status
RTL does not support Zicond 

### Risks
None

### Prerequisites
None

### KPI
None

### Description of Done
 PR of design implementation verified after running architectural tests

Issue #1271, opened 2023-06-20
[TASK] Zicond in the requirement specification


jquevremont:
### Is there an existing CVA6 task for this?

- [X] I have searched the existing task issues

### Task Description

When Zicond is ratified by RISC-V International and its RTL code is merged, update the requirement specification.

### Description of Done

PR merged

PR #1405, opened 2023-09-13
Zicond

fatimasaleem:
This PR completes the [Add Zicond Extension support in CVA6](https://github.com/openhwgroup/cva6/issues/1365)
Which is verified using the Architectural tests that are developed by the RISC-V compliance team. Will add those tests in the arch-tests yaml files when we have the upstream support for Zicond extension in toolchain.

### DIFF
--- a/core/alu.sv
+++ b/core/alu.sv
@@ -302,6 +302,13 @@ module alu import ariane_pkg::*; #(
                 default: ; // default case to suppress unique warning
             endcase
         end
+
+        if (ariane_pkg::RCONDEXT) begin
+           unique case (fu_data_i.operation)
+           CZERO_EQZ : result_o = (fu_data_i.operand_b) ? (fu_data_i.operand_a) : (0) ; // move zero to rd if rs2 is equal to zero else rs1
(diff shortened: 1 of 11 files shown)


JeanRochCoulon:
@jquevremont The requirement and User Manual docs should be updated to add the support of Zicond extension, as this option is maybe needed for embedded configuration (to be confimed). This add-on is decribed by task #1271.
BTW Zc extension should also be added. @fatimasaleem can you update #1271 to add Zc extension and the need to update User Manual too ?

fatimasaleem:
@JeanRochCoulon I was thinking of updating the requirement and User Manual docs in a followup PR for Zicond, once we have merged this PR. 
Also, sure I will create a separate task for adding Zc in requirement and user manual docs as well.

jquevremont:
Zicond is not yet ratified. I can't progress on task #1271 until it is ratified.

JeanRochCoulon (commented):
Fatima, the CI is not pass. And the error is difficult to debug: FPGA boot is down. When the recommended modifications will be done, we will try to debug FPGA.

JeanRochCoulon:
Fine !    The LINT resolution fixes the FPGA build !    That's a very good news.
Thanks for the modification. We wait for the better GCC to use these instructions.

PR #1449, opened 2023-09-19
dependabot: run at 19:30 CET

cathales:
The goal is that dependabot opens a PR just before the Dashboard runs.
This way the results will be available on Tuesday mornings, CET.


zarubaf (approved):
👍

Issue #1456, opened 2023-09-19
[BUG] Misaligned address Exception Bypass in case of no MMU


fatimasaleem:
### Is there an existing CVA6 bug for this?

- [X] I have searched the existing bug issues

### Bug Description

MMU Bypasses the Misaligned address exception information if there is not any MMU exception, from the `misaligned_exception` struct to the `mmu_exception` struct. 

But when the  `MMU_PRESENT` is zero we should be able to send the misaligned exception information directly to either load unit or store unit. 

But currently, in RTL the mmu_exception struct is tied to [zero](https://github.com/openhwgroup/cva6/blob/b952b0d7c3e30dbfa7dc4d89a4579307d73c7f53/core/load_store_unit.sv#L231) when `MMU_PRESENT` is zero. It should pass through the misaligned exception information instead.

fatimasaleem:
@JeanRochCoulon This will affect the embedded configuration.

PR #1457, opened 2023-09-19
Bypass misaligned address exception info in case of no MMU

fatimasaleem:
Resolves https://github.com/openhwgroup/cva6/issues/1456


JeanRochCoulon:
Very good. I approve. Just wait for the CI.

Issue #1466, opened 2023-09-22
[BUG] Instruction fetch misaligned exception raised when a conditional branch that is not taken


yaohsiaopid:
### Is there an existing CVA6 bug for this?

- [X] I have searched the existing bug issues

### Bug Description

Hi,

According to the RISC-V 2.2 specification p.11: 
> An instruction address misaligned exception is generated on a taken branch or unconditional jump if the target address is not four-byte aligned. **No instruction fetch misaligned exception is generated for a conditional branch that is not taken**

But in `core/branch_unit.sv`, the exception is raised in branch unit as long as target address is misaligned *without taking the conditional branch result into account* :
`if (branch_valid_i && target_address[0] != 1'b0) branch_exception_o.valid = 1'b1;` from [line 94](https://github.com/openhwgroup/cva6/blob/72143ef0a9b50f2aaffa708a55263a5971bf0354/core/branch_unit.sv#L94) 

The fix should augment the `if` condition to include condition check that validate it is either an unconditional jump  or a **taken** conditional branch.

Thanks!

JeanRochCoulon:
Since #1467 has been merged on master, do you confirm the bug is solved. In that case, can you close ths issue ?

PR #1467, opened 2023-09-22
Branch unit: Fix incorrect raising of  misaligned exception

yaohsiaopid:
This patch fixes the issue [cva6#1466](https://github.com/openhwgroup/cva6/issues/1466): the branch unit incorrectly raises exception when target address is misaligned in regardless of the type of branch (can be unconditional or conditional).
According to RISC-V specificaiton p.11:
> An instruction address misaligned exception is generated on a taken
branch or unconditional jump if the target address is not four-byte aligned.

This patch fixes this by taking into account the type of branch and do a check on result (taken or not) when the branch is a conditional one.

### DIFF
--- a/core/branch_unit.sv
+++ b/core/branch_unit.sv
@@ -86,11 +86,21 @@ module branch_unit #(
     end
     // use ALU exception signal for storing instruction fetch exceptions if
     // the target address is not aligned to a 2 byte boundary
+    //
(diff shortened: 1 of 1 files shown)


JeanRochCoulon:
Thanks Yao for this contribution. Waiting for the CI result.

PR #1487, opened 2023-09-28
[HOT FIX] fix synthesis job

JeanRochCoulon:
A bad operation lead to update directly master while the idea was to create a PR. The synthesis is broken. Try to fix it with this hot fix


JeanRochCoulon:
@zarubaf  I quite confused because I merge directly on master and created a not working cva6. This hot fix replaces the default config by embedded config. 
I will merge this PR to fix the database, but do not hesitate to comment the PR if needed

Issue #1489, opened 2023-09-28
make verilate failed


hhhsiang:
### Is there an existing CVA6 bug for this?

- [X] I have searched the existing bug issues

### Bug Description

I've come across a problem much like the one described in https://github.com/openhwgroup/cva6/issues/1410. When I run `make verilate`, it displays the exact same error.
![image](https://github.com/openhwgroup/cva6/assets/125340710/a9092072-f5d6-4b26-bfe1-6d8dd616a6bb)

I've been struggling with this issue for a while. Any help in finding a solution would be very welcome. Thank you in advance!

JeanRochCoulon:
Hello @hhhsiang 
Which Verilator version do you use ?
Regards

hhhsiang:
Hi,@JeanRochCoulon
I use verilator version 4.110

JeanRochCoulon:
Please refer to the README.md CVA6 repository file, the supported Verilator is not more 4.110. If it works with the recommanded version, do not hesitate to close the issue.

hhhsiang:
@JeanRochCoulon  I'm sorry, but I couldn't find the information about the required Verilator version in the [README.md](https://github.com/openhwgroup/cva6) file.

JeanRochCoulon:
You are right. The supported Verilator version is 5.008. README.md need to be updated.

JeanRochCoulon:
I have created #1492 to add information about Verilator version

hhhsiang:
@JeanRochCoulon, thanks for you help!
However, I'm not sure how to update the version of verilator. Should I use `./ci/install-verilator.sh` , and change the version to 5.008? Like this:
![image](https://github.com/openhwgroup/cva6/assets/125340710/f8179cd1-d085-495a-bcf3-dc97b13f145f)

JeanRochCoulon:
You have to execute the verif/regress/install-verilator.sh file. The Verilator5.008 will be installed.

hhhsiang:
@JeanRochCoulon, I have tried it, but it show this:
![image](https://github.com/openhwgroup/cva6/assets/125340710/da6d621b-9655-4cbb-917c-0276d55f830b)

JeanRochCoulon:
The README.md recommands to execute verif/regress/install-cva6.sh.
I suggest to execute it.

hhhsiang:
@JeanRochCoulon, it appears to have been installed successfully, but when I run the command `verilator -version`, it still displays version 4.110, as follows:
[...]

PR #1505, opened 2023-10-06
Provide only one way to run simulation

JeanRochCoulon:
README.md how to run a simulation by executing Makefile commands, for instance 'make verilate'. This way does not work because additional variables need to be setup before executing these commands, for instance TARGET=cv64a6_imafdc_sv39. Many GitHub issues request support as #1489. This PR removes the getting start related to these commands to keep the way which is support that is to say using cva6.py command.

@MikeOpenHWGroup @zarubaf can you approve ?

### DIFF
--- a/README.md
+++ b/README.md
@@ -415,7 +415,7 @@ To activate the different cache system, compile your code with the macro `DCACHE
 
 ## Planned Improvements
 
-Go to the cva6 Kanban Board which also loosely tracks planned improvements.
+Go to the CVA6 Kanban Board which also loosely tracks planned improvements.
 
 
 ## Going Beyond
--- a/README.md
+++ b/README.md
@@ -415,89 +415,10 @@ To activate the different cache system, compile your code with the macro `DCACHE
 
 ## Planned Improvements
 
-Check-out the issue tab which also loosely tracks planned improvements.
+Go to the cva6 Kanban Board which also loosely tracks planned improvements.
 
 
 ## Going Beyond
-
-The core has been developed with a full licensed version of QuestaSim. If you happen to have this simulator available yourself here is how you could run the core with it.
-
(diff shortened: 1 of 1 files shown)


hhhsiang:
@JeanRochCoulon, is there any documentation or README available that explains how to use cva6.py for running simulations? Specifically, if I want to test a single instruction like rv64ui-v-add, what command should I use, in which directory should I run these commands, and what initial settings or environment paths need to be configured? Your assistance is greatly appreciated!

JeanRochCoulon:
Recommandations are in README.md of CVA6 repository. Have you read it before ?
If you would like to run rv64ui-v-add, modify the slome-tests.sh script in that sense. cva6.py --help produces the user manual.

zarubaf (approved):
👍

hhhsiang:
> Recommandations are in README.md of CVA6 repository. Have you read it before ? If you would like to run rv64ui-v-add, modify the smoke-tests.sh script in that sense. cva6.py --help produces the user manual.

@JeanRochCoulon, I had previously read the README.md, but it wasn't as explicit in explaining how to do it as shown in the screenshot below:
image

 However, I have a better understanding now. If I want to test any specific instruction, I can simply add it to the smoke-tests.sh script and then execute it, correct? Like the following example:
 
image

Additionally, I'm a bit unsure about your statement:

> cva6.py --help produces the user manual.

I tried running both `cva6.py --help` and `make cva6.py --help`, but nothing happened:

image

JeanRochCoulon:
Indeed cva6.py --help produces the list of options (maybe the words "User manual" was to much in relaitn to this list).

hhhsiang:
@JeanRochCoulon, I apologize for bothering you again.

If I wish to obtain a .vcd file for debugging purposes, how can I achieve that? The smoke-tests.sh script does not seem to generate it, and I couldn't find any information about this in the README.md.

JeanRochCoulon:
@zchamski   Can you help on this point ?

PR #1517, opened 2023-10-09
ci: reorganize jobs

cathales:
This PR reduces the number of jobs run in a CI pipeline, focusing on the `cv32a6_embedded` configuration.

It reorganizes jobs in categories and in stages to not run the whole pipeline when a smoke test fails.


zarubaf (approved):
Since this lives within Thales space and you are happy then I am happy 👍

PR #1521, opened 2023-10-09
Remove partially RVFI for synthesis

JeanRochCoulon:
The lsu address generation is costy in term of gate count. That's why it is better remove the lsa address generation for synthesis


JeanRochCoulon:
Removing RVFI saves 7kgates. But PR need to be revisited, close at this time.

PR #1522, opened 2023-10-09
reduce store buffer size to reduce gate count

JeanRochCoulon:
Since #1349 the store buffer can be reduced (I cross the fingers)


JeanRochCoulon:
The gate count improvement is not significant, close at this time

Issue #1542, opened 2023-10-18
[BUG] Performance counters are cleared if configured event is absent for one or more cycles.


creinwar:
### Is there an existing CVA6 bug for this?

- [X] I have searched the existing bug issues

### Bug Description

In ```core/perf_counters.sv:115``` the ```generic_counter[i]``` register is reset to 0 in every cycle, where the configured event is not present.

PR #1546, opened 2023-10-18
perf_counters: Fix counter resetting to zero

bluewww:
`events[i]` can be `'0` when there is nothing to count but that shouldn't reset the current count.

Fixes https://github.com/openhwgroup/cva6/issues/1542


JeanRochCoulon:
@TulikaSi Can you approve the PR ?

PR #1550, opened 2023-10-19
ci(dev): run synthesis tests only when RTL has changed

cathales:
This will make "dev" pipelines (used for PRs) shorter when none of `core/` or `corev_apu/` are changed, as only frontend smoke tests will be run.


cathales:
@MikeOpenHWGroup we discussed about this, here it is! 😃

PR #1552, opened 2023-10-19
[bot] Add stale-bot

zarubaf:


cathales:
Looks like not all submodules can be updated at the moment 😬 

Maybe it would be interesting in the future to keep more submodules updated thanks to Dependabot?

zarubaf:
Oops might have hit the rate limit. Just saw that the the branch needs to be in the default branch. So will merge, the workflow is triggered manually and in debug mode at the moment. So nothing should get broken.

Issue #1404, opened 2023-09-13
[TASK] Compare Spike and RTL in tandem


zchamski:
**Background**

Lockstep ("tandem") co-simulation allows early detection of divergences between behaviors of different simulation models on the same test case.  Typically, an Instruction Set Simulator (ISS) and an RTL simulator are fed with the same binary program. The simulation can be stopped as soon as a divergence is detected, or can be pursued for a controlled amount of steps (instruction commits + exceptions) to observe the consequences of the divergence.

For CVA6, the current ISS simulator is Spike and the primary RTL simulator is VCS, set up in either "test harness" or full UVM mode.

**What (objective description)**

Build a software system in which Spike and VCS simulations are executed concurrently and the observable state of both models is compared at every instruction commit and at every exception triggered.

**How To**

* extend CVA6 RVFI implemented in SystemVerilog with the information on CSR state (or state changes), cf. #1402.
* extend/adapt Spike RVFI information to match CVA6-side CSR information, cf. #1384.
* connect the Spike simulator as a slave of the RTL simulation model
* feed instruction commits and exceptions raised on the RTL side to the Spike model
* compare state changes reported by Spike with those reported by the RTL.

**Current  Status**

In progress.  Working prototype available for `vcs-testharness` and `vcs-uvm`, but it only compares PC values, instruction content, and GPR and memory writes.

**Risks**

TBD

**Prerequisites**

Dependency on completion of #1402 and #1384.  For the Embedded configuration additional dependency on #1403.

**KPI**

TBD

**Description of DONE**

Identical set of PASS and FAIL results on CI tests between full trace comparison and lockstep simulation.

MarioOpenHWGroup:
A working version of the Tandem has been pushed to my repo that implements the following:
- Tandem comparing exceptions, interrupts, destination GPR, instruction and pc. 
- Common package that implements the functionality for the testharness and UVM implementation

MarioOpenHWGroup:
Tandem rebased to last cva6/master commit. Some discussion needed regarding integration Sim*.cc  `corev_apu/tb/common/Sim`  into riscv-isa-sim

MarioOpenHWGroup:
Fixing Github CI in order to merge.
Also fixing the issues commented in both PRs.

MarioOpenHWGroup:
CI has been fixed and it shows green on Thales CI and on Github

MarioOpenHWGroup:
PR has been merged already

[...]

Issue #1403, opened 2023-09-13
[TASK] Modify Spike to match embedded configuration


zchamski:

MarioOpenHWGroup:
Further exploration is needed to catch all the discrepancies between spike and cva6

zchamski:
We have three sources of discrepancies:
* differences in interpretation of implementation-dependent "may"/"can" cases in the RISC-V specs
* bugs in CVA6
* (unlikely but possible) bugs in Spike.

We need to capture formally the "may" and "can" cases in RISC-V specs.  From there we need a systematic review of how these are implemented on both sides.

MarioOpenHWGroup:
tval value is not matching between core and spike, this needs to be discussed or fixed in spike side

MarioOpenHWGroup:
Tval has been reverted to the original state and the instr field of the scoreboard_entry_t has been removed

MarioOpenHWGroup:
Issues are appearing regarding specific cva6 config

MarioOpenHWGroup:
It is mandatory in this task that we know what we have the specification to rely when changing the Reference Model. Many issues above exist because unspecified parts or inconsistencies between the spec and the current environment implementation.

MarioOpenHWGroup:
I am making some notes about the can/may parts of the spec to implement (some of them) in Spike.

MarioOpenHWGroup:
No updates done in this task.

zchamski:
The behavior of bit 0 in the `PMPADDRn` CSRs depends on the address matching mode in effect (encoded in the associated `PMPCFG` register) **and** the PMP granularity set for each core.

The corresponding logic appears to be present in Spike and the termination of the PMP granularity is affected by the Device Tree specification used by Spike:

- A default value of 4 is set in the constructor of the `processor_t` class.
- If a device tree seen by Spike contain an explicit PMP granularity entry for the given core, that value will be used (if valid).

By default, Spike reads the built-in device tree which specifies a hard-coded PMP granularity of 4 (a compile-time constant).

On CVA6 the PMP granularity is hardwired to 8, and therefore any device tree-suppllied values will be silently ignored.  The best solution seems thus to be a new `Param` entry that overrides the default PMP granularity set by Spike.

zchamski:
> Agree, I will add the parameter

[...]

Issue #898, opened 2022-06-03
[Bug Report] Incorrect *tval for ecall/ebreak


Phantom1003:
Our co-simulation framework found that the `*tval` of `ecall/ebreak` is incorrect.
In cva6, after `ecall/ebreak`, *tval will set to the machine code of the `ecall/ebreak` instruction.

In the following test case, after calling `ebreak` in s-mode, the value of `mtval` register is written to `0x100073`, which is the machine code of `ebreak` instruction.

```
[spike] core   0: 0x0000000080000174 (0x00100193) li      gp, 1
[cva6]       664ns      649 S 0000000080000174 0 00100193 li             gp, 1
[cva6]  Exception @     66500, PC: 0000000080000178, Cause: Breakpoint, tval: 0000000000100073
[spike] core   0: 0x0000000080000178 (0x00100073) ebreak
[spike] core   0: exception trap_breakpoint, epc 0x0000000080000178
[spike] core   0:           tval 0x0000000080000178
... /* in handler */
[spike] core   0: 0x0000000080000190 (0x343022f3) csrr    t0, mtval
[error] WDATA SIM 0000000080000178, DUT 0000000000100073
[error] check board clear 5 error
[CJ]  integer register Judge Failed
```

riscv-priviledged P41 ：If `mtval` is written with a nonzero value when a breakpoint, address-misaligned, access-fault, or page-fault exception occurs on an instruction fetch, load, or store, then mtval will contain the faulting virtual address.
According to specifications, `mtval` should be the faulting address (or zero).

Issue [448](https://github.com/openhwgroup/cva6/issues/448) tests the value in `stval` of `ecall` from user mode, our verification framework further discovered that `ebreak` also has the same bug, and both of them could be triggered under **any privilege modes**.

ebreak testcase: [cva6-1.zip](https://github.com/openhwgroup/cva6/files/8832584/cva6-1.zip)
ecall testcase: [cva6-2.zip](https://github.com/openhwgroup/cva6/files/8832585/cva6-2.zip)
> @LuminaDCIX helps reproduce the problem

zarubaf:
Indeed, the instruction bits are the default case for every instruction. Confirming that we are not complying.

`ebreak`/`ecall` should be able to be triggered from any privilege level, no? How would a syscall/debug call work otherwise from user space?

Phantom1003:
Thanks, the point we wanted to confirm was the mismatched *tval.
[...]

PR #1561, opened 2023-10-20
Spike Tandem Implementation using VCS simulator

MarioOpenHWGroup:
#1403 #1404 Implementation


.gitlab-ci.yml
```diff
@@ -139,6 +139,24 @@ smoke:
     - source verif/regress/smoke-tests.sh
     - !reference [.simu_after_script]
 
+smoke-tandem:
```
JeanRochCoulon, line 142: This job is similar to smoke-tests. Can be removed.

Makefile
```diff
@@ -214,7 +214,8 @@ fpga_src :=  $(wildcard corev_apu/fpga/src/*.sv) $(wildcard corev_apu/fpga/src/b
 fpga_src := $(addprefix $(root-dir), $(fpga_src))
 
 # look for testbenches
-tbs := core/include/$(target)_config_pkg.sv corev_apu/tb/ariane_tb.sv corev_apu/tb/ariane_testharness.sv
+tbs :=  core/include/$(target)_config_pkg.sv corev_apu/tb/ariane_tb.sv corev_apu/tb/ariane_testharness.sv
```
JeanRochCoulon, line 217: blank to be removed

core/cva6.sv
```diff
@@ -1313,21 +1313,15 @@ module cva6
       end
       for (int i = 0; i < CVA6ExtendCfg.NrCommitPorts; i++) begin
         if (commit_ack[i] && !commit_instr_id_commit[i].ex.valid) begin
-          $fwrite(f, "%d 0x%0h %s (0x%h) DASM(%h)\n", cycles, commit_instr_id_commit[i].pc, mode,
```
JeanRochCoulon, line 1316: As RTL has been reformated by Verible because lines were too long, need to unchange the format.

core/cva6.sv
```diff
@@ -1363,12 +1357,12 @@ module cva6
...
-        rvfi_o[i].insn     = ex_commit.valid ? ex_commit.tval[31:0] : commit_instr_id_commit[i].ex.tval[31:0];
+        rvfi_o[i].insn     = commit_instr_id_commit[i].instr[31:0];
         // when trap, the instruction is not executed
-        rvfi_o[i].trap = mem_exception;
-        rvfi_o[i].cause = ex_commit.cause;
-        rvfi_o[i].mode = debug_mode ? 2'b10 : priv_lvl;
-        rvfi_o[i].ixl = riscv::XLEN == 64 ? 2 : 1;
+        rvfi_o[i].trap     = exception;
+        rvfi_o[i].cause    = commit_instr_id_commit[i].ex.cause;
+        rvfi_o[i].mode     = debug_mode ? 2'b10 : priv_lvl;
```
JeanRochCoulon, line 1364: remove blanks

core/decoder.sv
```diff
@@ -1301,7 +1301,8 @@ module decoder
     if (~ex_i.valid) begin
       // if we didn't already get an exception save the instruction here as we may need it
       // in the commit stage if we got a access exception to one of the CSR registers
-      instruction_o.ex.tval  = (is_compressed_i) ? {{riscv::XLEN-16{1'b0}}, compressed_instr_i} : {{riscv::XLEN-32{1'b0}}, instruction_i};
```
JeanRochCoulon, line 1304: tval should be conditioned by ZERO_TVAL or SPIKE_TANDEM.
But the best solution to not condition the RTL would be to discard tval compare at Spike level.

core/include/ariane_pkg.sv
```diff
@@ -107,22 +107,17 @@ package ariane_pkg;
...
+
 `ifdef SPIKE_TANDEM
-  // enable performance cycle counter, if set to zero mcycle will be incremented
-  // with instret (non RISC-V conformal)
-  localparam bit ENABLE_CYCLE_COUNT = 1'b0;
-  // mark WIF as nop
-  localparam bit ENABLE_WFI = 1'b0;
-  // Spike zeros tval on all exception except memory faults
+  // Spike still places 0 in TVAL for ENV_CALL_* exceptions.
+  // This may eventually go away when Spike starts to handle TVAL for *all* exceptions.
```
JeanRochCoulon, line 118: Removed if comapre is done at Spike level

corev_apu/tb/common/SimDTM.sv
```diff
@@ -1,19 +1,19 @@
 // See LICENSE.SiFive for license details.
 //VCS coverage exclude_file
 
-import "DPI-C" function int debug_tick
```
JeanRochCoulon, line 4: to be clean-up

corev_apu/tb/common/SimDTM.sv
```diff
@@ -65,17 +65,17 @@ module SimDTM(
     end
     else
     begin
```
JeanRochCoulon, line 67: to be cleaned-up

verif/env/uvme/uvme_cva6_env.sv
```diff
@@ -50,7 +50,7 @@ class uvme_cva6_env_c extends uvm_env;
    virtual uvmt_axi_switch_intf  axi_switch_vif;
 
    //CSR register model
```
JeanRochCoulon, line 52: remove blanks

verif/sim/cva6.yaml
```diff
@@ -18,7 +18,7 @@
   # Always keep this value in sync with the settings of RTL simulators (cf.
   # <issrun_opts> values below).
   cmd: >
```
JeanRochCoulon, line 20: to be discussed

JeanRochCoulon:
Mario, the Spike is not recompiled in GitHub Action, that maybe the reason why the CI does not pass. Maybe @zarubaf can help to fix the issue ?

core/decoder.sv
MarioOpenHWGroup: https://github.com/openhwgroup/cva6/issues/898 to double check

JeanRochCoulon:
@zarubaf   The Tandem PR is (quite) huge, can you also have a look to it ?

JeanRochCoulon:
@ASintzoff   Can you have a look to the Tandem PR. Mario make a tval modification which need to be validated. We are waiting for your opinion on that. Moreover, as your have a solid reviewer reputation, we wonder whether you would could help in reviewing the PR

core/decoder.sv
MarioOpenHWGroup: For me it is fine either way, but I would like to specify which cases the tval has a specific value, because for example when a ECALL is done the tval is also set with the opcode of the ecall

corev_apu/tb/ariane_tb.cpp
```diff
@@ -335,15 +337,23 @@ int main(int argc, char **argv) {
...
 #endif
-  // memif.read(0x84000000, mem_size, (void *)top->ariane_testharness__DOT__i_sram__DOT__gen_cut__BRA__0__KET____DOT__gen_mem__DOT__gen_mem_user__DOT__i_tc_sram_wrapper_user__DOT__i_tc_sram__DOT__sram);
+  long long addr;
+  long long len;
+
+  size_t mem_size = 0xFFFFFF;
+  while(get_section(&addr, &len))
+  {
+    read_section_void(addr, (void *) MEM , mem_size);
+    if (addr == 0x80000000)
```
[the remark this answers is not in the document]
MarioOpenHWGroup: I have done this code because previously it was only loading the memory this particular memory address. If an iteration is done to the sections it would cause a non-ending execution. as it will override always from this particular address. I will try to make it prettier and if it does not convince you I will revert the change

verif/sim/Makefile
```diff
@@ -156,44 +164,41 @@ export CVA6_UVMT_PATH         = $(CVA6_REPO_DIR)/verif/tb/uvmt
...
-export DV_UVMA_INTERRUPT_PATH = $(CVA6_REPO_DIR)/verif/core-v-verif/lib/uvm_agents/uvma_interrupt
-export DV_UVMA_DEBUG_PATH     = $(CVA6_REPO_DIR)/verif/core-v-verif/lib/uvm_agents/uvma_debug
-export DV_UVMA_OBI_PATH       = $(CVA6_REPO_DIR)/verif/core-v-verif/lib/uvm_agents/uvma_obi
-export DV_UVML_TRN_PATH       = $(CVA6_REPO_DIR)/verif/core-v-verif/lib/uvm_libs/uvml_trn
-export DV_UVML_MEM_PATH       = $(CVA6_REPO_DIR)/verif/core-v-verif/lib/uvm_libs/uvml_mem
-export DV_UVML_LOGS_PATH      = $(CVA6_REPO_DIR)/verif/core-v-verif/lib/uvm_libs/uvml_logs
-export DV_UVML_SB_PATH        = $(CVA6_REPO_DIR)/verif/core-v-verif/lib/uvm_libs/uvml_sb
-export CV_CORE_PKG            = $(CVA6_REPO_DIR)/verif/core-v-verif/core-v-cores/$(CV_CORE_LC)
+export DV_UVMT_PATH           = $(CORE_V_VERIF)/$(CV_CORE_LC)/verif/tb/uvmt
+export DV_UVME_PATH           = $(CORE_V_VERIF)/$(CV_CORE_LC)/verif/env/uvme
```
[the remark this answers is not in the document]
MarioOpenHWGroup: In my opinion, I think so

core/include/ariane_pkg.sv
```diff
@@ -689,6 +684,7 @@ package ariane_pkg;
 
   typedef struct packed {
     logic [riscv::VLEN-1:0] pc;  // PC of instruction
+    riscv::xlen_t             instr;         // Instruction value
```
zarubaf, line 687: Imho we really shouldn’t do that. That bloats the already bloated scoreboard even more. This is on the critical path.

core/decoder.sv
zarubaf: Agree with @JeanRochCoulon , depending on the cause we should do the comparison with spike or not. Let’s not modify the RTL to confirm to spike but the other way around (as long as it is legal :-))

zarubaf: Not that the Spec itself is quite generous on the `tval` field (can be 0), but in reality requirements are more stringent. For efficient emulation of illegal instructions the tval field can be read, RVA requires stval to be set to the faulting address. (https://github.com/riscv/riscv-profiles/blob/main/profiles.adoc#rva20-profiles see Sstvala)

zarubaf (changes requested):
Awesome, thanks a lot for pushing this feature!

JeanRochCoulon:
Great result. Yesterday the Thales CI was broken. We will (try to) fix it Today.

MarioOpenHWGroup:
> Great result. Yesterday the Thales CI was broken. We will (try to) fix it Today.

If I can do something to help let me know :+1:

JeanRochCoulon:
The Thales CI ran this morning, here is the result (there is a regression)

image

JeanRochCoulon:
@MarioOpenHWGroup   I tried to make TVAL optional but it is used to generate RVFI. Removing it suppress the instruction trace generation and does not allow to do compare between Spike and RTL. Action is ongoing to generate RVFI wihtout the help of TVAL.

JeanRochCoulon:
Huge work !  Thanks to you @MarioOpenHWGroup !

PR #1594, opened 2023-11-03
Implement only 1 dcache_ctrl instances when ACC and MMU_PRESENT are disabled

JeanRochCoulon:
During ACCELERATE_PORT integration, NumPorts has been upgraded from 3 to 4 in wt_dcache.sv. This has implemented 3 times dcache_ctrl.sv in place of 2 times. @zarubaf Do you confirm/approve ?

### DIFF
--- a/.gitlab-ci/expected_synth.yml
+++ b/.gitlab-ci/expected_synth.yml
@@ -1,2 +1,2 @@
 cv32a6_embedded:
-  gates: 120239
+  gates: 119508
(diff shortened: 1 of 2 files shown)


domenicw:
Yes, that was on purpose. The third instance is needed for the read access of the accelerator.
I think I see the problem though, the signals are not correctly tied off if the fourth port is unused. 
To solve this, I think the loop should stay the same. But the `if` condition should be changed to not create the additional instance it for the fourth port, if the accelerator is not used. Plus the `else` block would need to be changed to use the `k` index instead of just zero.

JeanRochCoulon:
Thanks @domenicw for the feedback, I followed your directives. I assume you approve the PR. As soon as it will be green I will merge.

PR #1595, opened 2023-11-03
[HOT FIX] Add else to SRET instruction decode

JeanRochCoulon:
Add the generation of illegal instruction in case of SRET decoding and supervisor mode disable.

@AEzzejjari can you confirm ?


AEzzejjari:
> Add the generation of illegal instruction in case of SRET decoding and supervisor mode disable.
> 
> @AEzzejjari can you confirm ?

Yes, I'm good with that

PR #1596, opened 2023-11-03
Do not support DRET when DebugEn = 0

JeanRochCoulon:
When debug mode is disable, DRET instruction is not supported.


JeanRochCoulon:
@ASintzoff can you approve ?

PR #1607, opened 2023-11-08
Modify coding style to improve CC

JeanRochCoulon:
Modify coding style to improve Code Coverage.


JeanRochCoulon:
@Gchauvon @cathales can you give me your feedback on this PR ?

cathales:
If I understand correctly there were:

- missing `else` so a potential latch **as there is no default assignment to `result_o` at the beginning of the process**
- several untested conditions as we do not test FPU

With this PR we:

- do not have a potential latch anymore
- have only one untested condition (better coverage?)
- introduce an unreachable statement in the nested `default`

cathales (approved):
>     * introduce an unreachable statement in the nested `default`

Ah sorry I didn’t notice the `default` case was reached so the explicit cases were not exhaustive. So you are not introducing an unreachable case.

PR #1620, opened 2023-11-13
dependabot: update verif/sim/dv

cathales:
As mentioned in #1591


JeanRochCoulon:
@AyoubJalali    For information, @cathales has setup a dependabot to update the riscv-dv version on each Monday. If the smoke_gen is green, I merge. If you do not agree, and you think other tests need to be executed, please pop-up !

PR #1639, opened 2023-11-16
Remove csr test from CI

JeanRochCoulon:
As this i test is not adpated to embedded configuration, remove it from CI.


JeanRochCoulon:
@ASintzoff   Can you merge ?

PR #1644, opened 2023-11-20
remove outdated manifest files

### LINKED ISSUE #1628: Outdated manifest files
Hello,

`ariane.core` and `scr_files.yml` mention a `src/scoreboard.sv` file. However the `src/` directory was removed by 3ddf797e95923fd11113c8e443046105dfbf8843 (#725, merged on Sep 24, 2021). So these files seem to be outdated since more than two years now.

I could not find issues about these files in the repository. Are they used? Should we update them? Should we remove them?

EDIT: Additional question: do we need both `Flist.ariane` and `core/Flist.cva6`?

cathales:
Closes #1628


JeanRochCoulon:
@zarubaf @Jbalkind @niwis Can you confirm these files are not useful ?

Jbalkind:
We actively use these files please stop trying to delete them

cathales:
@Jbalkind please be more specific about which files you use. Also please take into account issue #1628

Jbalkind:
We use flist.ariane and ariane.core. I think the other file is used by bender or ESP or something but I don't know. Please stop trying to delete them every few months. Of course these files will be "outdated" because the committers keep accepting PRs which add/remove/move core files without updating these manifest files. By their nature, these files will always be somewhat out of date and need to be updated via PR whenever the downstream project needs to update its submodule pointer to HEAD.

PR #1647, opened 2023-11-22
Code coverage: Add NonIdemPotence parameter

AEzzejjari:
Condition RTL with NonIdemPotence parameters to identify the dead code and remove the dead gates from netlist.

Based on if() directives, VCS is able to identify the unused code, this will allow to reach 100% code coverage.


JeanRochCoulon:
image

PR #1650, opened 2023-11-22
[csr/scounteren] fix privilege violation logic

acaldaya:
This PR fixes a privilege violation check when accessing the counter CSR from U-mode.
According to the spec:

> ...U-mode may only access a counter if the corresponding bits in scounteren and mcounteren are both set.

Before this fix, the violation only happens if both are set.

Tested on Linux 5.19 while using `rdcycle` from U-mode.


JeanRochCoulon:
Thank you @acaldaya for this PR. Can you approve the Eclipse agreement, that is a requirement to become contributor.

@ASintzoff   Can you approve ?

PR #1651, opened 2023-11-23
csr: Implement `mconfigptr`

zarubaf:
`mconfigptr` became mandatory with privileged spec 1.12. It seemed easy enough to implement the bare minimum (tieing it to `0`). If this is not on the roadmap @JeanRochCoulon let me know. Also happy to add it to the CSR description file if you can point me towards it.


zarubaf:
Nvm, I was on an old version. This is already implememted/

PR #1652, opened 2023-11-23
No hard coded gcc param

valentinThomazic:
Currently, some GCC compilation parameters are hard-coded in CVA6.py whereas the same parameters are given in the yaml files of most tests, meaning they are given twice.

This also allows removing them easily in cases where they are not required.


JeanRochCoulon:
Good !

PR #1653, opened 2023-11-23
csr: Implement `menvcfg`

zarubaf:
Only the `fiom` bit needs to be implemented. Since we anyway treat all fences and atomics as fully sequential the value is currently not needed anywhere.

## 3.1.18 Machine Environment Configuration Registers (`menvcfg` and `menvcfgh`)

If bit FIOM (Fence of I/O implies Memory) is set to one in menvcfg, FENCE instructions executed in modes less privileged than M are modified so the requirement to order accesses to device I/O implies also the requirement to order main memory accesses. Table 3.8 details the modified inter- pretation of FENCE instruction bits PI, PO, SI, and SO for modes less privileged than M when FIOM=1.
Similarly, for modes less privileged than M when FIOM=1, if an atomic instruction that accesses a region ordered as device I/O has its aq and/or rl bit set, then that instruction is ordered as though it accesses both device I/O and memory.
If S-mode is not supported, or if satp.MODE is read-only zero (always Bare), the implementation may make FIOM read-only zero.


core/csr_regfile.sv
```diff
@@ -843,6 +850,10 @@ module csr_regfile
           mask  = riscv::MIP_SSIP | riscv::MIP_STIP | riscv::MIP_SEIP;
           mip_d = (mip_q & ~mask) | (csr_wdata & mask);
         end
+        riscv::CSR_MENVCFG: if (CVA6Cfg.RVS) fiom_d <= csr_wdata[0];
```
ASintzoff, line 853: To avoid error during ASIC synthesis (Error:  ../..//core/csr_regfile.sv:549: Variable 'fiom_d' is the target of both blocking and nonblocking assignments in the same always block. (VER-134)), could you change the nonblocking assignement to a blocking one?

ASintzoff (changes requested):

core/csr_regfile.sv
JeanRochCoulon: Hello @zarubaf   André has identified an error. Could you fix to be able to merge the PR?

zarubaf: Should be fixed, sorry for the delay!

PR #1657, opened 2023-11-27
Configure load to be wait state zero in embedded configuration

JeanRochCoulon:
Technology permitting, load can be done in wait state zero, which increases performance.


JeanRochCoulon:
@Gchauvon  can you approve the PR ?

PR #1670, opened 2023-12-04
Remove fixed-time division (ISA-110)

jquevremont:
After further investigation, the feature is not needed for security application.

For safety applications, variable-time division is only one of several sources of unpredictability and it does not make real sense to address it.


JeanRochCoulon:
For security application, we do not need to be data independant for branch and div instructions, but cva6 shall be data independant for other instructions, arithmetical operations for instance. Zkt is an extension we could ask to be compliant, to be more deeply analyzed.

PR #1673, opened 2023-12-05
Update version of the HPDcache submodule

cfuguet:
- Update reference of the HPDcache submodule and related packages

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,7 +35,7 @@ jobs:
     strategy:
       matrix:
         testcase: [asm-tests, mul, amo, fp, benchmarks]
-        target: [cv64a6_imafdc_sv39, cv64a6_imafdc_sv39_wb]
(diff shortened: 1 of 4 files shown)


JeanRochCoulon:
Hello @cfuguet,     I have a talk I would like to share with you.   PR updates the sub-module, but as no configuration using HPDcache is simulated by smoke tests, we really do not know whether HPDcache is running.  An integration test on a configuration using HPDCache should be added to the CI ?

cfuguet:
Yes, absolutely @JeanRochCoulon. That was my next item in the TODO list. However, I wanted to see with you if I append the hpdcache config in the .ci script, thus running all tests for WT, WB and HPDcache, or do you prefer I create an additional section in the CI script where only a few tests are ran in the HPDcache ?

JeanRochCoulon:
Adding the HPDcache configuration in the GitHub action CI is good to me. I support the idea. Tell me if you prefer to add this feature to the current PR  or in a next one, in that case I would merge the current PR immediatly.

PR #1675, opened 2023-12-06
axi: Upgrade to v0.39.1

niwis:
Upgrade the AXI dependency from v0.31.0 to v0.39.1. Note that most changes are within the vendored AXI repository. See the [changelog of the AXI repo](https://github.com/pulp-platform/axi/blob/master/CHANGELOG.md) for details.


JeanRochCoulon:
The update introduces a regression in term of cycles on Coremark.

JeanRochCoulon:
The coremark performance is fixed. But now the Linux boot is failed, which is less fun to debug.

niwis:
That's too bad. I will give it a try

JeanRochCoulon:
The common trick to found out the error is to compare the FPGA builds with and without the modification.

JeanRochCoulon:
@niwis, I close the issue as fiw has not been found

Issue #1434, opened 2023-09-18
[TASK] Generate CSR DVplan


JeanRochCoulon:
### Is there an existing CVA6 task for this?

- [x] I have searched the existing task issues

### Task Description

Generate the Dvplan for CSRs in embedded config: field access, illegal CSRs

### Required Changes

Updating STEP1 Dvplan

### Current Status

STEP1 dvplan 

### Risks

none

### Prerequisites

CSR Spec for embedded

### KPI (KEY Performance Indicators)

none

### Description of Done

CSR Dvplan file for embedded config

JeanRochCoulon:
is this task started ?

AnouarZajni:
I started it today.

AnouarZajni:
@JeanRochCoulon, task done waiting for merge approval.

JeanRochCoulon:
Can you refer the PR number ?

AnouarZajni:
CSR DVPLAN embedded is PR !86
I've also updated hvp and RAL files to achieve dvplan goals in !87

AnouarZajni:
@ASintzoff, updated CSR Dvplan and pushed to the same merge request !86

AnouarZajni:
@ASintzoff , RAL files updated to reflect Dvplan with Legal/Illegal values and CSR counter features. Pushed to 87

AnouarZajni:
@ASintzoff , Updated merge !86 following feedbacks

AnouarZajni:
Merge Done, closing the the task

JeanRochCoulon:
@MikeOpenHWGroup  Can I ask you to review this task ? If you think this task is completed, please change the status from "Done" to "finished". Thanks

https://github.com/openhwgroup/cva6/tree/master/verif/docs/VerifPlans/csr_access

MikeOpenHWGroup:
I am re-opening this Issue as @JeanRochCoulon has requested a review of the CSR DV Plan.

MikeOpenHWGroup:
I have conducted a review of the CSR DVPlan, see #1697.

JeanRochCoulon:
@AnouarZajni @ASintzoff  Is this task completed ?   If yes, please refer the related PR.

AnouarZajni:
@JeanRochCoulon , yes task is done in MR !85 with PR #1606

PR #1697, opened 2023-12-11
CV32A60X-specific CSR DVplan

MikeOpenHWGroup:
This pull-request is intended to cover the CSR DVplan review as part of Task #1434.

I have made the following changes:
1. Moved the DVplan from `docs/VerifPlans/csr_embedded_access` to `docs/VerifPlans/CV32E6QX/csr_access`.  The directory `CV32E6QX` is a placeholder until the part number for the "Embedded Application" core derived from CVA6 is determined.
2. Updated the DVPlan using VPtool. (This gave me an excuse to work with VPtool).

We may or may not want to merge in this PR, so I have marked it as a "draft".


JeanRochCoulon:
Mike,   An option exists in VPTOOL to indicate which is the part number for which the verification point is created. That's why the VPTOOL database (coded in yaml format) is done for all the different configurations. 
I would suggest to discuss the point at the next Verification meeting.
@asintzoff

JeanRochCoulon:
@MikeOpenHWGroup Dont be angry if I close your PR. It seems to me deprecated Today: VPTOOL handles different configurations, Part numebr has changed and CSR files have been moved by another PR.

PR #1701, opened 2023-12-12
Update embedded config to improve trade-off performance gate count

JeanRochCoulon:
Based on Coremark study, three parameters are fine tuned: BHT, RAS, and load buffer.


cathales (approved):

cathales:
Just need to update the expected performance and area

PR #1704, opened 2023-12-13
Draft: User config

cathales:
> :warning: This PR is not ready for merge yet. It builds CVA6 with embedded configuration on Verilator and runs CoreMark but there are still formatting + history cleanup + rebase + fixes to do for VCS and other configurations.

This PR is a step towards CVA6 parametrization.

1. It distinguishes two types: "user configuration" VS "extended configuration", which is interesting because:
   - It is not needed anymore to insert dummy values into a structure if the configuration file.
   - It appears in the type signature that configuration has been extended or not: when the extended configuration is needed, the appropriate type is used to force that the transmitted parameter is already extended, avoiding the accidental use of dummy values.
2. It creates a function to gather the extension logic.
3. It moves user-config-dependent information from packages to this configuration structure as it is not possible for packages to use values from a configuration structure.

Not everything has been moved yet, but as moving XLEN (and its dependents PLEN, VLEN and many more types and values) implied a lot of changes, it is better to commit them as soon as it is working to avoid an even bigger PR (and the associated conflicts).


cathales:
Note after discussion with @AngelaGonzalezMarino:

I will change how I define types before spreading them. I have not embraced a specific style yet and I am not consistent. I’m sometimes defining a type with `parameter` even if it is not supposed to be taken as a generic, which is confusing. Three options to consider:

- At least use `localparam` instead of `parameter`
- Move the `localparam` into the `module` body (instead of in the `#()` generic parameters)
- Use a `typedef` instead of a `localparam type`, but then I’m not sure if transmitting it will be correctly supported.

I would opt for option 2: `localparam` in the `module` body.

Thanks a lot to @AngelaGonzalezMarino for the review!

PR #1706, opened 2023-12-13
Added CSR covegroups for read and write operations, hvp updated accordantly

AnouarZajni:
Added covergroup read and write for each CSR.


AyoubJalali:
@JeanRochCoulon Do not merge need fix

PR #1709, opened 2023-12-14
refactor(decoder): simplify interrupt indexing

cathales:
The indexing expression was really complex.  It is constant so is is
possible to directly use an "unsized" localparam.  The values are less
than 32 so it is equivalent to previous code.  This coding style is the
one used in the CSR regfile.

It will simplify future modifications, including completing
parametrization.


JeanRochCoulon:
Can you rebase your branch to get the up-to-date expected gate count modification ?

cathales:
6e41bc8b5295d1bcfe99144c61fc2c22d4872130 (`master` branch) failed on ASIC Synthesis cv32a6_embedded too

PR #1711, opened 2023-12-14
Clarify `pmpcfgX` on illegal write combination (fixes #1694)

### LINKED ISSUE #1694: [BUG] `pmpcfg` `RWX` not `WARL`
### Bug Description

According to the ISA Spec `pmpcfg` fields `RWX` behaves as `WARL` with `R=0` and `W=1` as reserved values. Looking at the code (I don't have a test case for it) that isn't the case:

https://github.com/openhwgroup/cva6/blob/a837e94eac9ee6542c297a59e0fee69080f6bd89/core/csr_regfile.sv#L971-L992

This is just to note that we need to fix this behavior. Preferably clearing the field `RWX` when this combination is written.

zarubaf:


jquevremont (approved):
LGTM

PR #1717, opened 2023-12-18
[Skip CI] Changing part number in user manual

jquevremont:
We accelerate this action to stabilize the embedded part number for upcoming commits.

Consensus (not unanimity) on **CV32A60MX** for the embedded configuration. Indeed, the previous CV32E60X (or CV32E65X) part number does not reflect the CVA6 family name and is a real issue for Thales.

CV32A60MX will change to CV32A65MX is the configuration is upgraded with dual-issue.


jquevremont:
Wrong place. Sorry.

Issue #1708, opened 2023-12-14
[TASK] Add Zb* in User Manual


ASintzoff:
### Is there an existing CVA6 task for this?

- [X] I have searched the existing task issues

### Task Description

Write User Manual chapter describing Zb* instructions (bitmanip extensions).

### Required Changes

Add chapter in User Manual.

- Split the specification in 4 files (one for each group: Zba, Zbb, Zbc, Zbs) to ease definition of configurations with some of the groups: `RISCV_Instructions_RVZba.rst`, `RISCV_Instructions_RVZbb.rst`, `RISCV_Instructions_RVZbc.rst`, `RISCV_Instructions_RVZbs.rst`
 

### Current Status

RTL available

### Description of Done

PR merged

PR #1721, opened 2023-12-19
Update Zb* docs

gullahmed1:
#1708


JeanRochCoulon:
Thanks @gullahmed1 for this contribution, it was needed to verify Zb extension. Good to have split it into 4 files.

PR #1731, opened 2023-12-21
use cathales version of verible-formatter-action

cathales:
Temporary workaround before [chipsalliance/verible-actions-common#5](https://github.com/chipsalliance/verible-actions-common/pull/5) is merged


cathales:
Note: the workflow version used is the one of `master`. So Verible breaks in this PR but once merged it should work for all open PRs without requiring a rebase.

cathales:
[chipsalliance/verible-actions-common#5](https://github.com/chipsalliance/verible-actions-common/pull/5) is merged

PR #1733, opened 2023-12-21
UM: Part number + reshuffled Zb* RV32/RV64 instructions

jquevremont:
In the user manual:
- Updated part numbers: CV32A60X (embedded configuration), CV32A60AX (application configuration)
- In the Zb* extensions, grouped RV32+RV64 instructions on one side and RV64-specific instructions on the other side, to ease the upcoming templating.

The upcoming templating will allow to generate documentation that are specific to a given configuration.


JeanRochCoulon:
Good PR, good to have slit 32bits and 64bits RVB instructions

PR #1743, opened 2024-01-02
Fix inhibit impact on MCYCLE csr

### LINKED ISSUE #1742: [BUG] Inhibition of cycle counter

### Bug Description

Hello!
I found that the current implementation of the performance counter inhibition for the cycle counter does not actually inhibit the mcycle CSR increment every cycle but simple writes to it the value of the instruction retired count.

There might be a reason for this behavior to be desirable but it does not seem to follow the RISC-V specification.

https://github.com/openhwgroup/cva6/blob/2708df998d55bc3c826a850ef567558b29fd633c/core/csr_regfile.sv#L571-L572

JeanRochCoulon:
this PR fixes #1742

### DIFF
--- a/core/csr_regfile.sv
+++ b/core/csr_regfile.sv
@@ -569,7 +569,7 @@ module csr_regfile
       instret_d = instret;
       // increment the cycle count
       if (!mcountinhibit_q[0]) cycle_d = cycle_q + 1'b1;
-      else cycle_d = instret;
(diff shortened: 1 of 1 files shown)


JeanRochCoulon:
image

JeanRochCoulon:
@cathales Can you approve ?

cathales (approved):

PR #1744, opened 2024-01-03
Fix HPDCache to make it functional when ways=2

JeanRochCoulon:
In embedded configuration, caches do not need to be so big.

### DIFF
--- a/.gitlab-ci/scripts/report_benchmark.py
+++ b/.gitlab-ci/scripts/report_benchmark.py
@@ -18,7 +18,7 @@ iterations = None
 # Will fail if the number of cycles is different from this one
 valid_cycles = {
     'dhrystone': 217900,
-    'coremark': 665193,
+    'coremark': 683117,
 }
 
 for arg in sys.argv[1:]:
--- a/.gitlab-ci/scripts/report_benchmark.py
+++ b/.gitlab-ci/scripts/report_benchmark.py
@@ -18,7 +18,7 @@ iterations = None
 # Will fail if the number of cycles is different from this one
 valid_cycles = {
     'dhrystone': 217900,
-    'coremark': 683117,
+    'coremark': 665193,
 }
 
 for arg in sys.argv[1:]:
--- a/core/include/cv32a6_embedded_config_pkg.sv
+++ b/core/include/cv32a6_embedded_config_pkg.sv
@@ -33,11 +33,11 @@ package cva6_config_pkg;
   localparam CVA6ConfigDataUserEn = 0;
   localparam CVA6ConfigDataUserWidth = CVA6ConfigXlen;
 
-  localparam CVA6ConfigIcacheByteSize = 16384;
-  localparam CVA6ConfigIcacheSetAssoc = 4;
+  localparam CVA6ConfigIcacheByteSize = 2048;
+  localparam CVA6ConfigIcacheSetAssoc = 2;
   localparam CVA6ConfigIcacheLineWidth = 128;
-  localparam CVA6ConfigDcacheByteSize = 32768;
-  localparam CVA6ConfigDcacheSetAssoc = 8;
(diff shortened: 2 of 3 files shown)


JeanRochCoulon:
@Gchauvon  Can you approve it ?

JeanRochCoulon:
Hello @AyoubJalali 
I modified the cache configuration. The modification should be without impact on CI, but the generated smoke tests failed. Do you have any idea ?

AyoubJalali:
> Hello @AyoubJalali I modified the cache configuration. The modification should be without impact on CI, but the generated smoke tests failed. Do you have any idea ?

I'll debug it !!

JeanRochCoulon:
@cfuguet   When HPDcache is configured with 2 ways of 2048 bytes, it fails. To be discussed at the workshop. See u

cfuguet:
> @cfuguet When HPDcache is configured with 2 ways of 2048 bytes, it fails. To be discussed at the workshop. See u

Oh ! I will check this. In theory, there is no reason why the HPDcache does not support this configuration.

JeanRochCoulon:
@cfuguet 

The HPDCache package does not compile when number of ways = 2

As HPDCACHE_DATA_RAM_Y_CUTS = HPDCACHE_WAYS / HPDCACHE_DATA_WAYS_PER_RAM_WORD,
when 8 ways, HPDCACHE_WAYS = 8 and HPDCACHE_DATA_WAYS_PER_RAM_WORD =4, it is ok
when 2 ways, HPDCACHE_WAYS = 2 and HPDCACHE_DATA_WAYS_PER_RAM_WORD =4, it is KO

image

cfuguet:
> @cfuguet
> 
> The HPDCache package does not compile when number of ways = 2
> 
> As HPDCACHE_DATA_RAM_Y_CUTS = HPDCACHE_WAYS / HPDCACHE_DATA_WAYS_PER_RAM_WORD, when 8 ways, HPDCACHE_WAYS = 8 and HPDCACHE_DATA_WAYS_PER_RAM_WORD =4, it is ok when 2 ways, HPDCACHE_WAYS = 2 and HPDCACHE_DATA_WAYS_PER_RAM_WORD =4, it is KO
> 
> image

Ok ! now I see why. The issue comes from the computation of the HPDCACHE_DATA_WAYS_PER_RAM_WORD parameter. It is currently computed as 128 (bits) / PARAM_WORD_WIDTH. In the embedded case, PARAM_WORD_WIDTH is 2, then the HPDcache expects to be able to fit 4 ways in a single word of the RAM. If the number of ways is less than 4, you have a bad computation of the HPDCACHE_DATA_RAM_Y_CUTS parameter which in this case will be 0... It should be at least 1.

I will change the way that the HPDCACHE_DATA_WAYS_PER_RAM_WORD parameter is computed in the core/include/cva6_hpdcache_default_config_pkg.sv. That should fix the issue. I will let you know when it is done. I can either push the commit in this PR or make another Pull-Request on the CVA6, what is your preference ?

Thank you @JeanRochCoulon,

César

JeanRochCoulon:
Committing in the current PR could test the HPDCache fix, let's commit in it.

cfuguet:
> Committing in the current PR could test the HPDCache fix, let's commit in it.

Ok ! let me do a quick local test and then I will push the modifications

cfuguet:
@JeanRochCoulon, FYI, this should fix issues on simulation, but it may fail on synthesis. Actually when the HPDCACHE_DATA_RAM_Y_CUTS = 1, there are some types that have null ranges ([$clog2(HPDCACHE_DATA_RAM_Y_CUTS)-1:0] => [-1:0]). This is not well supported by some tools. I will open an issue on the HPDcache and make a fix ASAP to support correctly this case.

JeanRochCoulon:
ok, thanks for the information. I will commit your fix and resize the I and D cache in another PR.

PR #1761, opened 2024-01-15
Break Timing Loop in Axi Adapter Arbiter of WB Cache

domenicw:
There exists a timing loop in the WB cache caused by the `axi_adapter_arbiter` module. This PR fixes it by breaking the connection between the `rsp_i.gnt` and `rsp_o[sel_q].gnt` signals. At the same time, this PR also removes an unnecessary cycle delay between two requests on the same port.


zarubaf (approved):
Thanks @domenicw!

fyi @colluca @niwis

hhhsiang:
Hi @domenicw, when I synthesize cva6 core with WB cache, the timing loop is still exist. Do you encounter this? https://github.com/openhwgroup/cva6/issues/2557

PR #1767, opened 2024-01-16
Add ld/st priv mode, sum & PMP iface signals to acc_dispatcher

michael-platzer:
Hi, this PR adds 4 additional ports for the current load/store privilege mode, the current content of the `sum` field of the `mstatus` CSR, and the contents of the PMP configuration and address CSRs to the `acc_dispatcher`.

These signals are necessary to perform memory protection checks on memory requests originating from an accelerator in a custom implementation of the `acc_dispatcher` module. Those signals allow to determine the privilege mode of loads and stores performed by an accelerator and, for example, to add an instance of the `pmp` module to enforce the current physical memory protection rules on those requests.

This PR simply adds the necessary ports and connects the corresponding signals to the `acc_dispatcher` module. The protection checks that they enable need to be added to a custom implementation of the `acc_dispatcher` which interfaces with the respective accelerator.

CC @domenicw @zarubaf


zarubaf (approved):
lgtm, not touching anything inside CVA6 so merging.

PR #1768, opened 2024-01-16
Add `priv` to Spike and fix smoke-tests

MarioOpenHWGroup:
Add priv level to ` cva6.py` to pass it to Spike
Fix smoke tests to previously source `setup-env.sh`


MarioOpenHWGroup:
@valentinThomazic please take a look if you can :+1:

PR #1769, opened 2024-01-16
Allow hard-wiring of PMP regions

michael-platzer:
This PR adds three new fields to the `cva6_cfg_t` configuration struct, which allow to specify reset values for the PMP configuration and address CSRs as well as optionally making individual PMP entries read-only. The purpose is to allow hard-wiring of certain regions' privileges, which is explicitly allowed by the RISC-V Privileged Architecture specification Machine-Level ISA, v1.12 (see Sect. 3.7).

Hard-wiring of PMP regions is useful whenever a SoC has memory regions with fixed properties and protection rules. Instead of relying on SW to setup the PMP regions correctly on every boot, with this PR regions may be initialized using the new parameters and optionally made read-only.


zarubaf:
lgtm, I think we miss the configuration in some testbench :-) Maybe @cathales can help? 🥹

michael-platzer:
I tried a few things as far as I could parse the error messages from the logs but now it appears that the CI issues are unrelated to my changes. Does anyone have some feedback/suggestions as to how I could resolve these issues?

cathales:
Hi! We encountered similar issues while reviewing #1760.

1. It seems you are missing [a modification](https://github.com/openhwgroup/cva6/pull/1760/commits/d3604ab6302171945f5a0656acbb05866fb77246#diff-86f1f47bb1bcb5af0e128339306dbcc972c80a4e7fdca22d6d5d819949cc9d96R180) in `cva6.sv` to insert your values inside the extended configuration. So the fields of the extended configuration are shifted, the behavior becomes inconsistent and it might be hard to debug (a little bit like C’s undefined behavior).
2. It didn’t reach FPGA build, but [mind ariane_xilinx.sv too](https://github.com/openhwgroup/cva6/pull/1760/commits/2f237ccda7c360e51c16e1a062a2ec6d7e6b1102)

michael-platzer:
Thanks @cathales for the suggestions! Added the missing modifications, so let's see whether it works now :slightly_smiling_face:

zarubaf (approved):
Awesome! Thanks

PR #1771, opened 2024-01-19
fix: add missing parameters in cva6_rvfi.sv

cathales:
Both #1769 and #1762 PRs were right but they collided after merge.

1. #1769 adds parameters
2. #1762 needs all parameters

#1762 was merged without being updated with the new parameters from #1769.


JeanRochCoulon:
Merci for this fix @cathales

PR #1782, opened 2024-01-24
Fix verilator bin path

ytliu74:
The verilator installed by given script `verif/regress/smoke-tests.sh` has a new `bin` path, which looks like `cva6/tools/verilator-v5.018/verilator/bin`. So the original `setup-env.sh` fails to put verilator into $PATH.


JeanRochCoulon:
Hello @ytliu74   Your PR was not fully functional. #1854 should fix it. If you agree, please to close this PR and the related Github issue. Thanks for your contribution.

ytliu74:
Okay, no problem.

PR #1784, opened 2024-01-25
Parameterize TVAL to reduce size in embedded

Gchauvon:
Add a parameter TvalEn to remove tval gates in embedded configuration.

Behaviour when software is trying to write to MTVAL or STVAL should be determined. 
In this PR, it is forbidden and access to MTVAL or STVAL when TvalEn is disabled will raise an illegal instruction exception.

### DIFF
--- a/.gitlab-ci/expected_synth.yml
+++ b/.gitlab-ci/expected_synth.yml
@@ -1,3 +1,3 @@
 cv32a6_embedded:
-  gates: 114319
+  gates: 110519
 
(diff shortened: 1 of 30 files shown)


JeanRochCoulon:
@MarioOpenHWGroup   With this PR, the TVAL will no more contain the instruction. Is this a problem for Spike tandem ?

MarioOpenHWGroup:
> @MarioOpenHWGroup With this PR, the TVAL will no more contain the instruction. Is this a problem for Spike tandem ?

It shouldn't be a problem as currently the `rvfi_o` inside cva6 does not rely in `tval`

PR #1801, opened 2024-02-02
cva6_config: Add ZiCondExtEn localparam in v configs

niwis:
Re-align the v configs by adding the `CVA6ConfigZiCondExtEn` localparam


JeanRochCoulon:
I do not understand why the config_pkg.sv files were not aligned. Thank you for the fix.

Issue #1785, opened 2024-01-25
[BUG] Allow one outstanding killed miss


JeanRochCoulon:
### Is there an existing CVA6 bug for this?

- [X] I have searched the existing bug issues

### Bug Description

@colluca has sublitted a PR to fix the following performance bug #1497. As the PR has not been merged. This github issue backup the problem.

The I$ can receive a speculative request from the instruction fetch unit (e.g. upon a branch). If this instruction is not in the cache we will incur a miss and a refill request will be issued to the memory. If, when the branch is resolved, the speculated instruction is killed, the refill request will be killed (i.e. the speculated instruction will never be loaded into the cache). Despite being killed, the I$ still waits on the request from memory to be back, stalling all other operation. This implies that even with a hot cache, we can still experience cache miss stalls on speculated instructions which will never enter the cache, and hence the cache will never truely be "hot" (in this sense, speculated instructions should be considered part of the working set).

This PR attempts to improve performance in this condition by handling up to one killed refill request asynchronously. A speculated miss will only induce a stall if another miss occurs before the killed refill returns from memory.

JeanRochCoulon:
Miek, I cannot assign @colluca to this issue, can you include him to the  CVA6 project ?

JeanRochCoulon:
Hello @yanicasa 
During your OBI investigation, you likely face this issue. Do you think it is solved in feature/interconnect branch ?

yanicasa:
In the current state of feature/interconnect, if the frontend kills a request when the icache's FSM is in MISS, the frontend is freed directly (OBI valid), so it will be able to make a new request, but this request will be blocked as the MISS is not finished because the icache doesn't support HIT in a MISS (due to icache implementation). 

what has changed is that **non-idempotent speculative accesses no longer go to the icache**, they wait to be resolved in the frontend, so they **no longer create a MISS**.

In my opinion, @colluca's PR is always interesting.

yanicasa:
I think this issue is only related to the way icache manage refill mechanism. 
As CV32A65X is PipelineOnly, label can be updated.

PR #1497, opened 2023-10-04
cva6_icache: Allow one outstanding killed miss

colluca:
This PR relates to the following behaviour.

The I$ can receive a speculative request from the instruction fetch unit (e.g. upon a branch). If this instruction is not in the cache we will incur a miss and a refill request will be issued to the memory. If, when the branch is resolved, the speculated instruction is killed, the refill request will be killed (i.e. the speculated instruction will never be loaded into the cache). Despite being killed, the I$ still waits on the request from memory to be back, stalling all other operation. This implies that even with a hot cache, we can still experience cache miss stalls on speculated instructions which will never enter the cache, and hence the cache will never truely be "hot" (in this sense, speculated instructions should be considered part of the working set).

This PR attempts to improve performance in this condition by handling up to one killed refill request asynchronously. A speculated miss will only induce a stall if another miss occurs before the killed refill returns from memory.

PR #1804, opened 2024-02-05
cva6_icache: Allow one outstanding killed miss

colluca:
This PR fixes issue https://github.com/openhwgroup/cva6/issues/1785, implementing the same performance enhancement proposed in PR https://github.com/openhwgroup/cva6/pull/1497.

Compared to the latter, it fixes a problematic corner case, which was likely the cause of the regression errors observed. 

Said corner case occurred when the following two conditions were met:
- We incur in a second speculated instruction cache miss during a previous outstanding killed miss
- Upon branch resolution, the second refill is killed

The second miss brought the I$ in the `MISS` state, where it was correctly waiting for the previous outstanding killed miss to complete. However, the `kill_s2` signal was not being used here, effectively ignoring the branch resolution outcome.

This PR fixes this behaviour by counting the number of outstanding killed miss requests. Max one killed refill can be handled in parallel with other requests. Thus, as the second miss is killed, we stall the cache in the `WAIT_MISS` state, as originally was the case upon the first killed miss, so long as we don't again have a single outstanding miss.

I ran the smoke-tests locally, as described in the repo's README, and it seems to pass, but I'm not sure I'm doing this correctly.

### DIFF
--- a/.gitlab-ci/scripts/report_benchmark.py
+++ b/.gitlab-ci/scripts/report_benchmark.py
@@ -18,7 +18,7 @@ iterations = None
 # Will fail if the number of cycles is different from this one
 valid_cycles = {
     'dhrystone': 217900,
-    'coremark': 665193,
+    'coremark': 664736,
 }
 
 for arg in sys.argv[1:]:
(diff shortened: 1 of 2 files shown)


JeanRochCoulon:
Aie, when FPGA boot is fail, it is a mess. @Gchauvon would you have a root cause input ?

PR #1807, opened 2024-02-06
Add the condition for updating the tlb only after a miss is incurred

HudaSajjad:
Problem Statement:

In the TLB module of the MMU_sv32 unit during the update and flush mode, if the flush signal is high, the TLB is flushed according to the SFENCE.VMA command. However, in the event that the flush signal is not enabled the value in the TLB module is updated regardless of whether there was a hit or a miss in the TLB. If the updated value from the shared TLB is valid one of the entries in the TLB is updated with this value.

Proposed Solution:

During the update phase while checking that the updated value is valid or not it should also be checked that the TLB was missed and only in the case that the TLB was missed should the SHARED_TLB be accessed and an update value received from it.

### DIFF
--- a/core/mmu_sv32/cva6_tlb_sv32.sv
+++ b/core/mmu_sv32/cva6_tlb_sv32.sv
@@ -124,7 +124,7 @@ module cva6_tlb_sv32
         else if ((!content_q[i].g) && (vaddr_to_be_flushed_is0) && (asid_to_be_flushed_i == tags_q[i].asid[ASID_WIDTH-1:0]) && (!asid_to_be_flushed_is0))
           tags_n[i].valid = 1'b0;
         // normal replacement
-      end else if (update_i.valid & replace_en[i]) begin
+      end else if (update_i.valid & replace_en[i] & !lu_hit_o) begin
         // update tag array
         tags_n[i] = '{
             asid: update_i.asid,


JeanRochCoulon:
Thanks @HudaSajjad for this PR.  @AngelaGonzalezMarino Can I ask you to approve it ?

AngelaGonzalezMarino:
Sure, I will look at it and include it in the merged MMU

fatimasaleem:
> Sure, I will look at it and include it in the merged MMU

@AngelaGonzalezMarino you can look so that we can merge this separately, and MMU unifying work can continue alongside. Thanks.

fatimasaleem:
What is the update on this? @AngelaGonzalezMarino @JeanRochCoulon

AngelaGonzalezMarino:
Hi, it seems OK to me. I have tried with sv32 and sv39 configurations, both work OK (work means that Yocto boots on both of them). I don't see any performance gain either (booting time is the same). Is there a specific test I should do to see some change or benefit?

AngelaGonzalezMarino:
This is already integrated in the unified MMU https://github.com/openhwgroup/cva6/pull/1851

fatimasaleem:
Closing this one, since @AngelaGonzalezMarino included this change in https://github.com/openhwgroup/cva6/pull/1851

PR #1824, opened 2024-02-12
Enable reads on `CSR_HPM_COUNTERx`

luca-valente:
The privilege check is already in place in the `csr_regfile.sv`, but a read to `CSR_HPM_COUNTERx` should not return 0s.


JeanRochCoulon:
cv32a65x does not use this feature.
@zarubaf can you approve the is PR ?   I would prefer to forward you the features which are supported by your configuration, and not by cv32a65x.

PR #1863, opened 2024-02-21
Remove mcountinhibit from csr_test and UVM env

AyoubJalali:
This MR remove mcountinhibit from CSR tets and UVM environment


JeanRochCoulon:
This PR removes the dedicated tests related to the unimplemented CSR. But how is tested the illegal opcode exception when accessing the CSR ?

AyoubJalali:
> This PR removes the dedicated tests related to the unimplemented CSR. But how is tested the illegal opcode exception when accessing the CSR ?

it test it in the exception coverage model

PR #1874, opened 2024-02-26
[CI] cva6.py in Github CI

MarioOpenHWGroup:
Change the current Github CI to use `cva6.py`


JeanRochCoulon:
Hi @MarioOpenHWGroup , please to see this PR :-)

In Thales CI, we support Verilator 5.008 (and not v5.018 for which some limitations have been observed, also if I do not get clear visibility on them). Maybe @zchamski or @valentinThomazic could help to remind us.

verif/sim/setup-env.sh
```diff
@@ -30,7 +30,7 @@ fi
 
 # Set verilator and spike related variables
 if [ -z "$VERILATOR_INSTALL_DIR" ]; then
-    export VERILATOR_INSTALL_DIR="$ROOT_PROJECT"/tools/$(ls $ROOT_PROJECT/tools | grep verilator | head -n 1)
+    export VERILATOR_INSTALL_DIR="$ROOT_PROJECT"/tools/verilator-v5.018/
```
MarioOpenHWGroup, line 33: @valentinThomazic do you see it right?

valentinThomazic: His PR is waiting for this one to be merged, do you need theses changes to make the Github Actions work?

valentinThomazic: I think that setting `VERILATOR_BUILD_DIR` and `VERILATOR_INSTALL_DIR` to something like `tools/verilator` would make it easier since you would not have to care about the current used version of verilator. The install scripts will just install it there.

valentinThomazic: About your last commit, I'm not sure we understood each other !

I meant to say that if you set then env variable `VERILATOR_INSTALL-DIR` (and only this one, my previous comment was not completely accurate) in your CI, then I think you could leave `verif/sim/setup-env.sh` and `verif/regress/install-verilator.sh` as they are. 
Verilator will be installed in the specified path by the script. 

By doing so, having several verilator versions on `tools/` will still be a feature available to the users !

MarioOpenHWGroup: I reverted the changes to the `verif/sim/setup-env.sh` and `verif/regress/install-verilator.sh`. If you can double check now the changes :+1:

JeanRochCoulon:
Thanks @mario and @valentinThomazic for this PR. I merge

PR #1879, opened 2024-02-27
Connect CSRs info from RVFI_CSR in the testbench & update simulation target

AyoubJalali:
This MR content 2 commits : 
1. Connect CSRs info coming from RVFI (implamanted recently by @ycasamat) in the testbench
2. Update target simulation (use CV32A65X)


JeanRochCoulon:
In UVM testbench, RVFI is mapped to logic  (it is not used). This could explain the UVM test fails.
image

AyoubJalali:
It's related to some CSRs interface, I'm working on fix the Bug

PR #1881, opened 2024-03-04
Mmu user manual

AngelaGonzalezMarino:
Update the “RISC-V Virtual Memory” section in the CVA6 user manual with Sv32, Sv39 and Sv39x4 information after MMU merge

### DIFF
--- a/docs/01_cva6_user/Programmer_View.rst
+++ b/docs/01_cva6_user/Programmer_View.rst
@@ -169,11 +169,25 @@ CV64A6 supports the RISC-V **Sv39** virtual memory when the ``MMUEn`` parameter 
 Within CV64A6, the hypervisor extension is enabled and supports **Sv39x4** virtual memory when the ``CVA6ConfigHExtEn`` parameter is set to 1 (and ``Xlen`` is set to 64).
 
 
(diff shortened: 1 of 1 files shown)


jquevremont (approved):
LGTM. @fatimasaleem, can you double check?

fatimasaleem:
As far as I know, we have not integrated the hypervisor extension in the CVA6, and the parameter `CVA6ConfigHExtEn` does not exist in the latest master. Shouldn't we add the hypervisor related stuff in the user manual when it actually in the core?

jquevremont:
> As far as I know, we have not integrated the hypervisor extension in the CVA6, and the parameter `CVA6ConfigHExtEn` does not exist in the latest master. Shouldn't we add the hypervisor related stuff in the user manual when it actually in the core?

Thanks @fatimasaleem. We need to address the time until the hypervisor extension is merged.

I suggest to keep the text and add a note (that we'll remove when the hypervisor extension is merged). See https://github.com/planvtech/cva6/pull/42.

fatimasaleem:
Sounds good!

fatimasaleem (approved):

PR #1900, opened 2024-03-07
modify rvfi probes for param

yanicasa:
This PR makes sure that no instances of parametrized types go outside of CVA6, except RVFI-related types.
It will ease parametrization as types such as `scoreboard_entry_t` does not have to be transformed into a macro.

It additionally removes unused `IsRVFI` parameters.


zarubaf (approved):
🙌

Issue #1451, opened 2023-09-19
[TASK] Parametrization


JeanRochCoulon:
### Is there an existing CVA6 task for this?

- [x] I have searched the existing task issues

### Task Description

To be able to instantiate several CVA6 with different configurations, parameters must be defined as cva6 input parameters.

### Required Changes

Move parameters from packages to cva6 input parameters. But as not all the parameters can be configured, identify:
   - The paramaters that are actually used to configure the core.
   - The parameters that are not used now, e.g. the WB cache.
   - The parameters that are not used anymore, e.g. the renaming is not needed from GCC 13.

### Current Status

NOC done
RVFI done
Ariane_cfg_t ongoing
All others to be done

### Risks

To not finish the modification, to be in the middle of the bridge.

### Prerequisites

system verilog limitations

### KPI (KEY Performance Indicators)

remaining number of parameters to be moved from package to input parameters

### Description of Done

All parameters are moved from package to input parameters

zarubaf:
`Ariane_cfg_t` should be done.

JeanRochCoulon:
@zarubaf  I listed the parameters belonging to the config package to identify which are already in CVACfg ("done") and the others, some could be easily be parametrizable ("easy"), some less ("hard").
![image](https://github.com/openhwgroup/cva6/assets/56552370/56707115-a094-49c7-b3a4-0f618ecc29fc)

JeanRochCoulon:
Yes, #1233 was not defined as Kanban board task, but normal github issue. Thank for having added the link.

cathales:
Hello, I’m working on it

cathales:
I have started working on it last week but this week I did not have time to work on this topic.

cathales:
I am rebasing my previous work and will continue working on it.

cathales:
I have finished transferring `ariane_pkg` values which depend on configuration to the new parametrization system. However there are values in other packages that depend on these `ariane_pkg` "variable" values so I have to transfer them too.
It is not possible to elaborate until everything is transferred so I have to continue to completion before submitting changes.

It will require me a few more weeks before I can complete this task.

cathales:
`wt_cache_pkg` is now transferred too.

cathales:
Here is the current status: https://github.com/openhwgroup/cva6/pull/1704

Once this PR is completed and merged, the planned work will be to:

1. Remove all `localparam`s from config files so that only the `cva6_cfg` definition remains
[...]

PR #1908, opened 2024-03-08
Parametrization step 2

cathales:
This PR moves configurable types out of packages so that they can be configured by the configuration structure.

This is the second step for #1451

See also the announcement on Mattermost


JeanRochCoulon:
terrific !

zarubaf:
Awesome, thanks a lot @cathales!

PR #1911, opened 2024-03-09
Move re-aligner doc from ID stage to IF stage.

0BAB1:
# What does this PR do ?

After working on the 2024 softcore contest , I noticed that the re-aligner was not in ID but in IF, which makes more sense (the main intro CVA6 scheme also shows it).
Re-aligner doc was moved the re-aligner doc from ID stage to IF stage.
I might be wrong, looking forward getting your feedback.

### DIFF
--- a/docs/03_cva6_design/id_stage.md
+++ b/docs/03_cva6_design/id_stage.md
@@ -8,39 +8,10 @@ stage.
 With the introduction of compressed instructions (in general variable
 length instructions) the ID stage gets a little bit more complicated: It
 has to search the incoming data stream for potential instructions,
-re-align them and (in the case of compressed instructions) decompress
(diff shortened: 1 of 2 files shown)


JeanRochCoulon:
For information, this documentation is deprecated for a while. The cv32a65x is up-to-date and more accurate. Can you review the FRONTEND module of this cv32a65x document to confirm it is aligned on your expectations ?

0BAB1:
> For information, this documentation is deprecated for a while. The cv32a65x is up-to-date and more accurate. Can you review the FRONTEND module of this cv32a65x document to confirm it is aligned on your expectations ?

Yes, CV32A65X is way more up to date indeed and takes this change into account (and many others).
I'll go ahead and close that PR.

PR #1920, opened 2024-03-12
Allow single-stepping of commit stage from acc dispatcher

michael-platzer:
This PR gives the accelerator dispatcher the ability to single-step the commit stage, which avoids retiring instructions on another commit port than port 0. This is needed if the acc dispatcher cannot keep up with simultaneous commits on both commit ports, as well as for precise timing of pipeline flushes.


JeanRochCoulon:
LGTM, also if difficult for me to know whether the feature works

PR #1922, opened 2024-03-12
03 doc is deprecated

JeanRochCoulon:
GitHub Issues report 03 doc limitations. As it is not the main design document, we would like to notify that it is deprecated.


JeanRochCoulon:
@ASintzoff   Can you approve ?

PR #1962, opened 2024-03-25
AXI: DvPlan modification

AEzzejjari:
Modify the DvPlan according to changes in the AXI user guide


JeanRochCoulon:
@AEzzejjari could you integrate the #2065 feedbacks in the DV plan to be able to merge it ?

PR #1969, opened 2024-03-28
Functional coverage : no need for cvxif directed tests

AyoubJalali:
Hello, this MR remove cvxif directed tests, because the CVA32A65X doesn't support S or U mode, and as it known the cvxif directed test has one job to cover S & U mode request from the CVA6 to co-pro.

A [MR in core-v-verif](https://gitlab-tss.gemalto.com/riscv/ohg-pr/core-v-verif/-/merge_requests/93/diffs) ignore this bins if we don't support S or U mode

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -46,7 +46,7 @@ workflow:
 variables:
   GIT_SUBMODULE_STRATEGY: recursive
   DASHBOARD: cva6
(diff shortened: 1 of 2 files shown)


AyoubJalali:
we should update the core-v-verif version, before merging this PR

JeanRochCoulon:
You need to rebase your PR to ppint on the new core-v-verif

PR #1971, opened 2024-03-28
Modify the params order inside the cva6_user_cfg_t

JeanRochCoulon:
Modify the params order inside the cva6_user_cfg_t to gather extension params in top of structure and micro-architecture params in the bottom


cathales (approved):
I will update user configuration packages accordingly ASAP

cathales:
> I will update user configuration packages accordingly ASAP

All are not in the same (previous good) order so I won’t do it as of now. But I have other plans about it 😉

PR #2003, opened 2024-04-04
User config generator becomes a Python tool to work with configs

cathales:
The new Python script replaces the previous one. It is now not only a generator, but also a tool to get user-configuration as Python native variables. So we can re-use the same logic for distinct things.


cathales:
The Verible GitHub Action is modified in this PR to ignore configuration files.
However, the `verible.yml` file is taken from `master` branch, which does not have the modification yet; hence the comments.

Issue #2004, opened 2024-04-04
[BUG] MIE Supervisor Bits not RO if S is not present


MarioOpenHWGroup:
### Is there an existing CVA6 bug for this?

- [X] I have searched the existing bug issues

### Bug Description

The MIE fields `SSIE`, `STIE`, `SEIE` are not being disabled when supervisor is not enabled.

Hash: 5920e3d12596347e1796ef22e80e3c533182869f
Test: `csr_test.S`

JeanRochCoulon:
Bug is in RTL or Spike side ?

JeanRochCoulon:
RTL !

JeanRochCoulon:
I think (I hope) the issue has been closed by #2017. @MarioOpenHWGroup you can close if agree.

Issue #2008, opened 2024-04-05
[BUG] MIP Supervisor Bits not RO if S is not present


AyoubJalali:
### Is there an existing CVA6 bug for this?

- [X] I have searched the existing bug issues

### Bug Description

The MIP fields SSIP, STIP, SEIP are not being disabled when supervisor is not enabled.

Hash: https://github.com/openhwgroup/cva6/commit/5920e3d12596347e1796ef22e80e3c533182869f
Test: csr_test.S

![image](https://github.com/openhwgroup/cva6/assets/110232072/09251088-a00b-4ba6-888f-e0f32c073cf9)

PR #2017, opened 2024-04-08
Fix MIE CSR described in #2004 Github issue

JeanRochCoulon:
The MIE fields SSIE, STIE, SEIE are not being disabled when supervisor is not enabled. This PR fixes the issue.


AyoubJalali:
Hello,
After checking this PR on running the CSRs tests, it fix both issue #2004 & #2008

JeanRochCoulon:
Thanks @AyoubJalali for the pro-activity   :-)

Issue #1740, opened 2023-12-29
[BUG] : PMP CSRs


AyoubJalali:
### Is there an existing CVA6 bug for this?

- [X] I have searched the existing bug issues

### Bug Description

Hello,
In the RISCV spec section **3.7 Physical Memory Protection** : 
![image](https://github.com/openhwgroup/cva6/assets/110232072/f32dbb94-ffd5-463d-83bc-ec8268585407)

So in the pmpxcfg has the combination of X, W, R is reserved when W = 1 & R = 0 (as X, W, R form a collective WARL), so if we write a value in pmpxcfg which have this combination, we should clear only the bits X, W, R (normally clear W), but the CVA6 clear the entire pmpxcfg.

ASintzoff:
Do you mean when writing `b'xxxxxx10` where x can be 0 or 1 in a `pmpxcfg` byte, the result is `pmpxcfg` equal to 0?

Moschn:
This issue is fixed and can be closed

AyoubJalali:
Yes you right, I double check the fix !!
Thanks @Moschn

PR #2019, opened 2024-04-09
Fix PMPCFG WARL behavior

### LINKED ISSUE #1996: [Possible Issue] Non-selectable address matching mode leads to an incorrect PMP configuration update

### Bug Description

Instruction execution field `X`, and probably others, of PMP configuration CSRs is not updated correctly when trying to write non-selectable address matching value of field `A` in case PMP granularity `G` is `1`. Is this intended?

### RISC-V Specification
Unprivileged ISA: [20191213](https://drive.google.com/file/d/1s0lZxUZaa7eV_O0_WsZzaurFLLww7ou5/view) | Privileged Architecture: [20211203](https://drive.google.com/file/d/1EMip5dZlnypTk7pt4WWUKmtjUKTOkBqh/view) | External Debug Support: [0.13.02](https://drive.google.com/file/d/1XSPqW3WIfflIMeuGJBxGTMOEyeuRAQ5F/view)
> The PMP configuration registers are densely packed into CSRs to minimize context-switch time.
For RV32, sixteen CSRs, pmpcfg0–pmpcfg15, hold the configurations pmp0cfg–pmp63cfg for the 64
PMP entries, as shown in Figure 3.31.

> image

> Figure 3.35 shows the layout of a PMP configuration register. The R, W, and X bits, when set,
indicate that the PMP entry permits read, write, and instruction execution, respectively. When
one of these bits is clear, the corresponding access type is denied. The R, W, and X fields form a
collectiveWARL field for which the combinations with R=0 and W=1 are reserved. The remaining
two fields, A and L, are described in the following sections.

> image

> The A field in a PMP entry’s configuration register encodes the address-matching mode of the
associated PMP address register.

> image

> Although the PMP mechanism supports regions as small as four bytes, platforms may specify
coarser PMP regions. In general, the PMP grain is 2G+2 bytes and must be the same across all
PMP regions. When G ≥ 1, the NA4 mode is not selectable.

### Example Scenario

As shown below, the following sequence of instructions happens: 
> CSRRS x0, pmp_cfg0, x1

The instruction `CSRRS x0, pmp_cfg0, x1` is decoded at `t_id` and executed in machine mode resulting in updating the architecture state at `t_arch_update`. As a result, instruction execution field `X` of `pmp3cfg` is not updated as expected. Where it's supposed to be set according to the written value of `x1`. Probably the reason for that is trying to write `NA4` value to the `A` field of the same register `pmp3cfg`, which is not selectable when the granularity `G` is `1`, and thus updating `A` is ignored correctly, but `X` has to be updated correctly still!

Issue_15

### Steps to Reproduce

Git Hash: [de2e254c](https://github.com/openhwgroup/cva6/commit/de2e254cd4c2d3e3a6d4def1652918e8f2535512) | TARGET_CFG: [cv32a6_imac_sv0](https://github.com/openhwgroup/cva6/blob/master/core/include/cv32a6_imac_sv0_config_pkg.sv) | VCD: [issue_15.zip](https://github.com/openhwgroup/cva6/files/14836579/issue_15.zip)

### Component
**Component:RTL**

---
Product: Questa OneSpin Solutions
App: Questa Processor App
Tool's version: 2024.1_1

Moschn:
Change PMPCFG write behavior to support individual WARL fields. 

Fixes #1996, #1740

Note: I did not test this patch. Ideally, one of the reporters of #1996 and #1740 would check this before merging. Also, I never tested the PMP read-only feature, and I hope I did not break anything there (@michael-platzer).

### DIFF
--- a/core/csr_regfile.sv
+++ b/core/csr_regfile.sv
@@ -2431,15 +2431,15 @@ module csr_regfile
   always_comb begin : write
     for (int i = 0; i < 16; i++) begin
       if (i < CVA6Cfg.NrPMPEntries) begin
-        if(!CVA6Cfg.PMPEntryReadOnly[i]) begin
+        if (!CVA6Cfg.PMPEntryReadOnly[i]) begin
           // PMP locked logic is handled in the CSR write process above
           pmpcfg_next[i] = pmpcfg_d[i];
           // We only support >=8-byte granularity, NA4 is disabled
-          if(pmpcfg_d[i].addr_mode == riscv::NA4) begin
+          if (pmpcfg_d[i].addr_mode == riscv::NA4) begin
             pmpcfg_next[i].addr_mode = pmpcfg_q[i].addr_mode;
           end
           // Follow collective WARL spec for RWX fields
-          if(pmpcfg_d[i].access_type.r == '0 && pmpcfg_d[i].access_type.w == '1) begin
+          if (pmpcfg_d[i].access_type.r == '0 && pmpcfg_d[i].access_type.w == '1) begin
             pmpcfg_next[i].access_type = pmpcfg_q[i].access_type;
           end
         end else begin
--- a/core/csr_regfile.sv
+++ b/core/csr_regfile.sv
@@ -2431,9 +2431,17 @@ module csr_regfile
   always_comb begin : write
     for (int i = 0; i < 16; i++) begin
       if (i < CVA6Cfg.NrPMPEntries) begin
-        // We only support >=8-byte granularity, NA4 is disabled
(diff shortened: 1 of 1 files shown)


AyoubJalali:
Thanks for the Fix, if I understand normally the CVA6 should not update the pmpcfgx CSR if you try to write a reserved value (pmpxcfg.R == 0 & pmpxcfg.W ==1) is that right @Moschn !!

Moschn:
> Thanks for the Fix, if I understand normally the CVA6 should not update the pmpcfgx CSR if you try to write a reserved value (pmpxcfg.R == 0 & pmpxcfg.W ==1) is that right @Moschn !!

Yes that should be correct. With this fix applied, CVA6 should ignore any write to the pmpcfg RWX fields if R==0 and W==1. Before this fix, CVA6 treated the entire PMPCFG as one collective WARL field and thus did not update the A and L fields either if R==0 and W==1.

Issue #1758, opened 2024-01-12
[BUG] `zext.h_64-01` verilator simulation failed


Siris-Limx:
### Is there an existing CVA6 bug for this?

- [X] I have searched the existing bug issues

### Bug Description

I can pass smoke test by running `bash verif/regress/smoke-tests.sh`
However, when I try to run `bash verif/regress/dv-riscv-arch-test.sh`, it will terminate when testing `zext.h_64-01`, here is the output log:
```
Makefile:14: must set CVA6_REPO_DIR to point at the root of CVA6 sources and CVA6_TB_DIR to point here -- doing it for you...
make -C /home/sirisli/cva6/ verilate verilator="verilator --no-timing" target=cv64a6_imafdc_sv39 defines=
make[1]: Entering directory '/home/sirisli/cva6'
Makefile:143: XCELIUM_HOME not set which is necessary for compiling DPIs when using XCELIUM
[Verilator] Building Model
....
cd work-ver && make -j24 -f Variane_testharness.mk
make[2]: Entering directory '/home/sirisli/cva6/work-ver'
make[2]: Nothing to be done for 'default'.
make[2]: Leaving directory '/home/sirisli/cva6/work-ver'
make[1]: Leaving directory '/home/sirisli/cva6'
/home/sirisli/cva6//work-ver/Variane_testharness   /home/sirisli/cva6/verif/sim/out_2024-01-12/directed_asm_tests/zext.h_64-01.o +debug_disable=1 +ntb_random_seed=1 \
  +elf_file=/home/sirisli/cva6/verif/sim/out_2024-01-12/directed_asm_tests/zext.h_64-01.o +tohost_addr=0000000080002000
This emulator compiled with JTAG Remote Bitbang client. To enable, use +jtag_rbb_enable=1.
Listening on port 39375
/home/sirisli/cva6/verif/sim/out_2024-01-12/directed_asm_tests/zext.h_64-01.o *** FAILED *** (tohost = 2147483647) after 2000012 cycles
CPU time used: 31653.16 ms
Wall clock time passed: 31649.59 ms
make: *** [Makefile:141: veri-testharness] Error 255
```

Edit:
I have tired other cases in `dv-riscv-arch-test.sh`, they all passed, that is, only `zext.h_64-01` failed.
I notice only when running `zext.h_64-01` case, verilator will raise these warnings:
<details> <summary>Verilator Warnings</summary>

```
[...]

PR #2020, opened 2024-04-09
Fail Github CI pipeline on test failure

WorldofJARcraft:
On a local clone of the repository, I ran the `dv-riscv-arch-test.sh` script and experienced a test failure for the `riscv-test-suite/rv64i_m/B/src/zext.h_64-01.S` test in verilator. I first assumed something was set up incorrectly on my end, but then I checked the CI logs and found that the same test also fails in the latest [execute-riscv-tests (dv-riscv-arch-test, veri-testharness)](https://github.com/openhwgroup/cva6/actions/runs/8156293959/job/22293603774#logs) run.
However, the test script always exits with code 0, which is why the pipeline succeeds anyways.
This pull request fails the pipeline in case tests error out.

### DIFF
--- a/verif/regress/dv-riscv-arch-test.sh
+++ b/verif/regress/dv-riscv-arch-test.sh
@@ -28,6 +28,10 @@ if ! [ -n "$DV_SIMULATORS" ]; then
   DV_SIMULATORS=veri-testharness,spike
 fi
 
+EXITVAL=0
+
 cd verif/sim
-python3 cva6.py --testlist=../tests/testlist_riscv-arch-test-$DV_TARGET.yaml --target $DV_TARGET --iss_yaml=cva6.yaml --iss=$DV_SIMULATORS $DV_OPTS --linker=../tests/riscv-arch-test/riscv-target/spike/link.ld
+python3 cva6.py --testlist=../tests/testlist_riscv-arch-test-$DV_TARGET.yaml --target $DV_TARGET --iss_yaml=cva6.yaml --iss=$DV_SIMULATORS $DV_OPTS --linker=../tests/riscv-arch-test/riscv-target/spike/link.ld || EXITVAL=1
 cd -
+
+exit $EXITVAL
--- a/verif/regress/smoke-tests.sh
+++ b/verif/regress/smoke-tests.sh
@@ -38,36 +38,40 @@ fi
 
 export DV_OPTS="$DV_OPTS --issrun_opts=+debug_disable=1+UVM_VERBOSITY=$UVM_VERBOSITY"
 
-cd verif/sim/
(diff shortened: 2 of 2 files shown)


WorldofJARcraft:
An issue for the underlying bug already exists: https://github.com/openhwgroup/cva6/issues/1758

cathales:
Hello @WorldofJARcraft 

Thanks for this contribution

1. On GitLab we generate a yaml report to build HTML pages with the status. On GitHub I don’t know what we do. Cc @MarioOpenHWGroup 
2. Here you want to make the script fail so that the job fails. We have to check the impact on GitLab CI jobs too, to avoid a systematic 0/0 with no details when one test fails. A change in `.gitlab-ci.yml` could be needed. Cc @valentinThomazic 
3. Is it relevant to run all tests even if the first one failed ("no fail fast" mode)? What do you all think about `set -e` to stop after the first fail ("fail fast")? It would introduce way less diffs in this PR (only 1 line added in each file).

WorldofJARcraft:
> Hello @WorldofJARcraft
> 
> Thanks for this contribution
> 
>     1. On GitLab we generate a yaml report to build HTML pages with the status. On GitHub I don’t know what we do. Cc @MarioOpenHWGroup
> 
>     2. Here you want to make the script fail so that the job fails. We have to check the impact on GitLab CI jobs too, to avoid a systematic 0/0 with no details when one test fails. A change in `.gitlab-ci.yml` could be needed. Cc @valentinThomazic
> 
>     3. Is it relevant to run all tests even if the first one failed ("no fail fast" mode)? What do you all think about `set -e` to stop after the first fail ("fail fast")? It would introduce way less diffs in this PR (only 1 line added in each file).

Thanks for your reply!
I think `set -e` as you suggested would work. All that matters is that a non-zero exit code is returned to the CI when one or more tests fail.
As to whether the remaining tests should be executed after a test failed, I am not sure what the best course of action is. I will leave this decision to the maintainers.

cathales:
Hello @WorldofJARcraft 

After an internal discussion, we need the no-fail-fast mode.

The `verif/regress/*.sh` scripts are meant to be run locally too, so it might be better to not change their behavior.
We could have a script which gets the simulation result, just like we do in GitLab, except we would set exit code instead of generating a yaml script. A way to do so could be to just update the existing script, to detect the Dashboard-related environment values.

- If they are provided (GitLab), generate a yaml report using these values (current behavior).
- If they are not provided (GitHub), set exit code.

So two changes would be made:

1. Updating the Python script as described above
2. Run this Python script after running tests in GitHub CI

Note that the simulation result format will change in the following weeks so it could be better to wait for @valentinThomazic to be back to discuss this.

PR #2025, opened 2024-04-09
Move `DCacheType` to config struct

zarubaf:
As discussed on Mattermost, this moves the DCacheType into the new parameterization. I am also sneaking in some other Bender changes that should not make troubles for other users.

A general note: We are maintaining quite a lot of different configurations, wondering whether that makes sense.

/cc @domenicw

### DIFF
--- a/Bender.yml
+++ b/Bender.yml
@@ -14,154 +14,93 @@ dependencies:
   tech_cells_generic:
     { git: "https://github.com/pulp-platform/tech_cells_generic.git", version: 0.2.13 }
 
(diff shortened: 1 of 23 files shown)


cathales:
May I run some tests locally this afternoon before we merge?

zarubaf:
> May I run some tests locally this afternoon before we merge?

Sure

cathales (approved):
Okay I have found the issue I was looking for. I'll push a patch once this PR is merged.

PR #2030, opened 2024-04-11
fix exit code

### LINKED ISSUE #2012: [BUG] Unexpected Behavior in M Extension Instructions with Identical Operands

### Bug Description

When executing instructions from the M extension (such as `div a0, a0, a0`, `divu a0, a0, a0`, etc.) and `rd = rs1 = rs2 = a0`, the cva6 cannot operate normally to obtain the expected value.

youzi27:
To fix #2012

The previous code would fail to exit normally under certain circumstances. 

For example, before exiting, `a0=0b01` (which becomes `0b10` after left shifting), so it is necessary to clear the value of `a0` before exiting.

### DIFF
--- a/verif/tests/custom/common/crt.S
+++ b/verif/tests/custom/common/crt.S
@@ -245,8 +245,7 @@ exit:
   // Minimal low-level exit function.
   .global  _exit
 _exit:
-  sll  a0, a0, 1      # Set up exit code: shift a0 left by 1
-  add  a0, a0, 1      # and set bit 0.
+  li   a0, 0x1        # and set bit 0.
   sw   a0, tohost, t5 # Store bits [31:0] of a0 into variable tohost
   nop                 # Leave one cycle of slack before entering endless loop
 1:


youzi27:
Hi @JeanRochCoulon ,

If preserving the original information value is necessary, I believe it's also possible to `XOR a0` with another register to ensure that the value of `a0` is not lost (`a0` still affects the final architectural state). Because the current code indeed cannot exit the simulation normally.

Thank you.

zchamski:
> With the current PR, the value contained by a0 at the _exit entry will be lost, which is not ok. The goal of _exit function is to end the simulation by setting 1 in bit[0] and keeping the a0 entry value in bit [31:1].
> 
> @zchamski do you confirm it ?

Yes, I confirm: the value written into `tohost` consists of the two parts as described by @JeanRochCoulon.  This is the end-of-test message format used by Berkeley HTIF protocol and we want to follow it.

**NOTE:** `tohost` is a 64-bit variable and to be fully compliant with HTIF, we should write 16'b0 into bits [63:48] of `tohost`.  For XLEN == 32 we terminate the test on the write into  `tohost`'s low 32 bits, so we do not care about bits [63:32].

PR #2041, opened 2024-04-15
Add ifdef for VCS fix (UNSUPPORTED_WITH)

CoralieAllioux:
This fix is suggested in the context of supporting Cadence Xcelium simulator for cva6 verification.

There is a VCS bug which does not support correctly the difference between `with` and `iff` keywords. A missing define needs to be added in order to allow xcelium to compile.

This has been check with VCS 2021.09.


JeanRochCoulon:
@ASintzoff @AyoubJalali Can you give your feedback?

PR #2047, opened 2024-04-16
Fix `SuperScalar` config and add `CVA6Cfg` to first pass decoder

zarubaf:
The unused `SuperScalar` config had `SELRANGE` issues with `verilator`. Since `[1]` is not available in `SuperScalar = 0`.

### DIFF
--- a/core/frontend/instr_queue.sv
+++ b/core/frontend/instr_queue.sv
@@ -88,6 +88,9 @@ module instr_queue
     input logic [ariane_pkg::SUPERSCALAR:0] fetch_entry_ready_i
 );
 
+  // Calculate next index based on whether superscalar is enabled or not.
+  localparam NID = ariane_pkg::SUPERSCALAR > 0 ? 1 : 0;
+
   typedef struct packed {
     logic [31:0]                     instr;      // instruction word
     ariane_pkg::cf_t                 cf;         // branch was taken
@@ -302,9 +305,9 @@ ariane_pkg::FETCH_FIFO_DEPTH
   // ----------------------
   // as long as there is at least one queue which can take the value we have a valid instruction
   assign fetch_entry_valid_o[0] = ~(&instr_queue_empty);
-  if (ariane_pkg::SUPERSCALAR) begin : gen_fetch_entry_valid_1
+  if (ariane_pkg::SUPERSCALAR > 0) begin : gen_fetch_entry_valid_1
     // TODO Maybe this additional fetch_entry_is_cf check is useless as issue-stage already performs it?
-    assign fetch_entry_valid_o[1] = ~|(instr_queue_empty & idx_ds[1]) & ~(&fetch_entry_is_cf);
+    assign fetch_entry_valid_o[NID] = ~|(instr_queue_empty & idx_ds[1]) & ~(&fetch_entry_is_cf);
   end
 
   assign idx_ds[0] = idx_ds_q;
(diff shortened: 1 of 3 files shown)


core/frontend/instr_queue.sv
```diff
@@ -88,6 +88,9 @@ module instr_queue
     input logic [ariane_pkg::SUPERSCALAR:0] fetch_entry_ready_i
 );
 
+  // Calculate next index based on whether superscalar is enabled or not.
+  localparam NID = ariane_pkg::SUPERSCALAR > 0 ? 1 : 0;
```
cathales, line 92: So this is just `localparam NID = ariane_pkg::SUPERSCALAR`.

Also, IIUC this is not a real "next index". It is just a workaround so that Verilator does not cry because of out-of-range issues.

core/frontend/instr_queue.sv
```diff
@@ -340,7 +343,7 @@ ariane_pkg::FETCH_FIFO_DEPTH
 
       // output mux select
       for (int unsigned i = 0; i < CVA6Cfg.INSTR_PER_FETCH; i++) begin
-        // TODO handle fetch_entry_o[1] if superscalar
+        // TODO handle fetch_entry_o[NID] if superscalar
```
cathales, line 346: IMHO "1" is more clear than "NID". Anyway this comment will disappear soon so this change would just create a conflict for me to fix.

```suggestion
        // TODO handle fetch_entry_o[1] if superscalar
```

core/frontend/instr_queue.sv
```diff
@@ -364,26 +367,26 @@ ariane_pkg::FETCH_FIFO_DEPTH
           pop_instr[i] = fetch_entry_fire[0];
         end
 
-        if (ariane_pkg::SUPERSCALAR) begin
+        if (ariane_pkg::SUPERSCALAR > 0) begin
```
cathales, line 370: Well yes 😅  When we actually parametrize `SuperscalarEn`, we will likely define both `int unsigned PIPELINE_WITH` and `bit SuperscalarEn`.

cathales (changes requested):
So this PR addresses two topics:

1. It passes `CVA6Cfg` to first pass decoder
2. It fixes some Verilator warnings about superscalar modifications. I’m wondering:

    - Is it better to perform changes on the fly like this? It makes PRs more little and keeps the code base clean, but it will likely trigger conflicts with the subsequent superscalar-related PRs.
    - Is it better to collect the issues so that we definitely fix them once superscalar is completely merged, on July 5th? Because parametrization will have to be applied here.

zarubaf:
I can split up the thing in two PRs, but I would really prefer if we can merge the fixed to the superscalar parameterization. I am running behind with the bumping of CVA6 for a couple of weeks now since every day stuff is changing which breaks the instantiation.

core/frontend/instr_queue.sv
zarubaf: Ah I thought superscalar can be more than 1.

cathales:
IMO if we keep both changes we can have them in this same PR.

I have taken notes for the future definitive fixes so we can merge this.

Still, I find it really sad that despite the explicit `if` blocks to filter Superscalar-related stuff, Verilator does not understand that the things inside the blocks do not exist. I hope at least code coverage tools will understand.

zarubaf:
> IMO if we keep both changes we can have them in this same PR.
> 
> I have taken notes for the future definitive fixes so we can merge this.
> 
> Still, I find it really sad that despite the explicit `if` blocks to filter Superscalar-related stuff, Verilator does not understand that the things inside the blocks do not exist. I hope at least code coverage tools will understand.

I agree, but I am afraid that if `verilator` already struggles I am unsure what Spyglass and similar tools will say. Let's cross the bridge once we are there.

If you are okay with leaving everything in this PR, I should have addressed your concerns (committed the changes). And I also saw that you took notice about the parameterization. Are you okay with merging?

PR #2054, opened 2024-04-17
Generate the cv32a65x riscv specification out of the box

JeanRochCoulon:
Customized Risc-V spec to suite cv32a65x CVA6 config

Next : as the generation is out of the box, the vendorization of riscv-isa-manual can be replaced by submodule.


JeanRochCoulon:
@jquevremont  The riscv specification are customized to suite to cv32a65x. That's why the riscv specification are modified to specify the implementation options. This PR corresponds to the infrastructure setup. The implementation options will come in next PRs. Can you review it ?

PR #2063, opened 2024-04-21
FixBug: Add handling for +INF in sqrt

### LINKED ISSUE #2058: [BUG] Incorrect Accumulation of ‘OF’ Flag in fflags After Executing fsqrt.d on Infinity

### Bug Description

When executing the `fsqrt.d` instruction on a double-precision floating-point value representing positive infinity (0x7ff0000000000000), the Overflow (OF) flag in the fflags register is erroneously set. According to the IEEE 754 standard and the RISC-V specification, the fsqrt.d operation **should not** lead to an overflow situation when the input is infinity. This also results in inconsistency with Spike's output

**Expected Behavior:**
The `OF` flag in the fflags register should remain clear (i.e., not set) after performing a square root operation on an infinite value, as the result is well-defined and should be positive infinity.

**Actual Behavior:**
The `OF` flag is set in the fflags register, indicating an overflow, which contradicts the expected behavior defined by the IEEE 754 standard and RISC-V floating-point operation guidelines.

**Steps to Reproduce:**

1. Load a double-precision floating-point register with the value `0x7ff0000000000000` (positive infinity).
2. Execute the `fsqrt.d` instruction on this register.
3. Check the `fflags` register; observe that the `OF` flag is incorrectly set.

youzi27:
To Fix #2058 . 

`sqrt` should not trigger the `OF`(Overflow).

### DIFF
--- a/vendor/openhwgroup/cvfpu/src/fpu_div_sqrt_mvp/hdl/norm_div_sqrt_mvp.sv
+++ b/vendor/openhwgroup/cvfpu/src/fpu_div_sqrt_mvp/hdl/norm_div_sqrt_mvp.sv
@@ -150,25 +150,39 @@ module norm_div_sqrt_mvp
               Sign_res_D=1'b0;
               NV_OP_S = 1'b1;
             end
-          else if (Sqrt_enable_SI && Sign_in_DI) begin // catch sqrt(-inf)
-            Div_Zero_S=1'b0;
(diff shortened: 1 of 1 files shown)


youzi27:
I have resubmitted the PR in the [fpu_div_sqrt_mvp](https://github.com/pulp-platform/fpu_div_sqrt_mvp/pulls), so I will close this PR. Thank you.

PR #2064, opened 2024-04-21
FixBug: 0/0 should not set DZ fflags

### LINKED ISSUE #2061: [BUG] Incorrect Setting of DZ Flag for 0/0 Division

### Bug Description

I have observed an incorrect behavior in the CVA6 core regarding the handling of the 0/0 division operation. Specifically, the Divide by Zero (`DZ`) flag is incorrectly set for this operation, which according to IEEE 754 standards (Standard for Floating-Point Arithmetic section 7.3):

>7.3 Division by zero 7.3.0 
The divideByZero exception shall be signaled if and only if an exact infinite result is defined for an  operation on finite operands. The default result of divideByZero shall be an ∞ correctly signed according to the operation:
― For division, when the divisor is zero and the dividend is a finite non-zero number, the sign of the infinity is the exclusive OR of the operands’ signs (see 6.3).

So, it should only trigger the Invalid Operation (`NV`) flag, since 0/0 is an undefined operation resulting in NaN.

**The log is as follows:**
```
core   0: 0x0000000080005aa0 (0x001025f3) csrrs   a1, fflags, zero
1 0x0000000080005aa0 (0x001025f3) x11 0x0000000000000000
core   0: 0x0000000080005aa4 (0x221080d3) fsgnj.d ft1, ft1, ft1
1 0x0000000080005aa4 (0x221080d3) f 1 0x0000000000000000
core   0: 0x0000000080005aa8 (0x1a10f953) fdiv.d  fs2, ft1, ft1
1 0x0000000080005aa8 (0x1a10f953) f18 0x7ff8000000000000
core   0: 0x0000000080005aac (0x00102673) csrrs   a2, fflags, zero
1 0x0000000080005aac (0x00102673) x12 0x0000000000000018
```

youzi27:
Fix #2061 .

### DIFF
--- a/vendor/openhwgroup/cvfpu/src/fpu_div_sqrt_mvp/hdl/norm_div_sqrt_mvp.sv
+++ b/vendor/openhwgroup/cvfpu/src/fpu_div_sqrt_mvp/hdl/norm_div_sqrt_mvp.sv
@@ -185,7 +185,7 @@ module norm_div_sqrt_mvp
 
      else if(Zero_a_SI)
        begin
-         if(Div_enable_SI&&Zero_b_SI)
+        if(Div_enable_SI && !Zero_b_SI)              // x/0, x != 0
            begin
               Div_Zero_S=1'b1;
               Exp_OF_S=1'b0;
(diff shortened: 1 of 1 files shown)


JeanRochCoulon:
Hello @youzi27 
Thank you for your verification and RTL fix effort.
The file you are fixing does not belong to CVA6 repository. It comes from [fpu_new ](https://github.com/openhwgroup/cvfpu) repository which has been "vendorized" into cva6. That is why I would suggest to double check whether the bug is fixed in this repo and, if it is not the case, submit your PR to cvfpu repository.
Best

To become OpenHW Group contributor, could you sign the Eclipse agreementfdn/eca ?

youzi27:
Hello,

Thanks for the heads up about the file! I'll check if the bug is fixed in the fpu_new repo and if not, I'll submit the PR to cvfpu as suggested.

Also, I'm on it with the Eclipse agreement for joining the OpenHW Group.

youzi27:
Thank you for your reminder. 
I have resubmitted the issue and PR in the [fpu_div_sqrt_mvp @ 86e1f55](https://github.com/pulp-platform/fpu_div_sqrt_mvp/tree/86e1f558b3c95e91577c41b2fc452c86b04e85ac). May I close this PR?

Issue #1586, opened 2023-10-31
[TASK] Add extension and features tags on tests in testlists


Gchauvon:
### Is there an existing CVA6 task for this?

- [X] I have searched the existing task issues

### Task Description

 Add tags on tests in testlists to identify extension and/or micro-architectural features required by CVA6 to run the tagged test.
There is an example in [testlist_hwconfig.yaml](https://github.com/openhwgroup/cva6/blob/master/verif/tests/testlist_hwconfig.yaml)

```
- test: rv32ui-v-add
  iterations: 1
  needs:
    - xlen: 32
    - fpu: 1
    - c_ext: 1
  path_var: TESTS_PATH
  gcc_opts: "-DENTROPY=0x1 -static -mcmodel=medany -fvisibility=hidden -nostdlib -nostartfiles -I<path_var>/riscv-tests/isa/macros/scalar/ <path_var>/riscv-tests/env/v/entry.S <path_var>/riscv-tests/env/v/vm.c <path_var>/riscv-tests/env/v/string.c -I<path_var>/riscv-tests/env/v/ -I<path_var>/riscv-tests/riscv-target/spike/"
  asm_tests: <path_var>/riscv-tests/isa/rv32ui/add.S
```

### Required Changes

- Keep only one testlist for each regression ( riscv-tests, riscv-compliance, riscv-arch-test, ...)
- Tag all tests with extension and/or micro-architectural features in CVA6 required to run the specific test
- Compare the list of required extension and/or micro-architectural features with the configuration of the CVA6 used to run the test ( core/include/*_config_pkg.sv )
- CVA6 should only run tests that are compatible with its configuration.

### Current Status

An idea of what is expected is written in  [testlist_hwconfig.yaml](https://github.com/openhwgroup/cva6/blob/master/verif/tests/testlist_hwconfig.yaml), [config_pkg_generator](https://github.com/openhwgroup/cva6/blob/d92f0f76d053182c7ab41bc94bcf1a61e9757b00/util/config_pkg_generator.py) and [cva6.py](https://github.com/openhwgroup/cva6/blob/d92f0f76d053182c7ab41bc94bcf1a61e9757b00/verif/sim/cva6.py).

### Description of Done

Existing testlists are removed and replaced with tagged one.

JeanRochCoulon:
Related to #1569

JeanRochCoulon:
@valentinThomazic will address it on July...

PR #2073, opened 2024-04-24
Implement YAML anchor/alias for streamlined testlist structure

xiaoweish:
**Problem Statement**
The current testlist (e.g., [testlist_riscv-compliance-cv32a60x.yaml](https://github.com/openhwgroup/cva6/blob/master/verif/tests/testlist_riscv-compliance-cv32a60x.yaml)) contains many duplicate lines, such as `gcc_opts` and `path_var: TESTS_PATH`, while this repetition is necessary for each individual test case, it leads to a bloated file size and reduced readability.

**Proposed Solution**
1. Utilize YAML anchors and aliases to make the testlist YAML file more compact, as described in [YAML anchors](https://support.atlassian.com/bitbucket-cloud/docs/yaml-anchors/).
2. Introduce a new keyword **`testlist`** in the YAML file.
3. Ensure backward compatibility with the existing implementation.

The proposed changes will make the testlist YAML file more compact and readable.


JeanRochCoulon:
Thank you @xiaoweish fot this PR submission. Indeed the testlists are big in term of size.
It looks good to me to implement such an optimization.
But applying a patch, some dv submodule files will get the "Changed not stage for commit" attribute, it cannot be allowed.
Why not copy the dv function you would like to modify "out of the box" (outside the dv submodule). By renaming the function, you could call it from cva6.py.

BTW: to reduce the number of testlists, we get the idea to add a parameter in the testlist: https://github.com/openhwgroup/cva6/issues/1586

xiaoweish:
Updated. Thanks Jean-roch for your suggestion! #1586 is a good idea, indeed.

JeanRochCoulon:
Well done @xiaoweish !

PR #2082, opened 2024-04-26
Allow to pass custom location for the Boost library for Spike

cfuguet:
This allows to pass a custom location for the Boost library when compiling Spike.

In some cases, it is not possible to install libraries in standard locations using package managers (apt, yum, etc). For example, when you do not have administrator rights on the host machine.


JeanRochCoulon:
@zchamski   can you approve it ?

zchamski:
> @zchamski can you approve it ?

LGTM!

PR #2084, opened 2024-04-30
Address translation request moved from icache to frontend

yanicasa:
In order to aligned the ways address translation is managed between FETCH and DATA the address translation request is moved from the icache to the frontend.

The change is supposed to be transparent, no functional impact, same perf, same critical path...

### DIFF
--- a/core/cache_subsystem/cva6_icache.sv
+++ b/core/cache_subsystem/cva6_icache.sv
@@ -30,8 +30,6 @@ module cva6_icache
   import wt_cache_pkg::*;
 #(
     parameter config_pkg::cva6_cfg_t CVA6Cfg = config_pkg::cva6_cfg_empty,
-    parameter type icache_areq_t = logic,
-    parameter type icache_arsp_t = logic,
     parameter type icache_dreq_t = logic,
     parameter type icache_drsp_t = logic,
     parameter type icache_req_t = logic,
@@ -48,9 +46,6 @@ module cva6_icache
     input  logic         en_i,
     /// to performance counter
     output logic         miss_o,
-    // address translation requests
-    input  icache_areq_t areq_i,
-    output icache_arsp_t areq_o,
     // data requests
     input  icache_dreq_t dreq_i,
     output icache_drsp_t dreq_o,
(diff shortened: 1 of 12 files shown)


yanicasa:
@cfuguet can you review this PR, maybe this modif could be useful if you are planning to design an HPICache for CVA6

core/cva6.sv
```diff
@@ -50,31 +50,32 @@ module cva6
...
     localparam type icache_dreq_t = struct packed {
-      logic                    req;      // we request a new word
-      logic                    kill_s1;  // kill the current request
-      logic                    kill_s2;  // kill the last request
-      logic                    spec;     // request is speculative
-      logic [CVA6Cfg.VLEN-1:0] vaddr;    // 1st cycle: 12 bit index is taken for lookup
+      logic                    req;                // we request a new word
+      logic                    kill_s1;            // kill the current request
+      logic                    kill_s2;            // kill the last request
+      logic                    kill_req;           // kill the last request
```
cfuguet, line 68: This is more a note for me while I take a look on all files: Why do you need an additional kill signal ?

core/cva6.sv
```diff
@@ -50,31 +50,32 @@ module cva6
...
+      logic                    translation_valid;  // address translation valid
+      logic [CVA6Cfg.PLEN-1:0] paddr;              // 1st cycle: 12 bit index is taken for lookup
     },
     localparam type icache_drsp_t = struct packed {
       logic                                ready;  // icache is ready
       logic                                valid;  // signals a valid read
       logic [CVA6Cfg.FETCH_WIDTH-1:0]      data;   // 2+ cycle out: tag
       logic [CVA6Cfg.FETCH_USER_WIDTH-1:0] user;   // User bits
-      logic [CVA6Cfg.VLEN-1:0]             vaddr;  // virtual address out
-      exception_t                          ex;     // we've encountered an exception
```
cfuguet, line 77: I agree that the previous version of the Icache did not forward any error from the NoC interface (bad address access for example). However, I think it should... Hence, I would recommend to leave this field of the response, an use it to forward an error to the fetch unit.

For this PR, probably you can force it to 0, so you only modify a single functionality, but it can be used in another PR to support this error forwarding.

cfuguet (changes requested):

core/cva6.sv
cfuguet: With my comment below regarding the handling of an address translation exception, this field (kill_req) should not be needed anymore.

cfuguet:
Hi @yanicasa, I left you some comments

Best regards

core/cva6.sv
yanicasa: Maybe this could be merged with kill_s2 but this has an impact on the icache FSM READ state description and I wanted to be conservative.

This new icache kill_req follows the same principle as the dcache kill_req, it is an abort caused by an MMU exception during address translation

Note: the copy paste "// kill the last request" should me updated.

cfuguet: I would merge with the kill_s1. The exception is valid one cycle after the address translation request. I think there should not be any significant modification in the Icache FSM when doing that

core/cva6.sv
yanicasa: In fact this field was only used to forward the MMU exception towards the frontend, since it was the cache which received the response from the MMU.
I removed it in symmetry with the Dcache because Dcache also does not send any exception to the LSU.

If we add exception filed for Icache, we would have to do it for Dcache too, maybe it could be done in a dedicated PR if required?

core/cache_subsystem/cva6_icache.sv
```diff
@@ -261,18 +250,17 @@ module cva6_icache
...
+        cache_rden = cache_en_q;
 
-        if (areq_i.fetch_valid && (!dreq_i.spec || ((CVA6Cfg.NonIdemPotenceEn && !addr_ni) || (!CVA6Cfg.NonIdemPotenceEn)))) begin
+        if (dreq_i.translation_valid && (!dreq_i.spec || ((CVA6Cfg.NonIdemPotenceEn && !addr_ni) || (!CVA6Cfg.NonIdemPotenceEn)))) begin
           // check if we have to flush
           if (flush_d) begin
             state_d = IDLE;
             // we have a hit or an exception output valid result
-          end else if (((|cl_hit && cache_en_q) || areq_i.fetch_exception.valid) && !inv_q) begin
+          end else if (((|cl_hit && cache_en_q) || dreq_i.kill_req) && !inv_q) begin
```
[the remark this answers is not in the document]
yanicasa: I will to check to remove one of the 3 kill signals which actually seems to be superfluous.

On the other hand, the kill_req for me comes in second time and I have the impression that it will be more linked to kill_s2 than kill_s1, what do you think?

The problem I had with the kill_s2 is that it is used in the frontend to flush the realigned instruction. mixing it with the valid exception, it did weird things but I agree it seem to be possible to manage it in front end.

core/cva6.sv
cfuguet: Yes, I agree. We can do a different PR to support that

JeanRochCoulon:
@yanicasa @cfuguet  I would prefer to move step by step. As the first goal of the PR was to move the address translation request from icache to frontend, can we merge this modification before submitting another PR to improve even more the design ?

cfuguet:
IMHO, adding a third kill signal to the instruction fetch interface is adding more complexity. And it is kind of still leaving a part of the address translation mechanism in the Icache.

JeanRochCoulon:
Superseeded by #2179

PR #2104, opened 2024-05-13
Running spike first to expedite error detection, especially on ISG

xiaoweish:
Running Spike first greatly aids in detecting any potential issues with ISG (e.g., RISCV-DV) generation, particularly following modifications to ISG that may still lack stability.

### DIFF
--- a/verif/sim/cva6.py
+++ b/verif/sim/cva6.py
@@ -1167,6 +1167,19 @@ def main():
     setup_logging(args.verbose)
     logg = logging.getLogger()
 
+    # Check if --iss argument is provided
+    if args.iss:
+        # Split the string into a list
+        args_list = args.iss.split(',')
+
+        # If 'spike' is in the list, move it to the beginning
+        if 'spike' in args_list:
+            args_list.remove('spike')  # Remove 'spike' from the list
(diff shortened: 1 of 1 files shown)


JeanRochCoulon:
@ASintzoff   can you give us your feedback on the need of such modification ?  thanks

xiaoweish:
Let me explain the problem I'm trying to address:

We have made modifications to RISCV-DV in order to generate stimuli. However, there is a possibility that these modifications have introduced bugs into RISCV-DV.

Currently, when we run simulations with the $DV_SIMULATORS variable settings(=vcs-testharness,spike), they are executed in a specific order, starting with VCS and then followed by Spike. The VCS simulation is particularly time-consuming.

The issue we face is that if there is a bug in RISCV-DV, it may take us a significant amount of time to identify it using this approach.

To mitigate this problem, I propose running Spike first. By doing so, any errors or issues can be detected much earlier in the simulation process. This would save us valuable time in debugging and troubleshooting.

Please let me know your thoughts on this approach.

JeanRochCoulon:
Thanks to your explanation it is clearer. I merge.

PR #2105, opened 2024-05-13
Fixing --iss_timeout Passing and Setting Default Timeout Values for EDA Simulators

xiaoweish:
This PR fixes the issue where --iss_timeout was not being correctly passed to run_c, run_assembly, and run_elf functions. 
The default timeout value is set to 900 seconds (15 minutes) for EDA simulators (e.g., VCS).

Additionally, 
the Spike timeout value is now set to 1/10 of the --iss_timeout value. 
Similarly, the Verilator timeout value is set to --iss_timeout value multiplied by 35, maintaining the original ratio of 35.

### DIFF
--- a/verif/sim/cva6.py
+++ b/verif/sim/cva6.py
@@ -411,7 +411,7 @@ def gcc_compile(test_list, output_dir, isa, mabi, opts, debug_cmd, linker):
 
 
 def run_assembly(asm_test, iss_yaml, isa, target, mabi, gcc_opts, iss_opts, output_dir,
-                 setting_dir, debug_cmd, linker, priv, spike_params, test_name = None):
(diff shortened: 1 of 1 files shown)


JeanRochCoulon:
Hello @xiaoweish 
This will fix a bug we get for years !
Thales servers will not support the increase of the default timeout value, 15 minutes before failing is too long. Could you turn back to the original values ?  This allows on your own to setup it at another value if needed.

xiaoweish:
Hi @JeanRochCoulon , thanks for your suggestion, I've updated the default iss_timeout value as 500s, which is the same as original hardcoded timeout value in run_assembly().

JeanRochCoulon:
Perfect !

Issue #2112, opened 2024-05-15
AXI agent limitations


JeanRochCoulon:
### Is there an existing CVA6 bug for this?

- [X] I have searched the existing bug issues

### Bug Description

During the HPDCache, some limitations have detected in AXI agents, refer to #2097 :
- interleaving and out of order features are enabled/disabled by a single parameter which not relevant knowing some cache support only one of the two features. A second parameter is useful.
- interleaving and out of order features should be define as UV top parameter to be able to configure them in relation to the selected cache
- When integrating the HPDCache, some assert related to ID have been commented. This should be fixed.

@ASintzoff @AnouarZajni

cfuguet:
There other issues related to some illegal bins in coverpoints that are incompatible with the HPDcache. Hereafter you will find a patch commenting out these coverpoints. These bins are actually legal on AXI but they were tagged as illegal in your UVM environment because those ID values were not used by the other cache subsystems.

```diff
--- a/verif/env/uvme/cov/uvme_axi_covg.sv
+++ b/verif/env/uvme/cov/uvme_axi_covg.sv
@@ -72,11 +72,11 @@ covergroup cg_axi_b_channel(string name)
    option.per_instance = 1;
    option.name         = name;
 
-   bid:   coverpoint (item.b_id){
-      bins one   = {[1:3]};
-      illegal_bins ILLEGAL_BINS = {2};
-      ignore_bins  IGN_EXID = {3} iff(!RVA);
-   }
+   //bid:   coverpoint (item.b_id){
+   //   bins one   = {[1:3]};
+   //   illegal_bins ILLEGAL_BINS = {2};
+   //   ignore_bins  IGN_EXID = {3} iff(!RVA);
+   //}
    bresp: coverpoint (item.b_resp){
       bins zero  = {0};
       bins one   = {1};
@@ -84,7 +84,7 @@ covergroup cg_axi_b_channel(string name)
       bins three = {3};
    }   
 
-   b_axi_cross: cross bid, bresp;
+   //b_axi_cross: cross bid, bresp;
 endgroup : cg_axi_b_channel
 
 covergroup cg_axi_ar_channel(string name)
@@ -136,11 +136,11 @@ covergroup cg_axi_r_channel(string name)
    option.per_instance = 1;
    option.name         = name;
 
-   rid: coverpoint (item.r_data_trs[index].r_id) {
-      bins ID[] = {[0:3]};
-      illegal_bins ILLEGAL_BINS = {2};
[...]

Issue #2050, opened 2024-04-17
[BUG] riscv-arch-tests no longer executed in CI


niwis:
### Is there an existing CVA6 bug for this?

- [X] I have searched the existing bug issues

### Bug Description

It appears that since https://github.com/openhwgroup/cva6/commit/75f695f6659e7aed23d45dc26e4d385b4ff84ad5, `riscv-arch-tests` are no longer executed in the CI (job finishes after 2mins without meaningful results or failure, see [log](https://github.com/openhwgroup/cva6/actions/runs/8719432362/job/23919162490#step:6:297))

JeanRochCoulon:
Thanks for your feedback. I assign @MarioOpenHWGroup who took over the GitActions CI.

cainria:
I think this issue can be closed

PR #2114, opened 2024-05-15
Add in Github's CI a 64-bit configuration of the CVA6 using the HPDcache and restore WB cache test

cfuguet:
This PR adds to the Github's CI a tests with the cv64a6_imafdc_sv39_hpdcache configuration.

This configuration uses 64-bit variant of the CVA6 with "deep" structures related to the data cache:
- Up to 8 inflight loads to the cache (and thus up to 8 miss requests to the memory)
- Up to 8 inflight write requests to the memory

It also restores the testing of the WB cache into the CI.

FIXME: In addition, this PR comments out some coverpoints in the AXI agent because of illegal bins that are not compatible with the HPDcache. These coverpoints expect that the cache uses a very small subset of IDs in AXI transactions, however the HPDcache is able to use a larger subset as it can issue multiple inflight requests to the memory (both reads and writes).

I pushed a first proposal for the Github's CI. It runs the following configurations:

    cv64a6_imafdc_sv39_hpdcache
    cv64a6_imafdc_sv39_wb
    cv64a6_imafdc_sv39
    cv32a65x

However, the number of tests is limited. The selected tests allow to test arithmetic, memory and branch/jump instructions and a complete but very simple application (hello_world). In addition, the riscv-arch-test suite is executed with the cv64a6_imafdc_sv39_hpdcache and cv32a65x configurations. In the case of 64 bits configurations, I selected the unitary tests that enable the MMU.

The tests for a given configuration are run sequentially but all configurations are run in parallel to reduce the overall latency.

For 64 bits configurations:

    riscv-arch-test (only in the cv64a6_imafdc_sv39_hpdcache configuration)
    rv64ui-v-add
    rv64ui-v-ld
    rv64ui-v-sd
    rv64ui-v-beq
    rv64ui-v-jal
    hello_world.c

For 32 bits configurations:

    riscv-arch-test
    rv32ui-p-add
    rv32ui-p-lw
    rv32ui-p-sw
    rv32ui-p-beq
    rv32ui-p-jal
    hello_world.c

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,15 +49,20 @@ jobs:
           ci/setup.sh
 
 
-  execute-riscv-tests:
-    name: execute-riscv-tests
+  execute-riscv64-tests:
+    name: execute-riscv64-tests
     runs-on: ubuntu-latest
     env:
       SPIKE_TANDEM: 1
     strategy:
       matrix:
-        testcase: [ dv-riscv-arch-test , smoke-tests ]
-        target:   [ veri-testharness ]
+        testcase:  [ cv64a6_imafdc_tests ]
+        config:    [ cv64a6_imafdc_sv39_hpdcache, cv64a6_imafdc_sv39_wb, cv64a6_imafdc_sv39 ]
+        simulator: [ veri-testharness ]
+        include:
+          - testcase: dv-riscv-arch-test
+            config: cv64a6_imafdc_sv39_hpdcache
+            simulator: veri-testharness
     needs:
       build-riscv-tests
     steps:
@@ -98,11 +103,71 @@ jobs:
           export RISCV=$(pwd)/tools/riscv-toolchain/
           source ci/install-prereq.sh
           source verif/sim/setup-env.sh
-          DV_SIMULATORS=${{matrix.target}} bash verif/regress/${{matrix.testcase}}.sh
+          DV_SIMULATORS=${{matrix.simulator}},spike DV_TARGET=${{matrix.config}} bash verif/regress/${{matrix.testcase}}.sh
 
     - name: Upload Lint Report to Github
       uses: actions/upload-artifact@v4
       with:
-        name: ${{matrix.target}}.${{matrix.testcase}}
+        name: ${{matrix.simulator}}.${{matrix.testcase}}.${{matrix.config}}
+        path: 'verif/sim/out*'
+        retention-days: 10
+
+
+  execute-riscv32-tests:
+    name: execute-riscv32-tests
+    runs-on: ubuntu-latest
+    env:
+      SPIKE_TANDEM: 1
+    strategy:
+      matrix:
+        testcase:  [ dv-riscv-arch-test, cv32a6_tests ]
+        config:    [ cv32a65x ]
+        simulator: [ veri-testharness ]
+    needs:
+      build-riscv-tests
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        submodules: recursive
+
+    - name: Cache toolchain
+      id: cache-toolchain
+      uses: actions/cache@v3
+      env:
+          cache-name: cache-toolchain
+      with:
+          path: tools/riscv-toolchain/
+          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('ci/install-toolchain.sh') }}
+
+    - name: Cache verilator
+      id: cache-verilator
+      uses: actions/cache@v3
+      env:
+          cache-name: cache-verilator
+      with:
+          path: tools/verilator/
+          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('verif/regress/install-verilator.sh') }}
+
+    - name: Cache Spike
+      id: cache-spike
+      uses: actions/cache@v3
+      env:
+          cache-name: cache-spike
+      with:
+          path: tools/spike/
+          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('verif/regress/install-spike.sh', 'verif/core-v-verif/vendor/riscv/riscv-isa-sim/') }}
+
+    - name: Run Tests
+      run: |
+          set -x
+          export RISCV=$(pwd)/tools/riscv-toolchain/
+          source ci/install-prereq.sh
+          source verif/sim/setup-env.sh
+          DV_SIMULATORS=${{matrix.simulator}},spike DV_TARGET=${{matrix.config}} bash verif/regress/${{matrix.testcase}}.sh
+
+    - name: Upload Lint Report to Github
+      uses: actions/upload-artifact@v4
+      with:
+        name: ${{matrix.simulator}}.${{matrix.testcase}}.${{matrix.config}}
         path: 'verif/sim/out*'
         retention-days: 10
--- a/verif/regress/cv32a6_tests.sh
+++ b/verif/regress/cv32a6_tests.sh
@@ -48,11 +48,17 @@ cd verif/sim/
 errors=0
 
 # 32-bit configurations without MMU
-python3 cva6.py --testlist=../tests/testlist_riscv-tests-${DV_TARGET}-p.yaml --test rv32ui-p-add --iss_yaml cva6.yaml --target ${DV_TARGET} --iss=$DV_SIMULATORS $DV_OPTS
-[[ $? > 0 ]] && ((errors++))
-
-python3 cva6.py --testlist=../tests/testlist_riscv-tests-${DV_TARGET}-p.yaml --test rv32ui-p-lw --iss_yaml cva6.yaml --target ${DV_TARGET} --iss=$DV_SIMULATORS $DV_OPTS
-[[ $? > 0 ]] && ((errors++))
+riscv_tests_list=(
+  rv32ui-p-add
+  rv32ui-p-lw
+  rv32ui-p-sw
+  rv32ui-p-beq
+  rv32ui-p-jal
+)
+for t in ${riscv_tests_list[@]} ; do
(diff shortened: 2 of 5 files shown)


JeanRochCoulon:
Yes, having HPDCache in 64 bit config is a most.

Each time we add a new config to be tested in smok-tests.sh, it more or less 5 minutes of execution time increase. We do not want it. Can you reuse the cv64a6_imafdc_sv39 config, I mean replace the wt into hpd in this config ?

cfuguet:
> Yes, having HPDCache in 64 bit config is a most.
> 
> Each time we add a new config to be tested in smok-tests.sh, it more or less 5 minutes of execution time increase. We do not want it. Can you reuse the cv64a6_imafdc_sv39 config, I mean replace the wt into hpd in this config ?

Yes, I can of course. If no one will be upset because it is expecting something else, I can do it :)

JeanRochCoulon:
Could you post a message in MM to ask whether one organization will be impacted ?   But I think nobody use it for critical purpose.

Jbalkind:
Unless we have agreement that wtdcache is being deprecated, I'm not sure that I like this option a whole lot. You have five years of users who have been using wtdcache by default that you're potentially making a change for.

cfuguet:
Yes, I can understand your point, but at the same time, the HPDcache is actually adding functionality and performance transparently. With roughly the same area.

cfuguet:
But well, it is a slightly biased view... :)

JeanRochCoulon:
In Thales CI, we will not test two 64bits configurations.

To not impact the users, the best (as proposed previously) could be to create a new 64 bits config with hpdcache (64bits_hpdcache) and replace in smoke-tests the first 64bits config by the 64bits_hpdcache config.

verif/env/uvme/cov/uvme_axi_covg.sv
```diff
@@ -72,19 +72,19 @@ covergroup cg_axi_b_channel(string name)
    option.per_instance = 1;
    option.name         = name;
```
JeanRochCoulon, line 74: @cfuguet   can you elaborate on this modification, it would be great to add it to the #2112 Github issue. I would be please to get the @AEzzejjari comment on it. Better is to condition the lines as seen in the image below:
image

cfuguet: Hello @JeanRochCoulon,

I comment about the below modifications: I commented out coverpoints that are related to IDs. These coverpoints define illegal_bins which are actually not illegal when using the HPDcache. There were illegal for the other caches as they use 1 or 2 IDs at most, but the HPDcache can actually use much more than that (depending on the number of MSHRs, and write buffer entries).

Regarding the suggestion about using an if statement, the problem is that I do not think they are supported within a covergroup definition. I will try.

cfuguet: Yes, I confirm that if statements are not supported within covergroup definitions

JeanRochCoulon: In that case, the best could be to comment them and add a comment to keep track of it (a kind of TODO) in the code.

JeanRochCoulon: @AEzzejjari What is you position ?

cfuguet:
> In Thales CI, we will not test two 64bits configurations.
> 
> To not impact the users, the best (as proposed previously) could be to create a new 64 bits config with hpdcache (64bits_hpdcache) and replace in smoke-tests the first 64bits config by the 64bits_hpdcache config.

Hypothetically, if we choose this solution:
- shall we change the default configuration of the top Makefile from  cv64a6_imafdc_sv39 to cv64a6_imafdc_sv39_hpdcache ?
- Shall we change the configuration in the Github CI's from cv64a6_imafdc_sv39 to cv64a6_imafdc_sv39_hpdcache ?

verif/env/uvme/cov/uvme_axi_covg.sv
AEzzejjari: hello,

We can comment out assertions or use if statements to deactivate these assertions, but the problem here is that every time we want to run a test with a different configuration, we have to uncomment these assertions.

We can deactivate assertions more efficiently by utilizing a predefined SystemVerilog function that can be called from the testbench. However, we will still encounter the same configuration issue.

Another solution is to modify the assertions to be configurable so that they could work correctly with both configurations. We just need the CVA6 configuration to do so, and we have it in the testbench.

Moschn:
I have no skin in the game but I would argue to at least keep the WTCache in the Thales CI until the Github CI is fixed (see #2050). I understand that you do not want to test a config not useful to you in your CI. However, we have had the WBcache in the Github CI for years (somehow it is no longer there now, but IMHO the WBcache config should be tested in some CI). Don't get me wrong, the Thales CI is amazing, but we exclusively rely on it for PRs and if you want to remove configurations from it, I would strongly suggest to sanity-check these at least in the Github CI. 

In my personal opinion, removing tests/configurations from the CI is nonsensical. The only valid reason to remove a config from the CI is to deprecate it which is currently not the case.

cfuguet:
Yes, I agree with your point.

I'm asking myself if one of the (technical) issues currently is that we are running the same scripts (smoke-tests) in both the Github CI's and Thales CI's. But in Thales CI's there are additional tests running and even those smoke-tests are run multiple times because they are run in different test environments (Verilator testbench and UVM testbench using VCS).

May be we should have two different test scripts. One test script for the Github CI's that runs some tests with the 3 different supported caches (because as @Moschn suggests there is no official deprecation of any of them), and another in the Thales CI's with the configurations relevant to Thales.

cfuguet:
Do you want me to propose a script running some tests on the 3 caches that we can use in the Github's CI ?

JeanRochCoulon:
For information, we plans to test (only) the 2 Thales configs in Thales CI: one is 32bit, one is 64bits both with HPDCache. FPGA and ASIC synthesis will be done. As consequence the smoke-tests.sh will only test these two configurations. That is why testing the three caches in Github CI is useful. Feel free.

cfuguet:
> For information, we plans to test (only) the 2 Thales configs in Thales CI: one is 32bit, one is 64bits both with HPDCache. FPGA and ASIC synthesis will be done. As consequence the smoke-tests.sh will only test these two configurations. That is why testing the three caches in Github CI is useful. Feel free.

Ok great ! Thank you @JeanRochCoulon for the information (the HPDcache is very happy by the way).

In parallel, I will propose a script for the Github's CI testing all caches with a limited number of tests to keep the latency low.

cfuguet:
I pushed a first proposal for the Github's CI. It runs the following configurations:

- cv64a6_imafdc_sv39_hpdcache
- cv64a6_imafdc_sv39_wb
- cv64a6_imafdc_sv39
- cv32a65x

However, the number of tests is limited. The selected tests allows to tests 1 arithmetic instruction (add), 1 memory instruction (ld) and a complete but very simple application (hello_world). In the case of 64 bits configurations, we use the tests that enable the MMU.

For 64 bits configurations:
- rv64ui-v-add
- rv64ui-v-ld
- hello_world.c

For 32 bits configurations:
- rv32ui-p-add
- rv32ui-p-lw
- hello_world.c

What do you think @JeanRochCoulon, @Moschn ?

cfuguet:
:exclamation: There is a problem regarding the Spike version. All tests are passing but actually, tests are not being run because of a check in the Spike version by the cva6.py script. I saw the same issue in other currently active pull requests (so it is not related to the modifications in this PR).

@JeanRochCoulon do you have an idea about this issue ?

cfuguet:
Ok, I found how to flush the Spike cache. I rerun the jobs, and lets see.

cfuguet:
Well this is working now. It took a total of 32 minutes to run which is the latency of the riscv-arch-tests. The latency of all other tests is completely hidden by the latter (because they are all run in parallel).

This means that we can still add some unitary tests for the different configurations in order to balance a little bit the latencies.

Let me know what do you think.

:exclamation: By the way, I notice something that we need to be careful of. When we do multiple calls to the cva6.py script in a given script, we need to check each time for the exit status, and at the end of the script exit with an error if some of the calls failed. Otherwise, only the status of the last command in the script is returned. This means that, if you count on the return status of the script to determine in the CI if the tests failed or not, you may overlook some errors. This was what it was happening with the Spike version error. The CI was passing but the tests were not actually run. Hence, I added in the tests scripts the check of the return status. Of course, other solution is to add an additional step and parse all logs, I think this is the strategy you do in Thales CI's.

verif/env/uvme/cov/uvme_axi_covg.sv
cfuguet: Yes, for sure commenting out lines is not the solution. This is just temporary in order to be able to integrate the HPDcache in the tests. We need to find a solution as you said, to make this coverage more parametric. In the case of IDs, it should depend on the kind of cache subsystem, but also should consider parameters such as the AXI ID width parameter.

JeanRochCoulon:
Hello @cfuguet 
@AEzzejjari conditioned the assertions and coverage in AXI agent, that's why you do not need to comment it into your PR. Please rebase it to give me the opportunity to merge it.

cfuguet:
> Hello @cfuguet @AEzzejjari conditioned the assertions and coverage in AXI agent, that's why you do not need to comment it into your PR. Please rebase it to give me the opportunity to merge it.

Ok ! Now it is done. Thanks !

JeanRochCoulon:
I approve this PR. But I would be pleased to get the @MarioOpenHWGroup, who recently setup the Github Actions CI,  feedback before merging. Mario are you ok ?

MarioOpenHWGroup (approved):
LGTM :smile:

Moschn:
Sorry, I was offline the past few days. Thanks @cfuguet for the effort!

PR #2117, opened 2024-05-15
Mmu design document

AngelaGonzalezMarino:
Update of design document related to MMU Unification


JeanRochCoulon:
This documentation is very insteresting thanks @AngelaGonzalezMarino 
@slgth Do you have time to have a look and approve it ?

fatimasaleem:
Sure @jquevremont, since @YazanHussnain-10x created these figures and contributed the document, I will ask him to review this.

YazanHussnain-10x:
Great work, @AngelaGonzalezMarino. I have a few suggestions:

1. In section _"Flushing TLB Entries"_, HFENCE cases need to mentioned explicitly.
2. Please correct me if I'm mistaken, but in Figure 21, shouldn't the VPN be left-shifted by the Log2 of the PTE size rather than the PT_LEVELS? The PTE size can only be 4 bytes in the case of Sv32 and 8 bytes in other page-based virtual memory schemes, where PT_LEVELS, as mentioned in the configuration parameter table, can be any value greater than 1.
3. In the case of the Hypervisor, the state machine in PTW becomes nested. Transitioning of the internal state machine also needs to be mentioned in the diagram, or alternatively, you can add a separate one specifically for the internal state machine. Something like this
ptw_std
5. Can we add diagram for nested translation? something like this.
nested_trans

AngelaGonzalezMarino:
> Great work, @AngelaGonzalezMarino. I have a few suggestions:
> 
> 1. In section _"Flushing TLB Entries"_, HFENCE cases need to mentioned explicitly.
> 2. Please correct me if I'm mistaken, but in Figure 21, shouldn't the VPN be left-shifted by the Log2 of the PTE size rather than the PT_LEVELS? The PTE size can only be 4 bytes in the case of Sv32 and 8 bytes in other page-based virtual memory schemes, where PT_LEVELS, as mentioned in the configuration parameter table, can be any value greater than 1.
> 3. In the case of the Hypervisor, the state machine in PTW becomes nested. Transitioning of the internal state machine also needs to be mentioned in the diagram, or alternatively, you can add a separate one specifically for the internal state machine. Something like this
>    ptw_std
> 4. Can we add diagram for nested translation? something like this.
>    nested_trans

Hi @YazanHussnain-10x, thanks for the review!
About 1. You are right, I've added the hfence cases. Please have a look.
About 2. It is actually the same. PT_LEVELS is 2 for sv32 and 3 for sv39 and sv39x4, which already corresponds to Log2 of the PTE size (which is the same as XLEN, 32 for sv32 and 64 for sv39 and sv39x4). I updated the possible values of PT_LEVELS to avoid confusion
About 3. I extended the FSM of PTW. Please have a look too.
About 4. I think @ninolomata has done a nice graphical representation of this. Bruno, would you mind adding your diagram for nested translation to this documentation? 
Thanks everyone!!

YazanHussnain-10x (approved):
Now, it looks good to me. Only the diagram of nested translation is remaining.

AngelaGonzalezMarino:
> Now, it looks good to me. Only the diagram of nested translation is remaining.

Thanks! The diagram from @ninolomata is now added to this documentation

JeanRochCoulon:
Thanks to all of you !

PR #2123, opened 2024-05-16
Invalid pte reserved

AngelaGonzalezMarino:
Add reserved bits to PTE structure be used in sv39 and sv39x4 only

Check that they are 0, otherwise the PTE is considered invalid


AngelaGonzalezMarino:
@ninolomata can you please check if this is the right fix?
Thanks!

ninolomata (approved):

PR #2124, opened 2024-05-16
fix tval in mmu

AngelaGonzalezMarino:
Use vaddr for tval instead of paddr when translation is enabled


AngelaGonzalezMarino:
@ninolomata can you confirm that this is the correct tval?
Thanks!

JeanRochCoulon:
@AngelaGonzalezMarino Can I merge ?

PR #2127, opened 2024-05-17
Functional coverage report in CI

AyoubJalali:
Hello, this MR related to display functional coverage results for ISA, CSRs and TRAPs, another related MR while be open to report the results in dashbord


cathales:
This PR makes the CI report build functional coverage reports **overwriting** the line coverage instead of adding a new report or metric.

This PR has been merged on 2024-05-17. The work crashed on 2024-05-19. Then it completed successfully on 2024-05-26.

As the latter inserted a report with the wrong attributes, and the dashboard crashes since, as it gets a functional coverage table instead of a line coverage table.

PR #2130, opened 2024-05-20
CSR REG PREDICTOR : Skip getting package on traps

AyoubJalali:
After some modification on the rvfi agent, now the package is sent when trap = 0 only !!


AyoubJalali:
@JeanRochCoulon could you merge this please if it okey !

PR #2149, opened 2024-05-24
New target with MMU: cv64a6_mmu

slgth:
This merge request introduces a new configuration for the CVA6, specifically including an MMU. The following changes have been made:

* Added configuration files for the new target
* Updated the smoke-test.sh to include the new target
* Modified some verification scripts to enable the simulation of the new target


JeanRochCoulon:
Well done @slgth !

PR #2150, opened 2024-05-24
ADD RISC-V CONFIG SOFTWARE

AbdessamiiOukalrazqou:
This software takes a Risc -V Config Yaml description of CSR Registers, and generates ReStructuredText documents and Mardown Documents for CSR Registers and ISA Extensions. In the example/tb directory there is an example of how to generate the packages


JeanRochCoulon:
@AbdessamiiOukalrazqou   can you sign the eclipsefdn/eca agreement ?

ASintzoff:
Could you format the Python files with black?

JeanRochCoulon:
As we are in open source, the files need to be with header indicating it, also if they are generated, can you add headers ?

AbdessamiiOukalrazqou:
we will take care of reformatting files with black and Headers

PR #2151, opened 2024-05-27
Updated tb_wb_dcache and adapted it for 32-bits

cyprienh:
This PR provides an updated testbench for the WB cache which works for 64-bits and 32-bits (PR for 32-bits WB cache coming soon). The tests run with the current 64-bits version of the WB cache and allowed me to catch the bug from PR #2142.


zarubaf (approved):
Nice, thanks a lot!

PR #2154, opened 2024-05-27
add gen from riscv config software

AbdessamiiOukalrazqou:
This software takes a`Risc -V Config Yaml description of CSR Registers, and generates ReStructuredText documents and Mardown Documents for CSR Registers and ISA Extensions. In the example/tb directory there is an example of how to generate the packages


JeanRochCoulon:
@AbdessamiiOukalrazqou   Can you sign the Eclipse agreement, needed to contribute to the CVA6 repository.

Issue #2006, opened 2024-04-05
[BUG] AXI bus trouble with write-back cache


nlthales:
### Is there an existing CVA6 bug for this?

- [X] I have searched the existing bug issues

### Bug Description

Hi,
I got a SoC project using CVA6 cores, which is running well with its default configuration - cache in write-through mode (config_pkg::WT). I gave a try in write-back mode but without any success. Looking at the AXI bus, I can see some difference while doing the first write access, in an uncachable memory area. Discrepencies are shown hereafter:
![axi_cva6_wt_vs_wb](https://github.com/openhwgroup/cva6/assets/166120106/1ff00a0d-e224-4351-8890-c6556b2cda3e)
The AXI subsystem acknowledges the access in both situations, but in WB the access seems lost and does not result in an actual access to the target peripheral.

jquevremont:
Hi @niwis. I guess the WB cache is used at ETH. Do you think someone can look at this?

niwis:
Hello @nlthales, the differences in the AXI IDs should not have any functional implications (as long as they are consistent). Regarding the write strobe alignment, which XLEN are you using for CVA6? Afaik the WB cache has only been tested with 64-bit so far.

nlthales:
> Hello @nlthales, the differences in the AXI IDs should not have any functional implications (as long as they are consistent). Regarding the write strobe alignment, which XLEN are you using for CVA6? Afaik the WB cache has only been tested with 64-bit so far.

hello, thanks for your feedback. I do confirm that I use 32-bit configuration.

niwis:
There are some known incomatibilities of the WB cache and 32-bit XLEN (this is probably one of them). I think @cyprienh is tackling them. One point to start would be to port the [modular cache testbench](https://github.com/openhwgroup/cva6/tree/master/corev_apu/tb/tb_wb_dcache) to 32-bit and fix any issues that arise.

cyprienh:
Hi, I have indeed been working on adapting the WB cache for 32-bit ! I will try adapting the testbench to check that my changes do not break anything else and cover all the use-cases. If it all works, I will submit my changes.

cainria:
Can this issue be closed by #2170?

PR #2170, opened 2024-05-30
32 bits WB cache

cyprienh:
This PR adapts the WB cache so that it works with the 32 bits versions of CVA6. It was tested using the updated TB from PR #2151 and should also solve the issue #2006. It works by aligning the data to and from the AXI bus on the 64 bits of the bus (lower or upper 32 bits depending on the address).


JeanRochCoulon:
To solve the format errors, the CONTRIBUTING.md gives the VERIBLE command to be executed.

PR #2179, opened 2024-05-31
OBI protocol between Frontend to icache

yanicasa:
1. Address translation request moved from icache to frontend
2. OBI protocol between icache and frontend

work in progess, not fully OBI compliant


JeanRochCoulon:
In place of "do not merge", it is better to set the DRAFT mode !

JeanRochCoulon:
@yanicasa please set your PR in DRAFT mode to prevent from "merging".

PR #2181, opened 2024-05-31
Disable MISA WE in Reference Model

MarioOpenHWGroup:
- Disabling MISA WE as cva6 does not allow writes
- Add YAML report for all the tests


JeanRochCoulon:
@zchamski Can you approve ?

AyoubJalali:
I don't know why but I need to give the cva6.py the misa_we_enabled to get spike work !!

JeanRochCoulon:
MISA is read only. I do not understand why in Spike MISA need to be write-able !?
image

zchamski:
> I don't know why but I need to give the cva6.py the misa_we_enabled to get spike work !!

The `misa_we_enable` parameter (with value 'true') *enables* the control of the writability of MISA via the parameter `misa_we`. 
If `misa_we_enable` is left at `false`, the default behavior of Spike remains in effect, which is to keep MISA extension bits `MAFDQCHV` writable (https://github.com/openhwgroup/core-v-verif/blob/master/vendor/riscv/riscv-isa-sim/riscv/csrs.cc#L613)

zchamski (approved):
LGTM

Issue #1829, opened 2024-02-13
[TASK] Cadence simulator support


CoralieAllioux:
### Is there an existing CVA6 task for this?

- [X] I have searched the existing task issues

### Task Description

Support additional simulator: Cadence xcelium. The support concerns uvm and testharness. It needs to be transparent for VCS users and adapt to the current flow.

### Required Changes

Two new DV_SIMULATOR options to be add: xrun-uvm and xrun-testharness. It implies:
- UVM/RTL fixes in cva6 and core-v-verif
- Adapt cva6/verif/sim/Makefile
- Adapt cva6/verif/sim/cva6.yaml
- Adapt cva6/verif/sim/cva6-simulator.yaml
- Adapt cva6/verif/env/corev-dv/simulator.yaml

Probably additional files would be impacted and could be forgotten.

### Current Status

In progress
xrun-uvm almost done with riscv-dv support

### Risks

Fixes to run a new simulator might not be straight forward and challenge the current file hierarchy or methodology.

### Description of Done

Complete vcs flow supported by Cadence Xcelium, while using DV_SIMULATOR xrun-uvm and xrun-testharness.

cainria:
Hi 👋 can we close this issue as completed?

PR #2210, opened 2024-06-06
[Xcelium flow] corev dv yaml

CoralieAllioux:
This PR aims at adding xrun rule in `verif/env/corev-dv/simulator.yaml`, which is a step to support Cadence xcelium simulator while using riscv-dv instruction generator.
It contributes to task  #1829

### DIFF
--- a/verif/env/corev-dv/simulator.yaml
+++ b/verif/env/corev-dv/simulator.yaml
@@ -28,7 +28,7 @@
               -64
               +incdir+<setting>
               +incdir+<user_extension>
-              -f <cwd>/dv/files.f
+              -f <cwd>/../env/corev-dv/cva6-files.f
               +define+UVM_REGEX_NO_DPI
               -timescale 1ns/10ps
 	      -status
--- a/verif/env/corev-dv/simulator.yaml
+++ b/verif/env/corev-dv/simulator.yaml
@@ -17,6 +17,36 @@
 
 ###### This file is based on <cwd>/cva6-simuator.yaml
 
+- tool: xrun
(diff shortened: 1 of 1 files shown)


verif/env/corev-dv/simulator.yaml
```diff
@@ -17,6 +17,36 @@
...
+              -timescale 1ns/10ps
+	      -status
+	      -access +rwc
+              -log <out>/compile.log  <cmp_opts> <cov_opts> "
+    cov_opts: >
+#              -coverage all <out>
+#	      -nowarn COVDEF
+  
+#      -cm_dir <out>/test.vdb
+#              -f <cwd>/../env/corev-dv/cva6_instr_base_test.sv
```
JeanRochCoulon, line 42: Why keeping these lines ?

```suggestion

```

verif/env/corev-dv/simulator.yaml
```diff
@@ -17,6 +17,36 @@
...
+  sim:
+    cmd: >
+      xrun -R -messages -licqueue +gen="true" <sim_opts> -svseed <seed>  <cov_opts> 
+      
+    cov_opts: >
+#      -cm_dir <out>/test.vdb -cm_log /dev/null -cm_name test_<seed>_<test_id>
```
JeanRochCoulon, line 48: Same

```suggestion
```

verif/env/corev-dv/simulator.yaml
CoralieAllioux: You're right, it'll be removed

PR #2232, opened 2024-06-11
Fix PMPCFG number

JeanRochCoulon:
From 8 to 4 PMPCFG, from which 2 are read-only zero


JeanRochCoulon:
@ASintzoff can you approve (post-morten) ?

PR #2234, opened 2024-06-11
When NrPMPEntries == 8, set 2 upper PMPCfg and 8 upper CfgAddress as read-only CSRs

JeanRochCoulon:
Related logic is not removed, this will be addressed in following PR

### DIFF
--- a/core/csr_regfile.sv
+++ b/core/csr_regfile.sv
@@ -764,9 +764,12 @@ module csr_regfile
         riscv::CSR_PMPCFG1:
         if (CVA6Cfg.XLEN == 32) csr_rdata = pmpcfg_q[7:4];
         else read_access_exception = 1'b1;
-        riscv::CSR_PMPCFG2: csr_rdata = pmpcfg_q[8+:CVA6Cfg.XLEN/8];
+        riscv::CSR_PMPCFG2:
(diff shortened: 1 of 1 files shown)


Moschn:
Just FYI: the write logic is [here](https://github.com/openhwgroup/cva6/blob/91871d97f350dc3fd808ca4ebb893b4be3bcecac/core/csr_regfile.sv#L2514). I believe the current state should already achieve what you want to achieve in this PR. Maybe your change will save some area, but I assume a good compiler will optimize this away anyways.

Also the proposed changes are hardcoded for NR_PMP_ENTRIES=8. What happens if one uses 12 PMP entries? Or 7?

JeanRochCoulon:
THnaks you @Moschn for your feedback. I set the PR in draft and will fine tune the PR.

JeanRochCoulon:
ok, after deeper analysis of the code, the RTL seems functional as it is. @AyoubJalali , what is your feedback, have you seen some misfunctionality on PMPCFG or PMPADDR ?

AyoubJalali:
> ok, after deeper analysis of the code, the RTL seems functional as it is. @AyoubJalali , what is your feedback, have you seen some misfunctionality on PMPCFG or PMPADDR ?

I'll check this !!

PR #2237, opened 2024-06-12
interconnect alpha2

yanicasa:
Fetch:
- fix icache combi loop

LSU:
- Prepare OBI structure


JeanRochCoulon:
The synthesis, the FPGA Linux boot are failed. But as it is not the first priority, I merge as it is.

PR #2238, opened 2024-06-12
feat: extend cva6_hpdcache_if_adapter (support for hpdcache_icache as well)

takeshiho0531:
- The parts highlighted in light pink are the ones we want to implement in this PR.

IMG_75979DCD40C8-1

- extend [cva6_hpdcache_if_adapter](https://github.com/openhwgroup/cva6/blob/ade4c85e139eb4e2b743069e6d4f696d4a258e92/core/cache_subsystem/cva6_hpdcache_if_adapter.sv#L12C8-L12C32) so that it can also convert input request/response to the appropriate type for hpdcache when using hpdcache as icache. (ex. input type:icache_req_t → output type: hpdcache_req_t)

### DIFF
--- a/core/cache_subsystem/cva6_hpdcache_if_adapter.sv
+++ b/core/cache_subsystem/cva6_hpdcache_if_adapter.sv
@@ -22,7 +22,8 @@ module cva6_hpdcache_if_adapter
     parameter type hpdcache_rsp_t = logic,
     parameter type dcache_req_i_t = logic,
     parameter type dcache_req_o_t = logic,
-    parameter bit is_load_port = 1'b1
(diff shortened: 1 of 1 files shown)


yanicasa:
I am currently working on the redefinition of the interfaces between the cache and the front end and the LSU,
I think this is a subject related to yours, I have a dev branch open in the CVA6 repo where I simplified the icache kill mechanism and moved the fetch address translation request from icache to frontend.

For my part I am looking for someone who could adapt the caches to these new interfaces that I am developing in the front end and the LSU

Branch:
https://github.com/openhwgroup/cva6/tree/feature/interconnect

@cfuguet @JeanRochCoulon

PR #2246, opened 2024-06-12
Remove cv32a6_embedded configuration

JeanRochCoulon:
This reduces the smoke-test duration.
Embedded_config_pkg.sv has been moved to Embedded_config_pkg_deprecated.sv, and will be removed from repository into 2 months


cathales:
[Line 88 of the coremark regress script](https://github.com/openhwgroup/cva6/pull/2246/files#diff-0f94b605d882c4b091c558c9a42640e248f23df21a35f6afb45c84efcbba1e3eR88) should be updated as the `CVA6ConfigNrLoadPipeRegs` parameter does not exist anymore in the new `cv32a65x` configuration.

I have just got a look at the `cv32a65x` configuration and it seems that `NrLoadPipeRegs` is already 0 in this configuration, so the whole `+CVA6ConfigNrLoadPipeRegs=0` modifier should be removed.

Actually we *could* just stop using hwconfig in this job, but I think **we should keep using it**, even without modifiers, so that it is easier to perform design space exploration and measure the performance impact of configuration.

JeanRochCoulon:
@cathales  thnaks (always) for your reviews.
Can you plug your feedback in Github issue ?   We will handle it by another PR

cathales (approved):
LGTM if CI is green

PR #2251, opened 2024-06-13
Prepare for LLVM

math-gout:
This PR prepares the ground to use LLVM instead of GCC.


JeanRochCoulon:
@math-gout Can you rebase ?

PR #2255, opened 2024-06-13
Execute lint job when RTL is modified

JeanRochCoulon:
Lint check is only needed when RTL is modified


JeanRochCoulon:
@ASintzoff I moved the lint job to reduce the licence needs. Are you agree ?

JeanRochCoulon:
@ASintzoff I updated the PR to integrate your feedback

Issue #1447, opened 2023-09-19
[TASK] Extract PMP from MMU module


JeanRochCoulon:
### Is there an existing CVA6 task for this?

- [x] I have searched the existing task issues

### Task Description

Today PMP dedicated to data is inside MMU module. Removing MMU removes PMP, which is not wanted. 

### Required Changes

The objective is to instantiate data PMP outside MMU. Duplication of code must be avoided.

### Current Status

TODO

### Risks

Be careful to well reconnect the exceptions information.

### Prerequisites

none

### KPI (KEY Performance Indicators)

none

### Description of Done

PR merged in cva6 master. PMP functional for embedded configuration.

JeanRochCoulon:
@mike, I tried to assign @CoralieAllioux to this task, but Bosch contributors are not included in the Github cva6 project. Could you include Coralie, Nicolas, Oliver and Jean-Christian ?

JeanRochCoulon:
@CoralieAllioux, the different MMUs have been unified last week. The PMP extraction is possible. Can you update us about your plan ?

CoralieAllioux:
Hi @JeanRochCoulon, it is still plan to contribute to the PMP extraction. Since the unified MMU is on the master branch, the design will be start (updated) ASAP. However, we miss to define some tests to provide with the RTL. We are working on it.
Does it answer your question?

JeanRochCoulon:
Good !

JeanRochCoulon:
Done by #2528

PR #2260, opened 2024-06-14
[Pmp] Parameter for PMP without MMU

CoralieAllioux:
First step for the PMP standalone: adding a dedicated parameter in config package and all configs.

It contributes to https://github.com/openhwgroup/cva6/issues/1447

### DIFF
--- a/core/include/build_config_pkg.sv
+++ b/core/include/build_config_pkg.sv
@@ -160,6 +160,8 @@ package build_config_pkg;
     cfg.VpnLen = VpnLen;
     cfg.PtLevels = PtLevels;
 
+    cfg.PmpPresent = CVA6Cfg.PmpPresent;
+
     return cfg;
   endfunction
 
--- a/core/include/config_pkg.sv
+++ b/core/include/config_pkg.sv
@@ -196,6 +196,8 @@ package config_pkg;
     bit unsigned                 UseSharedTlb;
     // MMU depth of shared TLB
     int unsigned                 SharedTlbDepth;
+    /// Enable PMP without MMU
+    bit                          PmpPresent;
   } cva6_user_cfg_t;
 
   typedef struct packed {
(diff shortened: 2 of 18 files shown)


JeanRochCoulon:
Hello @CoralieAllioux 
The new parameter name tells the PMP instruction and data will be present when PmpPresent = 1. For instance:
- When MMU is present and PmpPresent = 0, we understand the PMP is not present
- When PmpPresent = 0, we understand the PMP instruction is not present whatever the MMU.
If it is the case, it would be great, do you confirm?  Otherwise, it could be confusing.

CoralieAllioux:
Hello @JeanRochCoulon,
About the parameter, it only instantiates the PMP when the MMU is **not** present. Therefore the parameter PmpPresent is ignore in that case.
I could then rename it to PmpStandalonePresent or something similar to be more explicit. Would it be better for you?

JeanRochCoulon:
This new parameter allows to disable the PMP _data_ when MMU not present and PmpPresent=0. To me this configuration does not need to be supported: this configuration is not compliant to RISC-V and I do not know a CVA6 user pleased to use it so far (maybe youself ?).

CoralieAllioux:
I am not sure to understand your last comment. _PmpStandalonePresent_ was a suggestion to **rename** _PmpPresent_ if not explicit enough. This is not a new parameter

Here the possible configuration we're proposing:
- **MmuPresent=0 and PmpPresent=0:** nothing changes
- **MmuPresent=1 and PmpPresent=0:** nothing changes
- **MmuPresent=1 and PmpPresent=1:** nothing changes
- **MmuPresent=0 and PmpPresent=1:** here a pmp data and a pmp instruction are inserted, working without mmu

JeanRochCoulon:
RISC-V specification tells that PMP is optional. But in CVA6, PMP regions are always 16.
To be compliant PMP data and instruction shall be supported. It is not the case Today when MMU is not present.
Your modification will make CVA6 compliant to RISCV when MMU is not present.

CoralieAllioux:
Oh, so you're telling MmuPresent=0 is sufficient? If so, PmpPresent is not necessary... and the PR is not needed anymore

CoralieAllioux:
Sorry for the missunderstanding. Then I cancel the PR.

PR #2267, opened 2024-06-18
Implement simple uart-based updater for the bootrom

grg-haas:
In Keystone (https://github.com/keystone-enclave/keystone), we'd like to be able to run automated CI tests for all of our platforms to make sure that code changes don't break things, as much as possible. One difficulty is that many boards don't support updating firmware remotely. For CVA6 specifically, the firmware update process involves removing the microSD card, flashing a new image, and inserting it again. This does not work well for our remote tests.

This PR implements a simple UART-based updater in the bootrom. By hitting a key during boot, a user can then transfer an image of a specific size directly to memory and boot it. We plan to transfer a uboot image with TFTP capabilities, allowing us to then quickly flash the much bigger OS image as well.

### DIFF
--- a/corev_apu/fpga/src/bootrom/src/main.c
+++ b/corev_apu/fpga/src/bootrom/src/main.c
@@ -7,12 +7,71 @@
 #include "sd.h"
 #include "gpt.h"
 
+// 1 second at 50MHz
+#define SECOND_CYCLES   (50 * 1000 * 1000)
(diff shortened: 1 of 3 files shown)


JeanRochCoulon:
@Gchauvon  This feature seems interesting to reload 32 or 64bits Linux image on SDCard and boot it. Can you review the PR ?

Gchauvon (approved):
LGTM

Issue #2274, opened 2024-06-19
[BUG] : MSTATUS.mpp


AyoubJalali:
### Is there an existing CVA6 bug for this?

- [X] I have searched the existing bug issues

### Bug Description

Hello,
This is a RTL issue related to mstatus's field mpp, according to the [CV32A65X Spec](https://docs.openhwgroup.org/projects/cva6-user-manual/04_cv32a65x/riscv/priv.html#_machine_status_mstatus_and_mstatush_registers), mpp flied is a WARL that can hold only the value 3'h11 (PRIV_LVL_M), after debugging the RTL Code coverage I have a coverage condition that should not be covered in our configuration : 
![image](https://github.com/openhwgroup/cva6/assets/110232072/327884a6-ffd7-427c-9ab9-b11e8383e1d7)
that means that the mpp hold the value 3'h10 (PRIV_LVL_HS), that's a bug in the cv32a65x, also the hole code ![image](https://github.com/openhwgroup/cva6/assets/110232072/5ee8c153-b034-40e0-9161-89d60703b34a) seems weird, how can test if mpp == PRIV_LVL_HS if we don't support Hypervisor mode.

this code was added by the PR #2035

Moschn:
Hmm, I am responsible for this behavior. The MPP field is supposed to hold the previous privilege mode before a trap. So to me it felt clear that without the RVH extension, the core cannot support the H privilege mode. After double checking, the H privilege mode is specified as reserved in the risc-v spec. 

Maybe I misunderstood something but it seems as if MPP should not accept PRIV_LEVEL_HS if the hypervisor mode is not enabled, which is the current behavior in the RTL.

Moschn:
Oh I understand the issue now. There are a few parts to this:

- The current RTL is partially wrong. The intended behavior is that if the MPP field is written with HS mode, then this is an illegal write and the value in the field should remain on a legal value (WARL). The current behavior just sets the MPP value to U-mode, which is incorrect as there are some configurations where this mode does not exist. Therefore, `mstatus_d.mpp` should not be set to user mode, but rather to the previous MPP value `mstatus_q.mpp`. 
- However, this line should be able to be tested by writing HS mode to MPP and then reading it back. The read-back value should not be HS (if the HS mode is not enabled). Essentially, just testing the WARL behavior of the MPP field.

JeanRochCoulon:
This issue should be fixed now thanks to @Moschn PR. @AyoubJalali, are you satisfied ?

[...]

PR #2285, opened 2024-06-21
Fix mstatus.mpp in relation to the possible legal values

JeanRochCoulon:
The possible legal value depend on the cva6 config: RVU, RVH, RVS.
When writing an illegal value, the previous mstatus.mpp value is kept


Moschn:
I like your change better than mine. I will close my PR.

PS: I think some of these checks are done at different places already, but its good to have them all in one place.

JeanRochCoulon:
Thanks @Moschn for your review. I will wait for the CI result to merge.

JeanRochCoulon:
Fixing #2274

PR #2309, opened 2024-07-02
add UVM interrupt agent

AyoubJalali:
This MR related to Interrupt verification : 
- UVM interrupt agent in uvme/
- Connect it into the CVA6 env & testbench
- Add some directed and generated tests into CI


AyoubJalali:
@xiaoweish could you review our interrupt agent, this is more compatible to CVA6, I saw that you're connecting the existent agent , AFAK I read this agent and it doesn't fit our verification goals, so what I'm requesting is to review our agent, and we can discuss more in CVA6 meeting

PR #2310, opened 2024-07-02
condition csr_regfile.sv

Asmaa-Kassimi:
This MR is intended to increase the code coverage.

Signed-off-by: Asmaa-Kassimi <asmaa.kassimi@external.thalesgroup.com>


core/csr_regfile.sv
```diff
@@ -2361,7 +2371,7 @@ module csr_regfile
 `ifdef PITON_ARIANE
   assign icache_en_o = icache_q[0];
 `else
-  assign icache_en_o = icache_q[0] & (~debug_mode_q);
+  assign icache_en_o = icache_q[0] & (!CVA6Cfg.DebugEn && ~debug_mode_q);
```
Gchauvon, line 2374: Why `!CVA6Cfg.DebugEn` ? If `DebugEn` is enable, icache will always be disable. We only want to disable icache when we enter debug mode which only exist if `DebugEn` is enable

```suggestion
  assign icache_en_o = icache_q[0] & (CVA6Cfg.DebugEn && ~debug_mode_q);
```

Gchauvon (changes requested):

core/csr_regfile.sv
Asmaa-Kassimi: Got it! Thanks for your remark!

PR #2314, opened 2024-07-03
feat: update cva6_hpdcache_icache_if_adapter (support continuous virtual address requests)

takeshiho0531:
- If a physical address does not arrive in the cycle following the arrival of a virtual address, an abort signal is asserted to the HPD cache.
- **I will update this PR after confirming that PR #2288  works correctly.**

### DIFF
--- a/core/cache_subsystem/cva6_hpdcache_icache_if_adapter.sv
+++ b/core/cache_subsystem/cva6_hpdcache_icache_if_adapter.sv
@@ -24,7 +24,7 @@ module cva6_hpdcache_icache_if_adapter
     parameter type fetch_drsp_t = logic,
     parameter type obi_fetch_req_t = logic,
     parameter type obi_fetch_rsp_t = logic,
-    parameter logic [CVA6Cfg.MEM_TID_WIDTH-1:0] RdTxId = 0
+    parameter logic [CVA6Cfg.MEM_TID_WIDTH-1:0] RdTxId = 0 // TODO
 )
 //  }}}
 
@@ -39,10 +39,10 @@ module cva6_hpdcache_icache_if_adapter
     input hpdcache_req_sid_t hpdcache_req_sid_i,
 
     //  Request/response ports from/to the CVA6 core
-    input fetch_dreq_t fetch_dreq_i,
-    output fetch_drsp_t fetch_dreq_o,
-    input obi_fetch_req_t obi_fetch_req_i,
-    output obi_fetch_rsp_t obi_fetch_rsp_o,
+    input fetch_dreq_t dreq_i,
+    output fetch_drsp_t dreq_o,
+    input obi_fetch_req_t fetch_obi_req_i,
+    output obi_fetch_rsp_t fetch_obi_rsp_o,
 
     //  Request port to the L1 Dcache
     output logic                        hpdcache_req_valid_o,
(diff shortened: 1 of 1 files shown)


takeshiho0531:
@cfuguet 

> I think this could be handled in the adapter you are developing. The adapter should keep an internal state that a request (with VA) was transferred to the HPDcache. If the next cycle there is no corresponding OBI request, in that case, the adapter asserts the abort signal to the cache. While aborting, you can however forward a new request (with VA) to the cache. It is a 2 stage pipeline

I implemented this as below, but I don't feel like it will work well....:’-(
If you have any hints or suggestions, please let me know.
```verilog
  logic va_transferred_q, va_is_transferred_d;

  always_ff @(posedge clk_i or negedge rst_ni) begin : transferred_state_magnegement
    if (!rst_ni) begin
      va_transferred_q <= '0;
    end else begin
      va_transferred_q <= va_is_transferred_d;
    end
  end

  assign va_transferred_d = dreq_i.req;
```
```verilog
  assign hpdcache_req_abort_o = dreq_i.kill_req || !(va_transferred_q && fetch_obi_req_i.req),
```
I have the variables `va_transferred_d` and `va_transferred_q`  express whether the VA has been transferred. (I'll change the veriables' name to better ones..)
It's not an explicit 2-stage pipeline, but basically, everything is connected with `assign` to handle new requests, while the abort signal is managed separately.

cfuguet:
> @cfuguet
> 
> > I think this could be handled in the adapter you are developing. The adapter should keep an internal state that a request (with VA) was transferred to the HPDcache. If the next cycle there is no corresponding OBI request, in that case, the adapter asserts the abort signal to the cache. While aborting, you can however forward a new request (with VA) to the cache. It is a 2 stage pipeline
> 
> I implemented this as below, but I don't feel like it will work well....:’-( If you have any hints or suggestions, please let me know.
> 
> ```verilog
>   logic va_transferred_q, va_is_transferred_d;
> 
>   always_ff @(posedge clk_i or negedge rst_ni) begin : transferred_state_magnegement
>     if (!rst_ni) begin
>       va_transferred_q <= '0;
>     end else begin
>       va_transferred_q <= va_is_transferred_d;
>     end
>   end
> 
>   assign va_transferred_d = dreq_i.req;
> ```
> 
> ```verilog
>   assign hpdcache_req_abort_o = dreq_i.kill_req || !(va_transferred_q && fetch_obi_req_i.req),
> ```
> 
> I have the variables `va_transferred_d` and `va_transferred_q` express whether the VA has been transferred. (I'll change the veriables' name to better ones..) It's not an explicit 2-stage pipeline, but basically, everything is connected with `assign` to handle new requests, while the abort signal is managed separately.

Hi @takeshiho0531,

I see some issues in your code.

- First you need to make the va_transferred_d signal dependant on the hpdcache_req_ready_i signal. This is because the virtual address is only effectively transferred when the cache accepts it.

```systemverilog
assign va_transferred_d = dreq_i.req & hpdcache_req_ready_i;
```

- Second, the abort signal, I would affect it as below. You need to abort a request only if you transferred previous a virtual address but the next cycle this request is aborted by the core, or the core did not forward the OBI part of the request :

```systemverilog
assign hpdcache_req_abort_o = va_transferred_q & (dreq_i.kill_req | ~fetch_obi_req_i.req),
```

PR #2324, opened 2024-07-05
Add CV64A6_MMU core in user manual

LQUA:
In the CVA6 user manual, two configurations are already explicit. 
This commit adds another one : CV64A6_MMU.

This MR is independant from the MR adding design documentation for CV64A6_MMU


JeanRochCoulon:
@jquevremont   Are you ok with these modifications ?

PR #2327, opened 2024-07-05
Update number of PMP regions

JeanRochCoulon:
from 16 to 64


JeanRochCoulon:
@zchamski    I have updated the specification to increase the PMP region number. This has been fixed in RTL. Can you approve?  This will also require an update on riscv-config side.

Issue #2230, opened 2024-06-11
GCC 13.1.0 Prerequisites to run benchmark.sh 


Tanishqgithub:
### Is there an existing CVA6 bug for this?

- [X] I have searched the existing bug issues

### Bug Description

Re-opening the issue #2200 

I saw that while reading the README for Verification Environment in https://github.com/openhwgroup/cva6/blob/master/verif/README.md there is a Prerequisites for GCC 13.1.0 Compiler . 
As per https://github.com/openhwgroup/cva6/blob/master/util/gcc-toolchain-builder/README.md we have already build the GCC 13.1.0 baremetal 
Is it necessary to built again the gcc 13.1.0 compiler to run benchmark.sh and others in ./verif/regress.
And if it is necessary i tried to install the toolchain but facing the issue : 
![image](https://github.com/openhwgroup/cva6/assets/108278729/fd992b2d-46c5-430c-a322-59b9ed8ef1f9)
![image](https://github.com/openhwgroup/cva6/assets/108278729/277ddc07-8b66-47c3-b115-7cacd6d226d5)
How to overcome in cloning the submodule binutils

And with reference to #2200 how can i run the benchmark.sh on cv32a60x or any other then cv64a6_imafdc_sv39 
[steps i need to follow to run on cv32a60x explicitly]

To run hello_world and smoke test : 
export RISCV=/home/pescampus/Desktop/CVA6/cva6/util/gcc-toolchain-builder/gcc-13.1.0-baremetal/
whereas to run bechmarks.sh and others in ./verif/regress.
![image](https://github.com/openhwgroup/cva6/assets/108278729/f025ca2d-5f81-4e9a-bf55-e941de58f5f1)
I just want to ensure I'm on the right track. Am I understanding correctly, or do I need clarification? Your feedback would be greatly appreciated.

JeanRochCoulon:
It is necessary to build one time GCC.
To use it (and prevent to re-build it), please setup the RISCV shell variable.

Tanishqgithub:
As you said if i use the gcc13.1.0 baremetal to run the benchmark.sh to setup the RISCV variables i don't have **/bin/riscv64-unknown- ; /bin/riscv64-unknown-gcc** to setup for  RISCV_PREFIX ; RISCV_GCC ; CV_SW_PREFIX variables . 

[...]

PR #2338, opened 2024-07-08
Remove duplicate and out of date infos on verif readme

valentinThomazic:
The verif readme is out of date and its instructions are both duplicate from the root and toolchain README and are misleading (see https://github.com/openhwgroup/cva6/issues/2230)

This remove the duplicate part leaving only the verif/coverage-oriented instructions and potentially useful verif env variable with updated values.


JeanRochCoulon:
@AyoubJalali   You maybe contributed to this README. Can you have a look ?

PR #2343, opened 2024-07-10
Update number of PMP entries

JeanRochCoulon:
From 16 to 64 entries


JeanRochCoulon:
@ASintzoff Can you merge ?

PR #2362, opened 2024-07-12
increase code coverage on second ALU

cathales:
A new parameter is added to condition the branch-related logic in the
`alu` module.  This parameter is disabled on the second ALU because it
is not bound to a branch unit.


cathales:
It was wasting 753 gates, and after an update with the `master` branch it is now saving 530 gates! This is the exact same modification. #2431 has the same issue.

PR #2369, opened 2024-07-16
fix single-step which was x in cv32a65x config and fix mcycle for double commit 

cathales:
Thanks to @AyoubJalali who pointed out the issue!


AyoubJalali:
@cathales we can merge this ?

PR #2374, opened 2024-07-19
Update csr_regfile.sv to fix #2373

### LINKED ISSUE #2373: [BUG] Wrong exception cause being raised in CSR Logic (illegal instruction instead of Virtual Instruction exception)

### Bug Description

It the Privilege Check section of CSR Logic 

https://github.com/openhwgroup/cva6/blob/8c709767599a12acd7c02976a42cc3dcc9f4604e/core/csr_regfile.sv#L2250

``` systemverilog
        if (access_priv < csr_addr.csr_decode.priv_lvl) begin
          if (v_q && csr_addr.csr_decode.priv_lvl == riscv::PRIV_LVL_HS)
            virtual_privilege_violation = 1'b1;
          else privilege_violation = 1'b1;
        end
```

Consider the case the hart is in VU mode and trying to access the Supervisor CSR `sstatus`, the logic would raise an illegal instruction exception (privilege_violation). The issue is that the `sstatus` (or any other supervisor CSR) has privilege level (`PRIV_LVL_S`). This is wrong functionality as spec says we should raise Virtual Instruction Exception.

The riscv-isa-manual says the following :

https://github.com/riscv/riscv-isa-manual/blob/3c1d60298f16523aba30d45ba4d7c9381a4e2d4c/src/hypervisor.adoc?plain=1#L2026C1-L2028C26
```
* in VU-mode, attempts to access an implemented non-high-half supervisor
CSR when the same access (read/write) would be allowed in HS-mode,
assuming `mstatus`.TVM=0;
```

The following functionality is found in spike RISCV ISA Simulator

https://github.com/riscv-software-src/riscv-isa-sim/blob/34601fc48851e49f7ecbbd87405b8b7afcc4e59b/riscv/csrs.cc#L44C1-L48C4
``` c
  if (priv < csr_priv) {
    if (state->v && csr_priv <= PRV_HS)
      throw trap_virtual_instruction(insn.bits());
    throw trap_illegal_instruction(insn.bits());
  }
```
Note the condition `csr_priv <= PRV_HS`

---

Possible solution: Modifying the condition in question to below would solve the above mentioned bug

``` systemverilog
        if (access_priv < csr_addr.csr_decode.priv_lvl) begin
          if (v_q && csr_addr.csr_decode.priv_lvl <= riscv::PRIV_LVL_HS)
            virtual_privilege_violation = 1'b1;
          else privilege_violation = 1'b1;
        end
```

somyadashora:
Merging in these changes fixes #2373

### DIFF
--- a/core/csr_regfile.sv
+++ b/core/csr_regfile.sv
@@ -2247,7 +2247,7 @@ module csr_regfile
       // precedence over interrupts
       if (csr_op_i inside {CSR_WRITE, CSR_SET, CSR_CLEAR, CSR_READ}) begin
         if (access_priv < csr_addr.csr_decode.priv_lvl) begin
-          if (v_q && csr_addr.csr_decode.priv_lvl == riscv::PRIV_LVL_HS)
+          if (v_q && csr_addr.csr_decode.priv_lvl <= riscv::PRIV_LVL_HS)
             virtual_privilege_violation = 1'b1;
           else privilege_violation = 1'b1;
         end


JeanRochCoulon:
@ninolomata @AngelaGonzalezMarino Can you approve this PR?

JeanRochCoulon:
Thanks @somyadashora fro your issue explanation and RTL fix. LGTM

PR #2408, opened 2024-07-26
Set the environment configuration only from env_cfg constraints.

AEzzejjari:
Define the environment configuration only from env_cfg constraints and do not reconfigure existing core_cntrl_cfg fields in cva6pkg_to_core_cntrl_cfg.

Set env_cfg.CVA6Cfg in the test class so that it can be used during the randomization process.


AEzzejjari:
Hello @MarioOpenHWGroup
"In this PR, I deleted the configuration in uvma_cva6pkg_utils.sv, which caused a bug in the Spike configuration when executing testharness. So, I think you use this file to configure Spike Tandem.

JeanRochCoulon:
@AEzzejjari  Your PR is in draft, if you would like to merge, can you switch it

PR #2420, opened 2024-07-30
TANDEM Configuration fixes

MarioOpenHWGroup:
* Disable yaml config for embedded configuration (it does not pass all the core params)
* Change RAM TANDEM setup

### DIFF
--- a/verif/regress/smoke-tests.sh
+++ b/verif/regress/smoke-tests.sh
@@ -33,7 +33,7 @@ if ! [ -n "$DV_SIMULATORS" ]; then
 fi
 
 if ! [ -n "$UVM_VERBOSITY" ]; then
-    export UVM_VERBOSITY=UVM_NONE
+    export UVM_VERBOSITY=UVM_LOW
 fi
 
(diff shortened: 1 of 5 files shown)


verif/tb/core/uvma_cva6pkg_utils.sv
zchamski, line 1: Lines 45-46: the two values should be exchanged (base 'h80_000_000, size 'h40_000_000).

verif/regress/smoke-tests.sh
```diff
@@ -45,6 +45,8 @@ CC_OPTS="-static -mcmodel=medany -fvisibility=hidden -nostdlib -nostartfiles -g
 cd verif/sim/
 make -C ../.. clean
 make clean_all
+python3 cva6.py --testlist=../tests/testlist_riscv-compliance-cv32a60x.yaml --test rv32i-I-ADD-01 --iss_yaml cva6.yaml --target cv32a6_imac_sv32 --iss=$DV_SIMULATORS $DV_OPTS
+exit
```
zchamski, line 49: Foe master the test should not shortened like this.

MarioOpenHWGroup: Yes, yes, this was intended for testing purposes only on CI as I did not have the error in local

PR #2431, opened 2024-08-02
core/store_buffer.sv: make address check XLEN dependant

fpetrot:
Address check is currently performed on [11:3] which (a) is (very slightly) suboptimal on rv32, but more important (b) hides the intent of the check.

Suggested-by: Pierre Ravenel <pierre.ravenel@kalray.eu>


JeanRochCoulon:
Hello @fpetrot,   I took the liberty of committing in your PR to (try to) make the CI pass. I hope it is ok for you.

fpetrot:
Thanks for the CI.
A quick anlysis of the failure show that the change leads to a rather small benefit on coremark (-243 cycles) while needing (to my surprise, given the small change) 216 additional gates. Interestingly enough, it lowers by 2 the number of spyglass warnings.
Anyway, I leave it to you to decide if this patch is worth integrating.

cathales:
It is a 1.25‰ area cost (actually more if we only consider the pipeline) for a 0.45‰ performance gain.
I think we should not merge this PR, at least in the current state.

It is not the first time we have strange variations in area.
Once, we had a +1kG for a change which was supposed to remove a few gates (or at least keep the same number due to unused gates elision). 🤷

PR #2436, opened 2024-08-08
[Spike tandem] Fix Yaml config files for CV32A65X. Fix Questa tandem. Add workaround for AXI end-of-test asserts.

zchamski:
Fix incorrect structure of the Spike Yaml file.
Fix ISA strings in riscv-config spec and outputs.
Fix Questa tandem invocation.
Add 100 cycle delay after end of test to settle outstanding AXI requests when using UVM.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -307,7 +307,7 @@ ifdef preload
 endif
 
 ifdef spike-tandem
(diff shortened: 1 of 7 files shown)


zchamski:
Simulation failures appear linked to AXI read transactions that are ongoing when the test terminates and the DUT clock is stopped.

Hypothesis (TBC): such transactions originate from the cache which is performing a line fill.

PR #2438, opened 2024-08-09
[CI] use spike tandem on smoke-tests

valentinThomazic:
* Enable Spike Tandem results to be taken into account on the CI with VCS
* tests are now only run with UVM on cv32a65x
* Fix for questa standalone simulation
* Add the dashboard on top of the readme to find it more easily


cathales:
This PR introduces failures in `master` branch.

PR #2460, opened 2024-08-23
fix CI

cathales:
Fixup for https://github.com/openhwgroup/cva6/pull/2438


cathales:
Ok it fixes 1 error, but it was not the only one

MikeOpenHWGroup (approved):

PR #2468, opened 2024-08-27
fix #2464 by implementing Inhibit CSR as read only zero

### LINKED ISSUE #2464: [BUG] : mcountinhibit raise exception

### Bug Description

There was a misunderstanding of the specification, the `mcountinhibit` should have zero in it if we don't support perf counter, and should not raise an exception, so it's accessible but not supported.

JeanRochCoulon:
Exception is not generated when INHIBIT CSR is accessed


JeanRochCoulon:
@AyoubJalali  Can you confirm the bug fix ?

cathales (approved):

JeanRochCoulon:
@MarioOpenHWGroup For information, inhibit CSR has been fixed, to be taken into account into your Spike-Tandem modification.

PR #2074, opened 2024-04-24
[CSR] Fix bits when RVS and RVU not available

MarioOpenHWGroup:
Disable some bits when RVU or RVS not present

PR #2469, opened 2024-08-27
Fix pmpaddr read logic considering G=2

### LINKED ISSUE #2465: [BUG] PMP config can not select `NA4` mode while granularity is set to 4

### Bug Description

While the granularity of the PMP can determine by following code is 4 bytes:

```
    asm volatile ("csrw pmpcfg0, x0");
    asm volatile ("addi a3,x0,-1");
    asm volatile ("csrw pmpaddr0, a3");
    asm volatile ("csrr a3, pmpaddr0"); // a3=0x003FFFFFFFFFFFFF -> 
```
`a3 = 0x003FFFFFFFFFFFFF` so, the least-significant bit that set is 0, so the PMP granularity is 4 bytes.

But CVA6 does not support setting the `NA4` address matching mode. In other words, if we set the address matching mode to `0x2`, the core rewrites the register with `0x0`, which means no protection.

To reproduce, execute following instruction:
```
int main(void){
    asm volatile ("auipc	a3,0x0");
    asm volatile ("addi     a3,a3,0x24");
    asm volatile ("addi   t1,x0,0x2");
    asm volatile ("srl      t1,a3,t1");
    asm volatile ("csrw pmpaddr0, t1");
    asm volatile ("addi t0,x0,0x94");
    asm volatile ("csrw pmpcfg0, t0");

    asm volatile ("csrr a4, pmpcfg0");
    asm volatile ("csrr a4, pmpaddr0");
    asm volatile ("auipc    a3,0x0");
    asm volatile ("lbu a4,0(a3)");
}
```

However, if you execute same code on Spike you will get `trap_load_access_fault` exception.

    - Spike version: `1.1.1-dev`
    - version: `CVA6 commit: 7435cb310ee98c246edd15d00ec236ad2c6ff490`
    - OS: `CentOS Linux release 7.9.2009 kernel: 5.15.0-78-generic`
    - Simulator: `VCS_2023`

Moschn:
fixes #2465

This is a regression from #2074 

As a side note, this would have been caught if we would run the pmp benchmark in the CI. While it is a benchmark by name, I think its more of a test.


JeanRochCoulon:
Hello @Moschn   You speak about a PMP benchmark. Where is this test/benchmark ?

PR #2479, opened 2024-08-29
fix jal riscv-arch-test

valentinThomazic:
Fix the riscv-arch-test jal test by increasing sram size


ASintzoff:
For the record, the goal of the modification is to increase the allocated SRAM which was too small to store the failing test.

PR #2508, opened 2024-09-23
Fix misaligned jump target exception handling

### LINKED ISSUE #2505: [BUG] Jump to Misaligned Address Fails to Trigger Exception

### Bug Description

In RISC-V, jumps to misaligned instruction addresses should trigger a misaligned fetch exception. However, in the latest CVA6 core, attempting to jump to a misaligned address dose not throw an exception. For example 0x80000005, which should trigger a misaligned fetch exception. Instead, the core executes the instruction without generating the expected exception. This vulnerability allows code to execute from misaligned addresses without raising an exception, leading to potential instability or undefined behavior.

To reproduce, execute following instruction:
```
int main(void){
     void (*misaligned_jump)() = (void (*)()) 0x80000005;  // Misaligned jump address
    asm volatile ("jalr %0, 0(%1)" : "=r" (rd_value) : "r" (misaligned_jump)); 
}
```
```
- Spike version: `1.1.1-dev`
- version: `CVA6 commit: f974e105bf88cf81c32f55789b0baab4fe4d16c9`
- OS: `CentOS Linux release 7.9.2009 kernel: 5.15.0-78-generic`
- Simulator: `VCS_2023`
```

teobiton:
As per the specifications, instruction-address-misaligned exceptions should be reported for instructions that are not 16-bit aligned if the C extension is supported, and for instructions that are not 32-bit aligned if not.

Fixes #2505

### DIFF
--- a/core/branch_unit.sv
+++ b/core/branch_unit.sv
@@ -118,12 +118,12 @@ module branch_unit #(
     branch_exception_o.gva   = CVA6Cfg.RVH ? v_i : 1'b0;
     // Only throw instruction address misaligned exception if this is indeed a `taken` conditional branch or
     // an unconditional jump
-    if (!CVA6Cfg.RVC) begin
-      jump_taken = !(ariane_pkg::op_is_branch(fu_data_i.operation)) ||
-          ((ariane_pkg::op_is_branch(fu_data_i.operation)) && branch_comp_res_i);
-      if (branch_valid_i && (target_address[0] || target_address[1]) && jump_taken) begin
-        branch_exception_o.valid = 1'b1;
-      end
+    jump_taken = !(ariane_pkg::op_is_branch(fu_data_i.operation)) ||
+      ((ariane_pkg::op_is_branch(fu_data_i.operation)) && branch_comp_res_i);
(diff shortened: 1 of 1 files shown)


zarubaf:
You shouldn't be able to trigger this exception if RVC is enabled. Since the LSB is cleared:

```
The indirect jump instruction JALR (jump and link register) uses the I-type encoding. The target
address is obtained by adding the sign-extended 12-bit I-immediate to the register rs1, then setting the least-significant bit of the result to zero.
```

teobiton:
Yes, I realized this after opening the PR.
I feel like the specifications are not clear on this point, because it states that this exception should be raised in case of misalignment ... but because bit 0 is supposed to be zero it _shouldn't_ happen.

zarubaf:
> Yes, I realized this after opening the PR. I feel like the specifications are not clear on this point, because it states that this exception should be raised in case of misalignment ... but because bit 0 is supposed to be zero it _shouldn't_ happen.

Indeed, could be clearer about it. But I think it is exactly how you wrote it. It should trigger it, but you can't trigger it.

PR #2523, opened 2024-10-02
Fix CVV#2531: Make mie.MSIE and mip.MSIP RO-zero, prevent SW writes to mip.

zchamski:
Modify Spike Yaml configuration for CV32A65X to match the annotated CV32A65X privileged ISA spec.


AyoubJalali:
LGTM

PR #2527, opened 2024-10-04
Interrupt verif : Implement clear mechanism in interrupt's agent

AyoubJalali:
This MR implement set/clear interrupt based on protocol in : https://github.com/openhwgroup/cva6/blob/master/verif/docs/Protocols/interrupt-verification.adoc


AyoubJalali:
Hello @MikeOpenHWGroup, Could you please take a look ?

verif/env/corev-dv/cva6_asm_program_gen.sv
```diff
@@ -213,7 +216,52 @@ class cva6_asm_program_gen_c extends riscv_asm_program_gen;
     if (cfg_cva6.mtvec_mode == VECTORED) begin
       push_gpr_to_kernel_stack(status, scratch, cfg_cva6.mstatus_mprv, cfg_cva6.sp, cfg_cva6.tp, instr);
     end
-    gen_signature_handshake(instr, CORE_STATUS, HANDLING_EXCEPTION);
+    push_used_gpr_to_kernel_stack(status, scratch, 3, cfg_cva6.mstatus_mprv, cfg_cva6.sp, cfg_cva6.tp, instr);
+    instr = {instr,
+             // The trap is caused by an exception, read back xCAUSE, xEPC to see if these
+             // CSR values are set properly. The checking is done by comparing against the log
```
MikeOpenHWGroup, line 222: > The checking is done by comparing against the log generated by ISA simulator (spike).

This is _not_ true in UVM environment of CORE-V-VERIF. In CVV, the checking is down by lock-step comparisons against the Spike-Tandem reference model.

verif/env/uvme/uvma_interrupt/seq/uvma_interrupt_seq.sv
```diff
@@ -11,25 +11,37 @@
...
     */
    extern function new(string name="uvma_interrupt_seq");
 
+   extern virtual task automatic clear_irq_channel(int channel, uvma_interrupt_seq_item_c  req_item);
+
    extern virtual task body();
 
 endclass : uvma_interrupt_seq_c
 
+task automatic uvma_interrupt_seq_c::clear_irq_channel(int channel, uvma_interrupt_seq_item_c  req_item);
```
MikeOpenHWGroup, line 31: There is an opportunity for deadlock here if the test-program fails to write `irq_addr`.   Recommend that this task should call `uvm_fatal()` after some "reasonable" number of cycles.

verif/env/uvme/uvma_interrupt/seq/uvma_interrupt_seq.sv
```diff
@@ -38,22 +50,46 @@ endfunction : new
...
+  if (cfg.enable_interrupt) begin
+      for (int i = 0; i < cfg.num_irq_supported; i++) begin
+          automatic int ii = i;
+          automatic uvma_interrupt_seq_item_c  req_item_c;
+            fork begin
+               forever begin
+                      // Set interrupt request per channel
+                      req_item_c = uvma_interrupt_seq_item_c::type_id::create("req_item_c");
+                      start_item(req_item_c);
+                      assert(req_item_c.randomize() with {
```
MikeOpenHWGroup, line 62: Please do not use `assert` for checking for successful randomization.  If/when randomization fails, the `assert` will produce a simulator-dependent error message and this will not be "seen" by the UVM messaging service and the simulation may pass with a false positive.

A better implementation is as follows:
```
if (!req_item_c.randomize() with {
                      req_item_c.interrupt_vector[ii] inside {0 , 1};
                      req_item_c.interrupt_channel_mask == 1<<ii;
                      })
   `uvm_error(get_type_name(), "req_item_c.randomize() failed!")
```

verif/env/uvme/uvma_interrupt/seq/uvma_interrupt_seq.sv
```diff
@@ -38,22 +50,46 @@ endfunction : new
...
+                         end
+                      end
+                   end
+               end
+         join_none
+      end
+      wait fork;
+  end
+  else begin
+     `uvm_warning(get_type_name(), $sformatf("Interrupts are disabled"));
```
MikeOpenHWGroup, line 91: Should this be a warning?  A lot of tests will be run without interrupts.

verif/env/uvme/uvma_interrupt/uvma_interrupt_mon.sv
```diff
@@ -42,73 +42,119 @@ class uvma_interrupt_mon_c extends uvm_monitor;
...
+endfunction
+
+
+function void uvma_interrupt_mon_c::build_phase(uvm_phase phase);
+
+   super.build_phase(phase);
+
+   void'(uvm_config_db#(uvma_interrupt_cntxt_c)::get(this, "", "cntxt", cntxt));
+   if (cntxt == null) begin
+      `uvm_fatal("build_phase", "monitor cntxt class failed")
```
MikeOpenHWGroup, line 84: The `ID` argument should be set to easily identify the source of the message.  "build_phase" is implemented by many uvm_components, so this is probably not useful.   Recommend using either `get_type_name()` or some string that makes it easy to find the source of the message (maybe "INTR_MON"?).

verif/env/uvme/uvma_interrupt/uvma_interrupt_mon.sv
```diff
@@ -42,73 +42,119 @@ class uvma_interrupt_mon_c extends uvm_monitor;
...
+   end
+
+   void'(uvm_config_db#(uvma_interrupt_cfg_c)::get(this, "", "cfg", cfg));
+   if (cfg == null) begin
+      `uvm_fatal("CFG", "Configuration handle is null")
```
MikeOpenHWGroup, line 89: Same comment.

MikeOpenHWGroup (changes requested):
Whew!  A very impressive contribution @AyoubJalali!  I am especially happy to see you using the templated RESET run-flow.  I have made a few comments on specific files, but in general this is really great stuff.

AyoubJalali:
Thanks @MikeOpenHWGroup 
I'm on it !

AyoubJalali:
I make the changes requested @MikeOpenHWGroup, but i had a git problem trying to update this branch
I'll create a new PR with the changes requested
I'll close this one when the other PR is ready

verif/env/uvme/uvma_interrupt/seq/uvma_interrupt_seq.sv
AyoubJalali: This is just to help for debug, if any one run a interrupt test without enabling interrupt, it gives a warning

AyoubJalali:
I rebase the branch, also I applied also your suggestions, thanks

verif/env/uvme/uvma_interrupt/seq/uvma_interrupt_seq.sv
AyoubJalali: I had this idea, but i didn't have this "reasonable" number of cycle, so I let it go like that in first place, but i'll consider it

MikeOpenHWGroup: There are two related questions here:
1.  **What is a "reasonable timeout value?** 
A typical "reasonable-response-timeout" is the average length of a test.  I am not certain about this, but I would guess that it should be 100_000 clock cycles.
3. **How to implement it?**
The timeout should be a random member of the Agent's configuration object and there should be a soft constraint to define a reasonable value.  That way, the timeout can be easily controlled from a `uvm_testcase`.

verif/env/uvme/uvma_interrupt/seq/uvma_interrupt_seq.sv
MikeOpenHWGroup: No, this is not correct.  Warnings are reported in the `report` phase and we cannot sign-off TRL-5 if there are warnings.  So this should be a `uvm_info()`, not `uvm_warning()`.

verif/env/uvme/uvma_interrupt/seq/uvma_interrupt_seq.sv
AyoubJalali: I implement TIMEOUT variable, I suggest to only give 10_000 clock cycle after exit the simulation

verif/env/uvme/uvma_interrupt/seq/uvma_interrupt_seq.sv
AyoubJalali: Done

MikeOpenHWGroup (approved):

PR #2529, opened 2024-10-04
Run benchmarks in single and dual issue config with big caches

JeanRochCoulon:
Display  report at the end of dhrystone and coremark executions
Run benchmarks in single and dual issue
Use bigger cache config


yanicasa:
ok lgtm

PR #2538, opened 2024-10-08
Multi commits to benchmark CVA6, to modify clock period and to create new vcs-uvm-gate

JeanRochCoulon:
* Create job dedicated to benchmark CVA6
* Set clock period at 20ns and fix vcs-uvm simulation time
* Add vcs-uvm-gate ISS target


Gchauvon (approved):
LGTM

PR #2543, opened 2024-10-11
Performance tb

AnouarZajni:
Add performance_mode to be set in line command using +tb_performance_mode to speed up simulation time by disabling functional coverage, agents logger, rvfi csr checks and scoreboard. 
This +tb_performance_mode switch has also been added in related regression tests


JeanRochCoulon:
Well done @AnouarZajni

PR #2548, opened 2024-10-15
Use uvm testbench to run gate simulations

JeanRochCoulon:
cva6.sv: enable CSR RVFI tracing during gate simulation

Thanks @ajalali and @ycasamat for their support


Gchauvon (approved):
LGTM

PR #2552, opened 2024-10-17
Move assertion statement outside of `always_comb` block

ricted98:
The assertion included in the `always_comb` block apparently violates the requirements in [section 9.2.2.2.2 of the SystemVerilog standard](https://ieeexplore.ieee.org/document/10458102):

> Statements in an always_comb shall not include those that block, have blocking timing or event
> controls, or fork-join statements.

My compilation flow uses Synopsis `vlogan`. Namely, the reported error is:

> Error-[SV-BCACF] Blocking construct in always_comb/ff
> 
> tag_cmp
>   Statements in an always_comb shall not include those that block, have 
>   blocking timing or event controls, or forkjoin statements. The always_ff 
>   procedure imposes the restriction that it contains one and only one event 
>   control and no blocking timing controls.
>   Try using simple always blocks.


JeanRochCoulon:
Great @ricted98 !!

PR #2555, opened 2024-10-17
increase code coverage in commit stage

cathales:
Fixes a code coverage issue reported by @AyoubJalali


cathales:
I do not understand why the ECA check fails. I re-validated several times, re-signed the eclipse agreement which was still valid, re-validated again, checked the name, e-mail address and GitHub account name on Eclipse side, and everything is correct.

PR #2558, opened 2024-10-21
doc: keep documentation in sync with the code

slgth:
Both the ISA and design documentations use some parameters generated from the RTL (ports, parameters).
As of now, they are committed to the repository and can be out of sync with the code.

This PR removes them from the repository and freshly generates them from the code when building HTML files.

This PR also removes prebuilt HTML files (design & ISA docs) and generates them when building the top-level Read the Docs documentation (`make -C docs`).


JeanRochCoulon:
Great @slgth !   Thank for your different CVA6 contributions

PR #2573, opened 2024-11-02
cache_ctrl: Generalise AXI offset generation

niwis:
For `XLEN = 64`, some tools (e.g. VCS) still elaborate the offset generation block for `XLEN = 32`, throwing an elaboration error (illegal bit access). Hence, VCS can currently not build the 64-bit WB configuration of CVA6. Fix this by generating the AXI offset in an equivalent, parameter-agnostic and tool-friendly way.


JeanRochCoulon:
Thanks @niwis for having used your personal Github repository to submit the PR !

PR #2587, opened 2024-11-08
csr_regfile: Fix S-mode traps when H extension is enabled

niwis:
Since #2346, if Hypervisor extension is enabled, the logic required to properly trap to S mode is excluded. Fix this by adjusting the if condition.

### DIFF
--- a/core/csr_regfile.sv
+++ b/core/csr_regfile.sv
@@ -1829,25 +1829,23 @@ module csr_regfile
 
       // trap to supervisor mode
       if (CVA6Cfg.RVS && trap_to_priv_lvl == riscv::PRIV_LVL_S) begin
-        if (CVA6Cfg.RVH) begin
-          if (trap_to_v) begin
-            // update sstatus
-            vsstatus_d.sie = 1'b0;
-            vsstatus_d.spie = vsstatus_q.sie;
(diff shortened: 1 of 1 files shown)


AyoubJalali:
Hello @niwis the logic doesn't change here ? that's right ?
Because of the reasons of the PR #2346 is increase code coverage, the simulator doesn't recognize the condition in  _if (CVA6Cfg.RVH && trap_to_v) begin_ as a constant to exclude the code related in code coverage, that's why we dispatch the condition

niwis:
@AyoubJalali #2346 *did* change the logic in a breaking way since the `else` block [below the diff](https://github.com/openhwgroup/cva6/pull/2587/files#diff-7dd6575b20e68a1a956b20a0921b8596ca0d2ab9790d8dec58110cfcaaa68bffL1852) is no longer taken if (`RVH && !trap_to_v`). This is a critical bug since S-mode traps no longer update `sstatus`, `scause`, `sepc`, and `stval` if `CVA6Cfg.RVH == 1`. This PR fixes this by reverting the breaking part of #2346.

JeanRochCoulon:
@niwis   I merged, then reverted this PR because the CI was failed (my bad). Feel free to submit again.

PR #2588, opened 2024-11-08
iro: remove an unreachable statement

cathales:
Also removes outdated comments.


JeanRochCoulon:
800 gates => pretty good !

PR #2590, opened 2024-11-08
add CVA6ConfigFpgaAltera parameter

AngelaGonzalezMarino:
First step to add FpgaAltera optimization parameter


core/include/build_config_pkg.sv
```diff
@@ -75,7 +76,7 @@ package build_config_pkg;
     cfg.RVZicntr = CVA6Cfg.RVZicntr;
     cfg.RVZihpm = CVA6Cfg.RVZihpm;
     cfg.NR_SB_ENTRIES = CVA6Cfg.NrScoreboardEntries;
-    cfg.TRANS_ID_BITS = $clog2(CVA6Cfg.NrScoreboardEntries);
+    cfg.TRANS_ID_BITS = CVA6Cfg.NrScoreboardEntries > 1 ? $clog2(CVA6Cfg.NrScoreboardEntries) : CVA6Cfg.NrScoreboardEntries;
```
JeanRochCoulon, line 79: I do not understand this modification: NrScoreBoardEntries is always >1 otherwise the pipeline will be stuck. To me, no need to overload the code with this condition.

core/include/cv32a6_thales_avs_config_pkg.sv
```diff
@@ -0,0 +1,161 @@
+// Copyright 2021 Thales DIS design services SAS
```
JeanRochCoulon, line 1: I highly suggest you to define this new configuration from cv32a65x_config_pkg.sv. A lot of localparams have been removed to increase the code visibility.

core/include/config_pkg.sv
```diff
@@ -214,7 +216,12 @@ package config_pkg;
     int unsigned VMID_WIDTH;
 
     bit FpgaEn;
+    bit FpgaAltera;
     bit TechnoCut;
+    /// Number of commit ports, i.e., maximum number of instructions that the
```
JeanRochCoulon, line 221: Is this comment needed ?   If you insist, this should be located right before the NrCommitPorts declaration.

core/include/cv32a60x_config_pkg_deprecated.sv
```diff
@@ -53,6 +53,7 @@ package cva6_config_pkg;
   localparam CVA6ConfigNrScoreboardEntries = 4;
 
   localparam CVA6ConfigFpgaEn = 0;
+  localparam CVA6ConfigFpgaAltera = 0; //usually disabled, default is xilinx. Needs FpgaEn =1 to be effective
```
JeanRochCoulon, line 56: Can we modify CVA6ConfigFpgaAltera into CVA6ConfigFpgaAlteraEn ?
Can we modify the comment into :
CVA6ConfigFpgaEn =1; // for Xilinx and Altera
CVA6ConfigFpgaAlteraEn =1; // for Altera (only)

core/include/cv32a6_thales_avs_config_pkg.sv
AngelaGonzalezMarino: that should not be there, sorry

core/include/config_pkg.sv
AngelaGonzalezMarino: no, I think this got mixed in the rebase. I am removing it

core/include/cv32a60x_config_pkg_deprecated.sv
AngelaGonzalezMarino: ok

core/include/build_config_pkg.sv
AngelaGonzalezMarino: we have used NrScoreboardEntries set to 1 to reduce resources. However this is not related to the fpgaAltera parameter, so I am removing the change here and will explain better in a future PR

core/include/config_pkg.sv
```diff
@@ -170,6 +170,8 @@ package config_pkg;
     int unsigned                 FetchUserWidth;
     // Is FPGA optimization of CV32A6
```
JeanRochCoulon, line 171:
```suggestion
    // Is FPGA optimization of CV32A6 for Xilinx and Altera
```

AngelaGonzalezMarino:
thanks!

PR #2592, opened 2024-11-08
Altera opt 1

AngelaGonzalezMarino:
The first optimization for Altera FPGA is to move the instruction queue to LUTRAM. The reason why the optimization previously done for Xilinx is not working, is that in that case asynchronous RAM primitives are used, and Altera does not support asynchronous RAM. Therefore, this optimization consists in using synchronous RAM for the instruction queue and FIFOs inside wt axi adapter. 

The main changes to the existing code are:

- New RAM module to infer synchronous RAM in altera with independent read and write ports (SyncDpRam_ind_r_w.sv) 

- Changes inside cva6_fifo_v3 to adapt to the use of synchronous RAM instead of asynchronous: 

- When the FIFO is not empty, next data is always read and available at the output hiding the reading latency introduced by synchronous RAM (similar to fall-through approach). This is a simplification that is possible because in a FIFO we always know what is the next address to be read. 
           
 - When data is read right after write, we can’t use the previous method because there is a latency to first write the data in the FIFO, and then to read it. For this reason, in the new design there is an auxiliary register used to hide this latency. This is used only if the FIFO is empty, so we detect when the word written is first word, and keep it in this register. If the next cycle comes a read, the data out is taken from the aux register. Afterwards the data is already available in the RAM and can be read continuously as in the first case. 

All this is only used inf FpgaAlteraEn parameter is enabled, otherwise the previous implementation with asynchronous RAM applies (when FpgaEn is set), or the register based implementation (when FpgaEn is not set).


core/cva6_fifo_v3.sv
```diff
@@ -9,9 +10,12 @@
 // specific language governing permissions and limitations under the License.
 
 // Author: Florian Zaruba <zarubaf@iis.ee.ethz.ch>
+// Additional contributions by:
+//                 Angela Gonzalez - PlanV Technologies
 
 module cva6_fifo_v3 #(
     parameter bit FALL_THROUGH = 1'b0,  // fifo is in fall-through mode
+    parameter bit FPGA_ALTERA = 1'b0,  // fifo is in fall-through mode
```
JeanRochCoulon, line 18: The comment shoub be removed/updated

AngelaGonzalezMarino: true

Gchauvon:
LGTM

JeanRochCoulon:
Hello @AngelaGonzalezMarino    The ASIC synthesis produces 1.2 more kgates. Do you know why ?

JeanRochCoulon:
I get an hypothesis: some registers have been added in cva6_fifo_v3. They need to be gated by the FPGA Altera Enable parameter.

AngelaGonzalezMarino:
> I get an hypothesis: some registers have been added in cva6_fifo_v3. They need to be gated by the FPGA Altera Enable parameter.

maybe, let me try

core/cva6_fifo_v3.sv
```diff
@@ -116,14 +126,22 @@ module cva6_fifo_v3 #(
...
+      end
       if (pop_i) begin
+        first_word_n = '0;
         status_cnt_n = status_cnt_q;
         read_pointer_n = read_pointer_q;
         write_pointer_n = write_pointer_q;
       end
     end
+
+    if (FPGA_EN) fifo_ram_read_address = (FPGA_ALTERA == 1) ? read_pointer_n : read_pointer_q;
```
JeanRochCoulon, line 143:
```suggestion
    if (FPGA_EN) fifo_ram_read_address = (FPGA_ALTERA == 1) ? read_pointer_n : read_pointer_q;
    else fifo_ram_read_address = '0;
```

AngelaGonzalezMarino: not sure if this will do something because if FPGA_EN=0 the module where this signal goes is not instantiated. But we can add it anyway

AngelaGonzalezMarino:
thanks!!

PR #2599, opened 2024-11-13
csr_regfile: Fix S-mode traps when H extension is enabled

niwis:
Supersedes #2587. If Hypervisor extension is enabled, the logic required to properly trap to S mode is currently excluded (Since #2346, the `else` block below the diff is no longer taken if `CVA6Cfg.RVH && !trap_to_v`). This means that supervisor traps are currently broken in the presence of H extension. Fix this by adjusting the `if` condition.


JeanRochCoulon:
image

niwis:
@JeanRochCoulon I have worked around the lint warning. Gate-level simulation still fails, but I can't see the reason for this

JeanRochCoulon:
HEllo @niwis    The gate simulation is failed, but it is a false fail. Do not worry about it. I will merge !   Thanks

PR #2600, opened 2024-11-13
Disable superscalar (Enable single issue)

JeanRochCoulon:
Since superscalar is currently not compatible with the execution of instructions with the Zcmp extension, I prefer to configure the 65x in single issue, and reserving the possibility of switching to superscalar if necessary.

Since 65x appears everywhere in the Gihub repository (RTL, Spike, documentation, collaborative projects, ...) and we do not know what the final option will be, I prefer to stick with the name "cv32a65x". We will continue to test the dual issue, Linux included, to avoid functional regressions.


JeanRochCoulon:
@cathales   Can you approve (again) ?

cathales (approved):

fatimasaleem:
@JeanRochCoulon In this PR do you want to enable ZCMP for this config as well?

JeanRochCoulon:
As cva6.py and compiler should be aligned to enable Zcmp, I plan to do it in another PR.

PR #2613, opened 2024-11-19
Altera opt 3

AngelaGonzalezMarino:
The third optimization for Altera FPGA is to move the register file to LUTRAM. Same as before, the reason why the optimization previously done for Xilinx is not working, is that in that case asynchronous RAM primitives are used, and Altera does not support asynchronous RAM. Therefore, this optimization consists in using synchronous RAM for the register file. 

The main changes to the existing code are: 

- Changes in ariane_regfile_fpga.sv file: The idea is the same as before, since synchronous RAM takes one clock cycle to read, we need to store the data when it is written, in case it is read right after. For this there is an auxiliary register that stores the last written data. On the read side, we need to identify if the data to be read is available in the RAM or if it is still in the auxiliary register (read after write). To compensate for the synchronous RAM delay the address is advanced one clock cycle. In this case there is a multiplexer in the output to select the block from where data is read, here we need to keep the read address for one clock cycle to select the right block when data is available. 

- Changes in issue_read_operands.sv file: adjust address to read from register file (when synchronous RAM is used reads take one cycle, so we advance the address). Since this address is an input, we need a new input port that brings the address in advance “issue_instr_i_prev”. 

- Changes in issue_stage.sv file: To connect the new input port that brings the address in advance “decoded_instr_i_prev”. 

- Changes in id_stage.sv file: To output the instruction to be issued before registering it (one clock cycle in advance). A new output port is needed for this “issue_entry_o_prev” 

- Changes in cva6.sv file: To connect the new output of the id_stage to the issue_stage to bring the address in advance to the register file (issue_entry_id_issue_prev)


JeanRochCoulon:
Hello @AngelaGonzalezMarino   I am invetigating why the FPGA is failed. It is maybe not related to your modification.

JeanRochCoulon:
As you have seen, Linux boot and gate simulation have just been fixed. We cross the finger for the Thales CI

AngelaGonzalezMarino:
> As you have seen, Linux boot and gate simulation have just been fixed. We cross the fingers for the Thales CI

 Smoke Generated test vcs-uvm fails, but I don't see errors in the logs. Do you have more visibility on this?

JeanRochCoulon:
Yes, log is limited to 10000 lines to prevent from server crash, but this test provide more lines... @valentinThomazic is fixing the issue.

valentinThomazic:
Hi @AngelaGonzalezMarino ! 
https://github.com/openhwgroup/cva6/pull/2636 has been merged, you should be able to see your errors if you rebase :)

PR #2634, opened 2024-11-26
CI fixes

valentinThomazic:
* Increase timeout for compliance testlist
* Switch to verilator for riscv-tests-v for faster sim
* fix reports for non tandem jobs


JeanRochCoulon:
Great fix !   We have a dream that one day the CI will be green !

PR #2635, opened 2024-11-27
Altera apu agilex7

AngelaGonzalezMarino:
This PR is adding the sources required to generate a bitstream for Altera FPGA.
- Altera proprietary IPs
- Avalon to AXI conversion modules
- Submodules to use PULP gpio in APU instead of Xilinx GPIO IP
- AXI data width adapter (non xilinx IP)
- adaptation of debug module to use with vJTAG IP (from Thales AVS)
- Top level of cva6 APU + cva6 peripherals adapted to Altera
- Pin mapping
- FPGA constraints

It does not include integration in the Makefile, that will be the last PR of the Altera support.


JeanRochCoulon:
Do you have time next CVA6 meeting to explain why all Altera techno files are added, this is huge and not sure it would be a good idea to overload the repository

AngelaGonzalezMarino:
> Do you have time next CVA6 meeting to explain why all Altera techno files are added, this is huge and not sure it would be a good idea to overload the repository 300klines are added

Sure, we can discuss next Tuesday. The files added are for Altera IPs needed in the APU. I think it could be possible to add only the IPs packaged, but I thought it was better to give the sources in case someone wants to modify (e.g. change frequency)

JeanRochCoulon:
I agree, it is better to provide the sources. In this case, why not creating a project with Altera library, the cva6 script would clone this project when needed
@MikeOpenHWGroup

AngelaGonzalezMarino:
> I agree, it is better to provide the sources. In this case, why not creating a project with Altera library, the cva6 script would clone this project when needed @MikeOpenHWGroup

I found the way to regenerate the sources from the .ip file, so it is much lighter now.

The remaining ones are files automatically generated by Quartus to connect the DDR IP with the AXI. Should these be moved to a library or is it OK to keep them here?

Thanks

PR #2640, opened 2024-11-28
use dcache_assoc_width

AngelaGonzalezMarino:
https://github.com/openhwgroup/cva6/blob/b718824e1ed0b6d2652f366308fa91f32fdfcf49/core/cache_subsystem/wt_dcache_missunit.sv#L202
Better to use the width parameter which already contemplates the case of 0 to avoid issues if associativity is set to 1
https://github.com/openhwgroup/cva6/blob/b718824e1ed0b6d2652f366308fa91f32fdfcf49/core/include/build_config_pkg.sv#L134


JeanRochCoulon:
Asap you rebase, I merge

AngelaGonzalezMarino:
Thanks!

PR #2646, opened 2024-12-03
add cv64a6_imafdch_sv39 config to cva6.py

Cra2yPierr0t:
Now that the Hypervisor Extension has been merged, how about making cv64a6_imafdch_sv39 available in cva6.py?


JeanRochCoulon:
If you would like to define a new configuration, the corresponding core/include/blabla_config_pkg.sv should be provided (otherwise it will not work) and I suggest you to share your motivation to define a new configuration.

Cra2yPierr0t:
@JeanRochCoulon
Thanks for the reply.

 If it is cv64a6_imafdch_sv39_config_pkg.sv, it already exists in core/include in the master branch.

PR #2457, opened 2024-08-20
[Pmp] Verif plan TXT

CoralieAllioux:
This PR aims at resolving task https://github.com/openhwgroup/cva6/issues/1447

Here's a first version of the Verification Plan of PMP, in txt format.
It is split in 2 files:
- verif/docs/VerifPlans/PMP/pmp_verif_plan.txt
- verif/docs/VerifPlans/PMP/pmp_verif_plan_features.txt

VPTool format is undergoing on our side.
We'd like to share already the content to get first feedbacks.

PR #2648, opened 2024-12-04
PMP Verif Plan and tests

OlivierBetschi:
Verification Plan provided in VP_TOOL for the  PMP. The verification plan should be complete, however only a partial set of the tests is available. This is not included in the CI but a bash script is available to run the test. Replacement of https://github.com/openhwgroup/cva6/pull/2457


JeanRochCoulon:
@zchamski, the ECE agreement fails due to the following error:   there is your name ?
image

JeanRochCoulon:
You should create a job in CI to run this test, that's a way to be sure they pass before merging.

JeanRochCoulon:
@valentinThomazic do you approve the DVPlan ?  and in case of need, help @OlivierBetschi to define a new job in CI.

valentinThomazic:
Some comments on the test script @OlivierBetschi : 

- linker script path is wrong 
- it would be great to be able to run all the tests simultaneously
- it is not needed to install all the test suites
- uvm_test arg does not exist on master cva6.py

valentinThomazic:
To get the result of the CI execution of the PMP tests directly on the PR pipeline report, please add the following job to the gitlab-ci:

<pre>
riscv-tests-p:
  extends:
    - .regress_test
  variables:
    DASHBOARD_JOB_TITLE: "[DEBUG! TO REMOVE FROM LIGHTS TESTS] PMP cv32a65x"
    DASHBOARD_JOB_DESCRIPTION: "PMP regression tests"
    DASHBOARD_SORT_INDEX: 0
    DASHBOARD_JOB_CATEGORY: "Basic"
    DV_SIMULATORS: "vcs-uvm"
    SPIKE_TANDEM: 1
  script: bash verif/regress/pmp_cv32a65x_tests.sh
  after_script: *simu_after_script
</pre>

OlivierBetschi:
Updated for review, but most only 2 tests are passing. I will have to debug this before the PR is merged. @valentinThomazic could you still have a look at the changes ? Thanks

.gitlab-ci.yml
```diff
@@ -399,6 +399,19 @@ riscv-tests-p:
...
 
+pmp_tests_cv32a65x:
+  extends:
+    - .fe_smoke_test
+  variables:
+    DASHBOARD_JOB_TITLE: "[DEBUG! TO REMOVE FROM LIGHTS TESTS] PMP cv32a65x"
+    DASHBOARD_JOB_DESCRIPTION: "PMP regression tests"
+    DASHBOARD_SORT_INDEX: 0
+    DASHBOARD_JOB_CATEGORY: "Basic"
+    DV_SIMULATORS: "vcs-uvm"
```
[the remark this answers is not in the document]
valentinThomazic: to have the logfiles on the dashboard

verif/sim/cva6.py
```diff
@@ -954,6 +961,27 @@ def load_config(args, cwd):
 
   args.spike_params = get_full_spike_param_args(args.spike_params) if args.spike_params else ""
 
+def edit_Makefile(UVM_TESTNAME="uvmt_cva6_firmware_test_c"):
```
valentinThomazic, line 964: I think it would be better to pass the uvm test to the Makefile as an env var instead of rewriting it at each simulation. This is how it is done for such things like the test path.
This is how you can do this : 

- add a `uvm_test `parameter to the `run_test` function in `cva6.py` (I admit that this function has too much parameters...)
- also add it in `get_iss_cmd` function 
- substitute it in the cmd like it is done for the elf path for instance
- add a `uvm_test `env var for the uvm targets in `verif/sim/cva6.yaml`
- set a default value in `verif/sim/Makefile` for the environment var `uvm_test`
- replace `UVM_TESTNAME=uvmt_qqchose` with `UVM_TESTNAME=$(uvm_test)` in `verif/sim/Makefile`

Do not hesitate to ask for more details if needed

valentinThomazic: I guess we could call each others if you think this could speed up things

valentinThomazic:
the uvm test is missing from the pr hence the fail

verif/sim/cva6.py
JeanRochCoulon: @OlivierBetschi I wonder why you use a different uvm test than "uvmt_cva6_firmware_test_c" ?
@AnouarZajni, can you help us to understand what is the best solution ?

OlivierBetschi: The different uvm_test came from the need to disable the cache for the PMP testing (that is my understanding at least).

OlivierBetschi:
Hello @JeanRochCoulon @valentinThomazic , I updated the PR : all tests are now passing with vcs-uvm on my side. The change of the uvm_test was not necessary after all. Just let me know what I should update for the gitlab ci.

valentinThomazic:
@OlivierBetschi there isn't any report on the dashboard because on your latest pipeline, there isn't any comparison made between spike and the rtl simulation. I see that you disabled spike tandem, this is what causes the issue.

.gitlab-ci.yml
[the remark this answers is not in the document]
JeanRochCoulon: Can you add "SPIKE_TANDEM: 1" to enable tandem and collect the ci results ?

JeanRochCoulon:
Hello @OlivierBetschi 
5 PMP tests are failed. As some features (as NAPOT) are not supported on cv32a65x configuration, I wonder whether some of these tests should be pass.

OlivierBetschi:
All the tests are self-checking and report a pass status, but the behavior of Spike and CVA6 are different. I thought we took the granularity into account. I would like to double check what is going on before making any change to the test though

OlivierBetschi:
Tests modified : comment in the regression scripts the tests using NAPOT mode and switch some of the address used in the test inside the DRAM range defined for spike

OlivierBetschi:
Github CI seem to fail due to changes from https://github.com/openhwgroup/cva6/pull/2711 . It looks like Verilator do not support multiple assert included inside each other. The mentioned PR also has the CI failing.

OlivierBetschi:
Hello @JeanRochCoulon , @valentinThomazic . PMP tests are now passing CI. I have commented all the tests for napot in the non-regression scripts. Should I remove the PMP tests from the gitlab CI too ?

OlivierBetschi:
I think the mismatch is on the expected value of pmpaddr0 at the end of the granularity test :
`CSR 3b0  Mismatch   [REF]: 0xfffffffe         [CORE]: 0xffffffff      `

The test configures pmpcfg0 to all 0 (i.e. OFF), write 0xFFFFFFFF into pmpaddr0 and check the read value of pmpaddr0, expecting 0xfffffffe (due to CVA6 PMP granularity). I think it comes from how the WARL nature of PMPADDR0 is managed.
In CVA6 rtl implementation, the value modification for the _read legal_ is done on the csr read portion, keeping the "true" value of 0xFFFFFFFF in the register. It looks like for SPIKE the legalization is done on the write portion and true value in the CSR register is the modified one.
It results that both SPIKE and CVA6 show the same behavior from a CSR read/write perspective, although handled differently internally. But then I think there is indeed a problem in CVA6; the wrong value for pmpaddr would be use instead of the legal one during PMP check.

If my understanding is correct, then the value modification done on csr reads for pmpaddr should also be done on the pmpaddr_o signal going out of the csr_regfile. The alternative is to modify the value during CSR write, but then if you change the mode of the associated pmpcfg, you cannot retrieve the original value back. So I am not sure what would be the proper way to handle this.

valentinThomazic:
Hello @OlivierBetschi 
I apologize for the late reply.
I think the issue comes from the RVFI which is directly connected to pmpaddr registers, without doing the legalization.
```
diff --git a/core/cva6_rvfi.sv b/core/cva6_rvfi.sv
index 4b4f863e..9e6b7b96 100644
--- a/core/cva6_rvfi.sv
+++ b/core/cva6_rvfi.sv
@@ -429,7 +429,7 @@ module cva6_rvfi
   genvar i;
   generate
     for (i = 0; i < 16; i++) begin
-      `CONNECT_RVFI_FULL(1'b1, pmpaddr[i], csr.pmpaddr_q[i][CVA6Cfg.PLEN-3:0])
+      `CONNECT_RVFI_FULL(1'b1, pmpaddr[i], {csr.pmpaddr_q[i][CVA6Cfg.PLEN-3:1], pmpcfg_q[i].addr_mode[1]})
     end
   endgenerate
   ;
```
This solves the issue.

valentinThomazic:
@OlivierBetschi, the issue with CVXIF is a master branch issue, unrelated with your PR.
Regarding cv32a65x configuration, everything seems fine now.
Before merging your PR, I was checking the tests that are unsupported by cv32a65x. 
I cannot get the 4th test (lsu_napot) to succeed with Spike only (of course I deleted every 65x-specific spike configuration). 
I was wondering if you manage to run it successfully wether it is with Spike or in RTL sim ?
(cv32a6_imac_sv32 supports NAPOT for instance)

OlivierBetschi:
@valentinThomazic I did not try to run the other tests with Spike only or other RTL. When we realized features like NAPOT were not supported we simply removed the tests from the script. As they were part of the verif plan and developed I wanted to provide them anyway. I think we tried them on the RTL of cv32a65x only so we expected them to fail.

valentinThomazic:
Other changes are ok for me

verif/sim/cva6.py
```diff
@@ -964,7 +964,6 @@ def load_config(args, cwd):
 
   args.spike_params = get_full_spike_param_args(args.spike_params) if args.spike_params else ""
 
-
```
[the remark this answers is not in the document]
OlivierBetschi: No problem, adding it back

valentinThomazic:
@JeanRochCoulon ok for me

JeanRochCoulon:
Will be committed as soon as it is rebased

JeanRochCoulon:
Thanks @OlivierBetschi for this contribution. It took a while to be merged, but right Now it is in the box ! 
A PMP job will be added in the CI by @valentinThomazic to collect the code coverage.

PR #2655, opened 2024-12-06
Added cv64a65x_config_pkg.sv for CVA6 64bits config. To be reviewed!

ludovicpion:
As discussed during CVA6 Project Meeting on 2024-12-03, I added the cv64a65x_config_pkg.sv for CVA6 64bits config. 

To be reviewed

### DIFF
--- a/core/include/cv64a65x_config_pkg.sv
+++ b/core/include/cv64a65x_config_pkg.sv
@@ -1,0 +1,101 @@
+// Copyright 2022 Thales DIS design services SAS
+//
+// Licensed under the Solderpad Hardware Licence, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// SPDX-License-Identifier: Apache-2.0 WITH SHL-2.0
+// You may obtain a copy of the License at https://solderpad.org/licenses/
+//
+// Original Author: Jean-Roch COULON - Thales
+
+package cva6_config_pkg;
+
+  localparam CVA6ConfigXlen = 64;
+
+  localparam CVA6ConfigRvfiTrace = 1;
+
+  localparam CVA6ConfigAxiIdWidth = 6;  // axi_pkg.sv
+  localparam CVA6ConfigAxiAddrWidth = 64;  // axi_pkg.sv
+  localparam CVA6ConfigAxiDataWidth = 64;  // axi_pkg.sv
+  localparam CVA6ConfigDataUserWidth = 64;  // axi_pkg.sv
+
+  localparam CVA6ConfigSuperscalarEn = 0;  // superscalar
+
+  localparam CVA6ConfigNrScoreboardEntries = 8;  // cvxif_pkg.sv
+
+  localparam config_pkg::cache_type_t CVA6ConfigDcacheType = config_pkg::HPDCACHE;
+
+  localparam config_pkg::cva6_user_cfg_t cva6_cfg = '{
+      XLEN: unsigned'(CVA6ConfigXlen),
+      FpgaEn: bit'(0),
+      NrCommitPorts: unsigned'(2),
+      AxiAddrWidth: unsigned'(CVA6ConfigAxiAddrWidth),
+      AxiDataWidth: unsigned'(CVA6ConfigAxiDataWidth),
+      AxiIdWidth: unsigned'(CVA6ConfigAxiIdWidth),
+      AxiUserWidth: unsigned'(CVA6ConfigDataUserWidth),
+      MemTidWidth: unsigned'(CVA6ConfigAxiIdWidth),
+      NrLoadBufEntries: unsigned'(8),
+      RVF: bit'(1),
+      RVD: bit'(1),
+      XF16: bit'(0),
+      XF16ALT: bit'(0),
+      XF8: bit'(0),
+      RVA: bit'(1),
+      RVB: bit'(1),
+      RVV: bit'(0),
+      RVC: bit'(1),
+      RVH: bit'(0),
+      RVZCB: bit'(1),
+      RVZCMP: bit'(0),
+      XFVec: bit'(0),
+      CvxifEn: bit'(1),
+      RVZiCond: bit'(1),
+      NrScoreboardEntries: unsigned'(CVA6ConfigNrScoreboardEntries),
+      PerfCounterEn: bit'(1),
+      MmuPresent: bit'(1),
+      RVS: bit'(1),
+      RVU: bit'(1),
+      HaltAddress: 64'h800,
+      ExceptionAddress: 64'h808,
+      RASDepth: unsigned'(2),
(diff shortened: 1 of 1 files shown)


ludovicpion:
Hi,
Obviously the name of this config will be changed to be coherent with the actual parameters.

Any other feedback before making the change?

Thanks

JeanRochCoulon:
Yes. We highly suggest to remove all the localparams from this file. Because this produces RTL issues: CVA6config can be override from outside (when instantiating the CVA6) while localparam cannot be redefined.

ludovicpion:
Ok, no problem to remove localparam. 

Current localparam were just based on the format of the cv32a65x config pkg file.

jquevremont:
> "65" means superscalar is enabled, which is not the case in this new configuration. Ok to create a new configuration, but the name need to be updated and approved by OpenHW. @jquevremont @Floadv298

Configuration names are vetted by the Cores TG, either in monthly meetings or on the Mattermost channel.

With what I read, the likely name would be CV64A60AX (the second A for Application = MMU, X for CV-X-IF). I think it collides with what @zarubaf has in mind for his company's configuration (but it's not vetted). The first action would for @ludovicpion and @zarubaf to get in touch together and identify the differences between their configurations.

core/include/cv64a65x_config_pkg.sv
```diff
@@ -0,0 +1,101 @@
...
+  localparam CVA6ConfigDataUserWidth = 64;  // axi_pkg.sv
+
+  localparam CVA6ConfigSuperscalarEn = 0;  // superscalar
+
+  localparam CVA6ConfigNrScoreboardEntries = 8;  // cvxif_pkg.sv
+
+  localparam config_pkg::cache_type_t CVA6ConfigDcacheType = config_pkg::HPDCACHE;
+
+  localparam config_pkg::cva6_user_cfg_t cva6_cfg = '{
+      XLEN: unsigned'(CVA6ConfigXlen),
```
MikeOpenHWGroup, line 28: In [cv32a65x_config_pkg.sv](https://github.com/openhwgroup/cva6/blob/5ff6b2d32e890403e7e0342ce274fa0409fc7108/core/include/cv32a65x_config_pkg.sv#L23) there is a parameter called VLEN (Virtual address length) defined as well.  It is a bit of a surprise that the CV32A65X should define anything about virtual addressing, but I would expect to see it defined for an Application-class core such as the CV64A65X.

ludovicpion: Maybe it is a parameter that we have missed. I will check.

Thanks for the feedback, it is useful.

core/include/cv64a65x_config_pkg.sv
```diff
@@ -0,0 +1,101 @@
...
+      FpgaEn: bit'(0),
+      NrCommitPorts: unsigned'(2),
+      AxiAddrWidth: unsigned'(CVA6ConfigAxiAddrWidth),
+      AxiDataWidth: unsigned'(CVA6ConfigAxiDataWidth),
+      AxiIdWidth: unsigned'(CVA6ConfigAxiIdWidth),
+      AxiUserWidth: unsigned'(CVA6ConfigDataUserWidth),
+      MemTidWidth: unsigned'(CVA6ConfigAxiIdWidth),
+      NrLoadBufEntries: unsigned'(8),
+      RVF: bit'(1),
```
MikeOpenHWGroup, line 37: Happy to see the F and D ISAs included in this core!  I would strongly encourage you to update the floating-point unit the in CVA6 to the latest version of [cvfpu](https://github.com/openhwgroup/cvfpu/tree/master) as the floating-point unit currently in CVA6 is very out of date and can be expected to have many known and unknown bugs.   A couple of additional things to note:
1. The F ISA of cvfpu has been extensively verified by v1.83. of CV32E40P.  Please check out the CV32E40P user Manual to see which parameters of the cvfpu have been verified.
2. Support for the D ISA has never been verified in an OpenHW core.  I do not know the status of this feature in cvfpu, but it is safe to assume that the CV64A65X will have a lot of work to do for double-precision floating point verification.

core/include/cv64a65x_config_pkg.sv
```diff
@@ -0,0 +1,101 @@
...
+      RVD: bit'(1),
+      XF16: bit'(0),
+      XF16ALT: bit'(0),
+      XF8: bit'(0),
+      RVA: bit'(1),
```
MikeOpenHWGroup, line 42: Happy to see this!

core/include/cv64a65x_config_pkg.sv
```diff
@@ -0,0 +1,101 @@
...
+      NonIdempotentAddrBase: 1024'({64'b0, 64'b0}),
+      NonIdempotentLength: 1024'({64'b0, 64'b0}),
+      NrExecuteRegionRules: unsigned'(3),
+      ExecuteRegionAddrBase: 1024'({64'h8000_0000, 64'h1_0000, 64'h0}),
+      ExecuteRegionLength: 1024'({64'h40000000, 64'h10000, 64'h1000}),
+      NrCachedRegionRules: unsigned'(1),
+      CachedRegionAddrBase: 1024'({64'h8000_0000}),
+      CachedRegionLength: 1024'({64'h40000000}),
+      MaxOutstandingStores: unsigned'(7),
+      DebugEn: bit'(1),
```
MikeOpenHWGroup, line 79: Again, the Debug feature is not well verified.

core/include/cv64a65x_config_pkg.sv
JeanRochCoulon: In CPU core, addresses are virtual.
if no MMU, PLEN address = VLEN
if MMU, PLEN address = translation(VLEN)

MikeOpenHWGroup (commented):
This is great @ludovicpion, thanks for this. I am happy with it, but will leave it up to either @JeanRochCoulon or @jquevremont to approve.  I've added comments directly to specific lines in the pkg file and will also make the following comments and queries:
1. I see that you have not set `SuperscalarEN`.  If we follow our current naming strategy, that implies that this core should be the CV64A60X.
2. Are you interested in defining a core that aligns with one of the RVI Profiles?  At this point, this config is _close_ to the `RVB23U64`.  To get there we'll need to add Zicsr, Zicntr, Zihpm, Ziccif, Ziccrse, Ziccamoa, Za64rs and Zihintpause (whew!).  As the CVA6 predates the definition of RVI Profiles, the [config_pkg.sv](https://github.com/openhwgroup/cva6/blob/5ff6b2d32e890403e7e0342ce274fa0409fc7108/core/include/config_pkg.sv#L68) does not have variables defined for many of these Zi* features.   Please give it some thought - at a minimum it would be useful to add these variables to the config package and set them to zero.

PR #2667, opened 2024-12-13
bug fix: canonical check on virtual address for data accesses

fatimasaleem:
As mentioned in the spec, we need to perform a canonical check on the virtual address for instruction fetch, load, and store. If the check fails, it will cause the page-fault exception. 

image

This PR fixes the above two:

1) changes INSTR_ACCESS_FAULT to INSTR_PAGE_FAULT
2) Adding virtual address check on data accesses as well


JeanRochCoulon:
Great finding !
As your branch is protected, I cannot rebase before merging. Can you do rebase ?

PR #2672, opened 2024-12-17
Proposal for a CVA6 64 bits configuration file updated

ludovicpion:
Hello,

Based on  the feedback received on PR https://github.com/openhwgroup/cva6/pull/2655, I have made the following changes:
- changed the name of the file to cv64aTBD_config_pkg.sv. The TBD will be changed later to an fully accurate name. In the meantime, I would prefer to have feedback of the content of the file and so the parameters values.
- I have removed all the localparam but CVA6ConfigSuperscalarEn because I did not find which parameter of cva6_user_cfg_t is related to superscalar features
- Regarding PLEN and VLEN paremeters, I would be happy to explicitly set them but, as for superscalar,  I did not find these parameters within cva6_user_cfg_t. Only XLEN is visible and so set to 64.

Feedback and/or advices are welcomed!

Regards,
Ludovic


ludovicpion:
> I fear this PR will not work. Because for instance, the AXI is not declared, have you tried to use it to simulate a pattern?

Hello,

As I do not know CVA6 well, so the idea is to have feedback on the content of the configuration , not to have the PR accepted. For example if there is incoherent parameters values... 
It was not clearly said in the message coming with the PR ;-)

So regarding for you comment (thanks),
- no I did not use this config to simulate a pattern. How could I do it? Which TB?
- What do you neam by "_the AXI is not declared_"? 
=> I took cv32a65x config pkg file as a starting point. I have removed as much as localparam as possible and updated parameters values based on what we (CEA) think our CVA6 64 bits config should be.
=> For the AXI, there are the following parmaters:

-        AxiAddrWidth: unsigned'(64),
-       AxiDataWidth: unsigned'(64),
-       AxiIdWidth: unsigned'(6),
-       AxiUserWidth: unsigned'(32),

Are there other parameters to configure?

Regards,
Ludovic

JeanRochCoulon:
A few feedbacks:
- if the PR must not be merged, it is better to declare it as "draft"
- To simulate a test on another configuration, follow the README.md.
- Then modify the target option of cva6.py by replacing it by your configuration. This should not work, then fix the config_pkg.sv to make it functional.
Cheers

ludovicpion:
Thanks for the info

ludovicpion:
Will do an other draft PR if needed.

Issue #2665, opened 2024-12-12
[BUG] PMPCFG : Dead code in pmp module


AyoubJalali:
### Is there an existing CVA6 bug for this?

- [X] I have searched the existing bug issues

### Bug Description
Hello,

 I noticed that the write of the value (2'b10 == NA4) in the pmpcfg.A is ignore for all the CVA6 configuration, That's means we don't support NA4, that's why we don't NA4 code in the pmp module.

We had some code coverage holes : 
![image](https://github.com/user-attachments/assets/2376a442-b830-41c6-a65a-66c556ee9d67)

@OlivierBetschi what do you think ?

AyoubJalali:
> ### Is there an existing CVA6 bug for this?
> * [x]  I have searched the existing bug issues
> 
> ### Bug Description
> Hello,
> 
> I noticed that the write of the value (2'b10 == NA4) in the pmpcfg.A is ignore for all the CVA6 configuration, That's means we don't support NA4, that's why we don't NA4 code in the pmp module.
> 
we had some code coverage holes in that : 

![image](https://github.com/user-attachments/assets/62e3b989-cb8b-43dc-93f8-dd4602d50aa8)

OlivierBetschi:
According to https://docs.openhwgroup.org/projects/cva6-user-manual/01_cva6_user/PMP.html the NA4 mode is not selectable because of the granularity of the PMP being 1. This is confirmed in the RISC-V privileged ISA (3.7.1 Physical memory protection CSRs) : _When G ≥ 1, the NA4 mode is not selectable_. 

So NA4 not being supported makes sense. Then the question if why the granularity is set to 1, as I don't see anything in the CVA6 requirements ( https://docs.openhwgroup.org/projects/cva6-user-manual/02_cva6_requirements/cva6_requirements_specification.html ). I don't know who could answer that ?

AyoubJalali:
@zchamski we had some discussion on this before

JeanRochCoulon:
To my mind, G can be equal to zero for cva6. I am submitting a PR which adds a new config parameter called PMPG (PMP granularity) to define G.
For 60x and 65x, G=1 which means granularity = 8 Bytes and NAPOT is not supported.
For other configurations, G=0 which means granularity = 4 Bytes and NAPOT can be supported.
This PR has been approved Today and the CI is green.

@zchamski @ASintzoff @AyoubJalali @OlivierBetschi @moimfeld please warm me if you are not ok with this.

[...]

PR #2677, opened 2024-12-19
fix pmpaddr

### LINKED ISSUE #2657: [BUG] [CV32A65X] Writes into PMPADDR registers do not follow specification

### Bug Description

Since the CV32A65X specification has been updated to make bit 0 of `pmpaddrN` registers read-only zero (see https://github.com/openhwgroup/cva6/pull/2651), the implementation of `pmpaddrN` CSR writes is not in line with the spec.  The incriminated code is https://github.com/openhwgroup/cva6/blob/master/core/csr_regfile.sv#L1710C1-L1713C71.

JeanRochCoulon:
fix #2657
define a new PMPG parameter to set the PMP granularity: G=1 (8 Bytes) for 65x and 60x


Moschn:
see https://github.com/openhwgroup/cva6/issues/2665#issuecomment-2552952549

PR #2684, opened 2025-01-03
Update rvfi_tracer and cva6.py

AyoubJalali:
This MR update rvfi_tracer to help the team debug, also I add the capability of randomizing the sv_seed by default


verif/sim/cva6.py
```diff
@@ -775,7 +775,7 @@ def parse_args(cwd):
                       help="switch AXI agent mode: yes for Active, no for Passive")
   parser.add_argument("--gen_sv_seed", type=int, default=0,
                       help="Run test N times with random seed")
-  parser.add_argument("--sv_seed", type=str, default="1",
+  parser.add_argument("--sv_seed", type=str, default=str(random.getrandbits(31)),
```
JeanRochCoulon, line 778: If the seed is random, the CI tests will be executed with different seeds. This means the CI will not be deterministic. This can be a problem when executing Dhrystone or Coremark tests for instance for which we monitor cycle number.
Moreover some CI tests can have random behavior in relation to the seed. Debug can be more difficult.

AyoubJalali: You can fix you're sv_seed using cva6.py option --sv_seed = n , by default sv_seed is always 1 and this limit finding bugs and use 100% of the UVM advantage (randomization), for now in CI all tests are running with only one sv_seed.

But we should store uvm seed in a specific file and store them in the artifacts

Issue #2400, opened 2024-07-25
[BUG] PMP CSR : NAPOT MODE isn't supported


AyoubJalali:
### Is there an existing CVA6 bug for this?

- [X] I have searched the existing bug issues

### Bug Description

Hello, according to the [cv32a65x](https://docs.openhwgroup.org/projects/cva6-user-manual/04_cv32a65x/riscv/priv.html#_machine_status_mstatus_and_mstatush_registers), in the pmp entries the NA4 & NAPOT modes isn't supported, but in the RTL is only excluding NA4 from the add_mode logic in [csr_regfile.sv](https://github.com/openhwgroup/cva6/blob/fd489a16fbfbb791d598c4a61afa765267ab12ff/core/csr_regfile.sv#L2690)

is it a bug in the RTL or the Specification ?

AyoubJalali:
I think this is a RTL bug

zchamski:
Results of a careful review of the spec and RTL with @ASintzoff:
- the spec of CV32A65X **does not specify** the actual WARL behavior upon writing an invalid (illegal/reserved) multi-bit value for RWX and A fields; this is required given the definition of WARL at https://cva6.readthedocs.io/en/latest/04_cv32a65x/riscv/priv.html#_write_any_values_reads_legal_values_warl;
- the RTL implements 'no change on reserved' policy for the aggregate RWX field;
- the RTL implements 'no change on illegal' policy for A fields but **does** allow `A == NAPOT` which conflicts with the CV32A65X spec  supporting only `A == OFF` and `A == TOR`.

The RTL needs to be fixed for CV32A65X but without affecting other configurations.

PR #2685, opened 2025-01-06
Fix #2665 #2400 #2657

### LINKED ISSUE #2665: [BUG] PMPCFG : Dead code in pmp module

JeanRochCoulon:
* Fix 2400 by defining a new CVA6 config parameter, called PmpNapotEn, to disable NAPOT PMP mode for 60x and 65x.
* Fix 2665 by removing code related to NA4. This allows to increase code coverage.
* Fix 2657 bit 0 of PMPAddr is read only zero for 65x and 60x.


JeanRochCoulon:
@AyoubJalali   As you can see 3 GitHub issues should be fixed. What do you think ? Do you confirm it ?

AyoubJalali:
> @AyoubJalali As you can see 3 GitHub issues should be fixed. What do you think ? Do you confirm it ?

YES for #2400 and #2665, but I can't see any changes related to fixing #2657.

PR #2729, opened 2025-01-22
Spike : mtvec doesn't support vectored mode for cv32a65x

AyoubJalali:
This is a spike fix for cv32a65x config


JeanRochCoulon:
Great @AyoubJalali

PR #2749, opened 2025-01-28
Fix RVFI rs1/rs2 len from VLEN to XLEN

JeanRochCoulon:
RVFI rs1 and rs2 operands were VLEN, it has been fixed to be XLEN.


JeanRochCoulon:
Approved internally by @ASintzoff

PR #2764, opened 2025-02-10
Simplify the Verilog "inside"

flaviens:
Hi! Some tools like morty struggle with this expression. I suggest this very simple rewrite. No need for fancy constructs here.
Thanks!
Flavien

### DIFF
--- a/core/cva6.sv
+++ b/core/cva6.sv
@@ -1366,10 +1366,11 @@ module cva6
         .inval_valid_i     (inval_valid),
         .inval_ready_o     (inval_ready)
     );
-  end else if (CVA6Cfg.DCacheType inside {
-      config_pkg::HPDCACHE_WT,
-      config_pkg::HPDCACHE_WB,
-      config_pkg::HPDCACHE_WT_WB})
+  end else if (
+        CVA6Cfg.DCacheType == config_pkg::HPDCACHE_WT ||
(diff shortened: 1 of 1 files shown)


flaviens:
Yes very likely. Some of my PRs already got merged in the last couple of years.

JeanRochCoulon:
Your email address is not recognized by the bot. You must sign and submit PR with the same email address. That's maybe the issue.

flaviens:
Indeed probably hadn't done it for this email address. Done now. Thanks!

cathales:
Hello,

For reference, here is the related issue: https://github.com/pulp-platform/morty/issues/61

Morty is not officially supported by CVA6 (yet?). `inside` is a construct implemented by most of the tools.
I am not especially in favor of changing our RTL to comply with tools that we do not support.

I think a more constructive option would be to send a PR to Morty so that it supports the `inside` operator.

cathales (approved):
Okay, new fact to consider: according to IEEE Std 1800-2017, an `inside_expression` is not a `constant_expression` (which is absurd IMHO) so it cannot be used in this context.

Your PR improves our compliance with the standard so I approve it.

flaviens:
Strange. I signed it, used the same email as in the commit, and the Eclipse website seems happy. Should I sign with another email?
image

PR #2769, opened 2025-02-12
[Skip CI] Update to CONTRIBUTING.md

jquevremont:
Writing some rules that were previously implicit.


cathales:
> To ease maintenance, all common code lines shall exist only once.

Related reading, for reference: https://en.wikipedia.org/wiki/Don%27t_repeat_yourself

JeanRochCoulon:
Thanks @jquevremont for this add-on !

PR #2770, opened 2025-02-13
cva6_icache: Fix formatting

niwis:
Run verible verilog format to fix upstream formatting.

Perhaps it would be a good idea to let CI fail if verible catches mismatches to prevent them from being accidentally merged. I can look into that.


cathales:
Hello Nils,

Thanks for your feedback.

#2771 Will mark the verible job as failed when formatting suggestions are emitted.
I cannot remember if this check must pass before a PR is merged; it depends on the configuration of the CVA6 GitHub project, which I do not have access to.

niwis:
Thanks @cathales! At least this will show a red X. The PR suggestions alone seem to get lost too easily in a long discussion.

cathales (approved):

PR #2772, opened 2025-02-14
Add parameter type to define which coprocessor is instantiated on CVXIF

Gchauvon:
Add parameter `CoproType` to select which coprocessor to instantiate when `CvxifEn == 1`


JeanRochCoulon:
CoproType is a platform parameter, not a processor parameter.

Gchauvon:
It extends the supported ISA and must match Spike configuration. I believe in a way it is a CPU parameter.

JeanRochCoulon:
Ready to merge it into cv32a60x branch (and not master)

PR #2776, opened 2025-02-19
Simplify the Verilog "inside" (from @flaviens)

JeanRochCoulon:
Hi! Some tools like morty struggle with this expression. I suggest this very simple rewrite. No need for fancy constructs here. Thanks @flaviens for this contribution


cathales (approved):
This change is needed because, according to IEEE Std 1800-2017, an `inside_expression` is not a `constant_expression`.

For reference, see initial PR #2764

PR #2791, opened 2025-02-26
Fix access to tp register for LLVM in syscalls.c

math-gout:
The old syntax for accessing the tp register was not recognised by LLVM. This PR fixes that.

https://gcc.gnu.org/onlinedocs/gcc/Local-Register-Variables.html
https://gcc.gnu.org/onlinedocs/gcc/Global-Register-Variables.html


JeanRochCoulon:
I updated the bench results to make the CI pass, but why bench are faster?  This modification should not impact them so much

PR #2792, opened 2025-02-26
change struct assignment syntax for quartus

AngelaGonzalezMarino:
Since the syntax of this assignment was changed in https://github.com/openhwgroup/cva6/pull/2772, Quartus gives an error when generating a bitstream with CVXIF enabled
The proposed change works for both Vivado and Quartus

### DIFF
--- a/corev_apu/src/ariane.sv
+++ b/corev_apu/src/ariane.sv
@@ -133,7 +133,7 @@ module ariane import ariane_pkg::*; #(
         .cvxif_resp_o         ( cvxif_resp                     )
       );
     end else begin: gen_COPRO_NONE
-      assign cvxif_resp = '{compressed_ready: 1'b1, issue_ready: 1'b1, register_ready: 1'b1, default: '0};
+      assign cvxif_resp = '{1'b1, x_compressed_resp_t'(0), 1'b1, x_issue_resp_t'(0), 1'b1, 1'b0 ,x_result_t'(0)};
     end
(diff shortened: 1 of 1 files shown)


Gchauvon:
Which version of Quartus and Vivado are you using ?

It is not the only place this syntax is used. 
Does the line raise the same error ? https://github.com/openhwgroup/cva6/blob/ab89beaebbcb5ade0a9f3dd51f0f79ffadad5bbd/core/csr_regfile.sv#L2550

AngelaGonzalezMarino:
> Which version of Quartus and Vivado are you using ?
> 
> It is not the only place this syntax is used. Does the line raise the same error ?
> 
> https://github.com/openhwgroup/cva6/blob/ab89beaebbcb5ade0a9f3dd51f0f79ffadad5bbd/core/csr_regfile.sv#L2550

We use Vivado 2022.2 and Quartus Prime 24.1

You are right, that is the same syntax and Quartus does not complain (DebugEn is set to 1 in cv64a6_imafdc_sv39_config_pkg so I guess we should see it).

Then maybe the issue is with the functions used to build the type? The example in the csr is simpler in that sense

> Which version of Quartus and Vivado are you using ?
> 
> It is not the only place this syntax is used. Does the line raise the same error ?
> 
> https://github.com/openhwgroup/cva6/blob/ab89beaebbcb5ade0a9f3dd51f0f79ffadad5bbd/core/csr_regfile.sv#L2550

PR #2793, opened 2025-02-26
bp: add BHT with private history

ricted98:
This PR adds a new two-level BHT predictor with private history, as depicted below. The new `BPType` parameters allow choosing between the original BHT and the new one.

immagine


core/frontend/bht2lvl.sv
```diff
@@ -0,0 +1,122 @@
+module bht2lvl #(
```
JeanRochCoulon, line 1: Add the header to declare license, author,...

core/include/cv32a6_embedded_config_pkg_deprecated.sv
```diff
@@ -62,6 +62,8 @@ package cva6_config_pkg;
   localparam CVA6ConfigRASDepth = 2;
   localparam CVA6ConfigBTBEntries = 0;
   localparam CVA6ConfigBHTEntries = 32;
+  localparam CVA6ConfigBHTHist = 3;
```
JeanRochCoulon, line 65: Better to not use localparam, this can generate bugs because localparam is static and parameter is dynamic (can be configured from outside).

ricted98: Are you referring only to the `CVA6ConfigBHTHist` parameter?

JeanRochCoulon: The solution is to proceed as in cv32a60x configuration: remove localparam declaration and set it as seen below:
image

ricted98: However, ```BHTHist``` should be a tuneable parameter like the remaining ones. My understanding is that cv32a60x and cv32a65x are hardcoded, while remaining config packages have the ```localparam``` pattern to tune them.

core/include/cv32a60x_config_pkg.sv
```diff
@@ -62,7 +62,9 @@ package cva6_config_pkg;
       ExceptionAddress: 64'h808,
       RASDepth: unsigned'(2),
       BTBEntries: unsigned'(0),
+      BPType: config_pkg::PH_BHT,
       BHTEntries: unsigned'(32),
+      BHTHist: unsigned'(3),
```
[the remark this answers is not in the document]
ricted98: Yes. I just wanted to wait for you to see the CI. I will update them now.

core/include/cv32a6_embedded_config_pkg_deprecated.sv
JeanRochCoulon: Globally, it would be good to remove all the localparam, from 60x/65x/other configurations. I made it for 60x and 65x. It would be better to avoid using localparam for others configurations too.

ricted98: Done!

core/frontend/bht2lvl.sv
```diff
@@ -0,0 +1,135 @@
...
+// this License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+// CONDITIONS OF ANY KIND, either express or implied. See the License for the
+// specific language governing permissions and limitations under the License.
+//
+// Original author: Gianmarco Ottavi, University of Bologna
+// Description: Private history BHT
+
+module bht2lvl #(
+    parameter config_pkg::cva6_cfg_t CVA6Cfg = config_pkg::cva6_cfg_empty,
+    parameter int unsigned NR_ENTRIES = 256,
```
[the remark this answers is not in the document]
ricted98: Done!

JeanRochCoulon:
Thanks @ricted98 
BTW could you add the picture of the new BHT into the following directory ? 
https://github.com/openhwgroup/cva6/tree/master/docs/03_cva6_design/_static

ricted98:
Thanks for the reviews. I will add the image as suggested.

PR #2799, opened 2025-02-27
Improve Code Coverage in Frontend

JeanRochCoulon:
The last instruction cannot be a 32 bit instruction when this instruction is not word aligned (in single or dual issue mode).
This cannot be (easily) conditioned. One solution was to exclude the lines, but as it is not so user friendly, Côme proposed to rewrite the code to hide the uncovered case.


AyoubJalali:
LGTM for CC

PR #2820, opened 2025-03-11
Add more granularity to kgates report

MaxCThales:
Changing the type of Kgates from int to float in order to add more granularity on the report (usefull when the difference is under 1k Gates)


JeanRochCoulon:
Great !
Welcome to the new contributor !

PR #2832, opened 2025-03-17
Improve Code Coverage

JeanRochCoulon:
Improve code condition knowing that:
- ex_rvalid is always 0 in cv32a60x : MMU and PMP are not present
- PRIV_LVL_M is always 1 in cv32a60x
- commi_ack and exception can be both at 1 if and only if CVA6Cfg.SpeculativeSb is 1
- Frontend generates exception only if MMU or PMP is present


ASintzoff (approved):
LGTM

PR #2833, opened 2025-03-17
Initialize compressed related signals in id_stage when RVC is disabled

### LINKED ISSUE #2819: [BUG] <32 bit bitstream for FPGA config fails>

### Bug Description

After https://github.com/openhwgroup/cva6/commit/2ef1c1b1fca419354920c5487293bc605294904e the bitstream  using FPGA config fails (no message from bootrom).
If the C extension is enabled, everything seems OK.
@Gchauvon do you have any idea why not using the C extension would fail after this change?

Gchauvon:
Add `else` case to initialize signals going into decoder.
Should fix #2819

### DIFF
--- a/core/id_stage.sv
+++ b/core/id_stage.sv
@@ -269,6 +269,13 @@ module id_stage #(
     end else begin
       assign stall_instr_fetch[0] = stall_macro_deco;
     end
+  end else begin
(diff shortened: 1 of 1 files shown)


JeanRochCoulon:
Bien !

Issue #2831, opened 2025-03-17
[BUG] mtinst corrupted


KatCe:
### Is there an existing CVA6 bug for this?

- [x] I have searched the existing bug issues

### Bug Description

Hello,
I have noticed an unexpected behavior.
The signal x_exception_o in cvxif_fu has several unconnected fields, which is partially intentional (at least for tval) according to the specification.
In certain scenarios, and with the hypervisor extension enabled, it is possible that these undriven fields are read and the CSR mtinst is written with these arbitrary values.
I have created a small test bench here: https://github.com/KatCe/cva6/tree/bug_mtinst_corruption/bug_simulations/x_exception_mtinst
In this test bench, the bug is triggered by an unexpected bus response that also triggers an assertion:

	** Fatal: Trying to pop data although the FIFO is empty.
	#    Time: 25 ns Started: 25 ns  Scope: tb_top.i_ariane.gen_cache_wt.i_cache_subsystem.i_adapter.i_rd_icache_id.empty_read File:../core/cva6_fifo_v3.sv Line: 228

To model what would happen in silicon I have initialized the tinst field of x_exception_o with an arbitrary value for demonstration (https://github.com/KatCe/cva6/blob/bug_mtinst_corruption/core/cvxif_fu.sv#66) and converted the assertion to a warning (https://github.com/KatCe/cva6/blob/bug_mtinst_corruption/core/cva6_fifo_v3.sv#228).
In that case we see the arbitrary value written to the mtinst CSR.

While this exact scenario can only happen if cva6 is integrated with another component that does not follow the bus protocol, I believe that cva6 should not rely on a bug free external world. Furthermore, a fault injection attack could trigger this CSR corruption.

I have not yet investigated whether this corruption can also be triggered by other scenarios.

Gchauvon:
Does #2844 fix this issue ? 

If it does please close it.

BR, Guillaume

PR #2844, opened 2025-03-19
[CVXIF] Initialize exception fields for RVH

Gchauvon:
Following what was done in branch_unit, I set up a default value for hypervisor exception fields in `cvxif_fu`.
Should fix issue #2831


JeanRochCoulon:
@Gchauvon  This PR fixes the master branch, do we need to fix cv32a60x ?

PR #2869, opened 2025-03-26
CODEOWNERS: remove `zarubaf` from global owners

zarubaf:
I haven't reviewed anything in quite a while so I believe its best for me to depart the `CODEOWNER`s file. I'll still be watching the repo and any mentions, and of course,e we are using CVA6 heavily internally so still supporting the project!


JeanRochCoulon:
Hello @zarubaf 
Snif... But happy to know you are always involved !

PR #2805, opened 2025-03-03
issue_stage: add register renaming

ricted98:
This PR introduces a register renaming scheme to eliminate WAW hazards, as depicted below. Note that this change is not toggled by a parameter, rather it is a redesign of a portion of the issue logic.

immagine

PR #2881, opened 2025-03-28
WAW hazards elimination

cathales:
This PR introduces a new RAW hazard detection mechanism to eliminate WAW hazards in CVA6 issue stage.

It first checks for hazards in all scoreboard entries in parallel.
Then it filters found hazards before vs after the current issue pointer.
It then finds the index of the last hazard before (resp. after) the issue pointer.
Finally, it gives precedence to a hazard before the issue pointer over the one after the issue pointer.

image


cfuguet (approved):
It looks good to me. This is a nice contribution.

I was expecting the number of gates to increase because you are adding one priority encoder and some extra comparators to determine which entries are before/after the issue pointer. However, I was looking the Thales-CIs, and the number of gates actually decreased...

cathales:
Thanks César,

This new mechanism replaces the previous one and is more compact, hence the area reduction.

Cc @ricted98

cfuguet:
Ok, but you agree with me that is kind of surprising... it is probably because there was a problem with the previous implementation. Otherwise, blocking the instruction issue during a WAW shall simplify the RAW logic.... You are sure that the newest value can come from either the RF or only one of the entries of the SB. No need for priority encoders or less or equal comparators for the issue pointer

cfuguet:
Yes, I understand why you need the priority encoders and the comparators for disambiguation for this new implementation.

However, I do not understand why the previous implementation needed to be that complex... Normally, as you can only have one possible candidate to bypass the data in case of RAW, then the logic shall be simpler. But it looks like it was not the case. That is why I was saying that, in my opinion, there was a problem with the previous implementation.

Anyway, the only thing we need to be careful with this new implementation is the timing. But I have not recently synthesized the CVA6 to know where is the critical path.

cathales:
The critical path did not involve this part of the pipeline in my synthesis. (It seemed to be related to the multiplier, in the execute stage)

cathales:
This new version makes renaming mandatory, thus removing the option from the CVA6 configuration.

It also fixes Linux boot (there was an issue with Vivado).
Thanks a lot to @ricted98 for #2805 that we used as a reference to fix Linux boot.

---

GitHub CI crashed but it seems to be unrelated:

```
Err:9 http://download.opensuse.org/repositories/home:/phiwag:/edatools/xUbuntu_20.04  InRelease
  The following signatures couldn't be verified because the public key is not available: NO_PUBKEY BB95FA80F959D90B
Reading package lists...
W: GPG error: http://download.opensuse.org/repositories/home:/phiwag:/edatools/xUbuntu_20.04  InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY BB95FA80F959D90B
E: The repository 'http://download.opensuse.org/repositories/home:/phiwag:/edatools/xUbuntu_20.04  InRelease' is not signed.
```

cathales:
2.6 kGates saved

JeanRochCoulon:
Many thanks to @ricted98 and @cathales for having taken time to compare the two Renaming solutions. After PPA compare, this solution has been selected to be merged into cva6.
Waiting green light to merge...

cfuguet:
Out of curiosity, I concur that this solution has a better area and good timing for a reduced number of entries in the scoreboard. But does it scale well when the number of entries increases ? I do not know if you made the test with for example 16 or 32 entries in the scoreboard. The timing of the priority encoder in this solution would degrade linearly with the number of entries. In the other solution, I would expect a log2 increase in timing (but still worse area).

cathales:
Hello,

We have not measured this.
With a naive implementation (`idx = 0; for i in 0..n { if valid[i] { idx = i } }`, timing would indeed be O(n), but if I understand `rr_arb_tree` correctly, it seems to be optimized to be O(log n).

I guess that even with the naive solution, the synthesis tool could probably optimize things to switch to the O(log n) solution if needed? This could be "only" balancing a binary tree.

cfuguet:
Ok, you are right 👍

PR #2882, opened 2025-03-28
Fix #1625: Allow unmapped memory accesses in Spike for cv32a60x/cv32a65x.

### LINKED ISSUE #1625: [Spike] : Mismatch of address mapping between spike & RTL

### Bug Description

Hello, 
So after the merge of spike tandem, one of the directed tests that I wrote (to cover some corner cases in the ISACOV) failed, in this case spike reach  his maximum steps, so after a quick debug, I found that spike raise an exception of loadaccess_fault on an address (example : 0xffff_fff0), in the opposite side the CVA6 does raise any exception and threat the address in his valid range.

image

So to reproduce the problem you only need to execute the command line :  
python3 cva6.py --testlist=../tests/testlist_isacov.yaml --test load_reg_hazard --iss_yaml cva6.yaml --target cv32a6_embedded --iss=vcs-uvm,spike (run it in cva6/verif/sim) of course using the new spike (cva6 HASH 5e68fc0a3fde3333d645d0b43fbce58bebabe6cf)

zchamski:
Enable silent fetch/load of zeroes when accessing unmapped memory locations in Spike:
- Bump CVV to use Spike supporting unmapped accesses when core-specific parameter `allow_unmapped_mem_access` is set to `True`.
- Add `allow_unmapped_mem_access: True` field to core-specific section of CV32A60X and CV32A65X Spike Yaml files.

### DIFF
--- a/config/gen_from_riscv_config/cv32a60x/spike/spike.yaml
+++ b/config/gen_from_riscv_config/cv32a60x/spike/spike.yaml
@@ -58,3 +58,4 @@ spike_param_tree:
       mcycleh_implemented: false
       mhpmevent31_implemented: false
       cvxif_x_num_rs: 2
+      allow_unmapped_mem_access: true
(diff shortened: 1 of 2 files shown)


AyoubJalali:
LGTM, thanks for the fix

PR #2885, opened 2025-03-31
Increase code coverage priv_lvl in csr_regfile

Gchauvon:
Increase code coverage priv_lvl in csr_regfile

### DIFF
--- a/core/csr_regfile.sv
+++ b/core/csr_regfile.sv
@@ -2571,9 +2571,7 @@ module csr_regfile
                               ? 1'b1
                               : 1'b0;
   end else if (CVA6Cfg.RVS) begin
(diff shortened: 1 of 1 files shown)


core/csr_regfile.sv
```diff
@@ -2570,10 +2570,11 @@ module csr_regfile
                                priv_lvl_o != riscv::PRIV_LVL_M && v_q)
                               ? 1'b1
                               : 1'b0;
+  end else if (CVA6Cfg.RVS) begin
```
JeanRochCoulon, line 2573: The modification creates a bug in the case RVU=1 and RVS=0.
```suggestion
  end else if (CVA6Cfg.RVU) begin
```

PR #2888, opened 2025-03-31
add macros for assertions else clauses

### LINKED ISSUE #1624: Immediate assertions in the CVA6 RTL

### Bug Description

This issue is motivated by #1619.

A quick grep of the CVA6 RTL reveals approximately 50 `assert` statements of the form `assert (expression);`. These are called "_immediate assertions_" in the SystemVerilog LRM.  Immediate assertions without an `else` clause are dangerous because they will only emit a simulator-specific message to stdout, without interacting with the simulator or the testbench.  That is, once the assertion fires and the message is written to stdout, there is no further record of the assertion, so it is not possible for the testbench to know that the assertion happened, and bugs can be missed.

For that reason there should be an _else_ clause for all assertions:
```
assert (expression)
  else `uvm_error("<Identifier>", "Useful (?) error message")
```
In non-UVM environments. the `uvm_error()` could be replaced with either $error() or $fatal().  In these cases, the script that launches the simulation must post-process the logfile to check for these messages.  This is not considered ideal since it is easy for the error message in the logfile to be missed.

To resolve this Issue I propose the following:
- Create a "_cvv_error()_" macro that will emit either a `uvm_error()` for UVM simulations or `$fatal()` for non-UVM sims.
- Update all immediate assertions to add an `else` clause as follows:
```
assert(expression)
  else `cvv_error("<Identifier>", "Useful (?) error message")
```

valentinThomazic:
Fix https://github.com/openhwgroup/cva6/issues/1624
Add macros for else clauses in immediate assertions throwing either SystemVerilog warnings/errors/fatals or UVM ones when applicable.
__FILE__ and __LINE__ are not displayed in the error message because Spyglass was throwing "Undefined Macros" errors. 
VCS does print them anyway.

### DIFF
--- a/core/csr_regfile.sv
+++ b/core/csr_regfile.sv
@@ -12,6 +12,7 @@
 // Date: 05.05.2017
 // Description: CSR Register File as specified by RISC-V
 
+`include "utils_macros.svh"
 
 module csr_regfile
   import ariane_pkg::*;
@@ -2747,7 +2748,7 @@ module csr_regfile
   // check that eret and ex are never valid together
   assert property (@(posedge clk_i) disable iff (!rst_ni !== '0) !(eret_o && ex_i.valid))
   else begin
-    $error("eret and exception should never be valid at the same time");
+    `ASSERT_ERROR("eret and exception should never be valid at the same time");
     $stop();
   end
   //pragma translate_on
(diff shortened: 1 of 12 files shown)


ASintzoff (approved):
LGTM

PR #2896, opened 2025-04-02
Verif : Don't bind obi_amo_if if we don't support atomic ext

AyoubJalali:
This MR improve functional coverage by condition binding amo assertions in tb


ASintzoff (approved):
LGTM

PR #2902, opened 2025-04-07
Update RS3 RType and CUS_CADD instructions to respect to CVXIF instruction. Fix forwardings issues.

zchamski:
Update the specification of CV-X-IF verification instructions to simplify decoding and forwarding.
This fixes forwarding issues caused by implicit x10 destination that would break decoder+ forwarding regularity by requiring additional lookups. 

In CUS_ADD_RS3_RTYPE rd is now same as rs3. To stay uniform with tools and RISCV UNPRIV, rd == rs1 in CUS_CADD.


ASintzoff (approved):
LGTM

This PR fixes cvxif-regression jobs.

PR #2915, opened 2025-04-11
Clean up irrelevant FIXME/TODO

yanicasa:
cvxif : Exec stage is non speculative
decode.sv TODO is code cleaning


ASintzoff (approved):
LGTM

PR #2925, opened 2025-04-14
Code coverage : exclude second instantiation

AyoubJalali:
this exclude second instantiation for fifo instr & instr scan


ASintzoff (approved):
LGTM

PR #2936, opened 2025-04-16
Indicate dirty in mstatus.FS correctly

### LINKED ISSUE #2934: [BUG] mstatus.FS does not indicate dirty correctly

### Bug Description

I have found two issues with the implementation of `mstatus.fs` (specifically, the *dirty* flag):
- as of commit https://github.com/openhwgroup/cva6/commit/3ce44b1b4e3344b86053b6a34eb6954fc24efb8f, `dirty_fp_state_o` in `commit_state.sv` is not computed correctly any more.
  - Instructions such as `fld` do not set `dirty_fp_state_o`, as `commit_instr_i[i].fu != ACCEL`.
  -  I believe this part of the condition is missing parentheses: `commit_instr_i[i].fu == ACCEL && commit_instr_i[i].vfp`
- the dirty flag in `mstatus.fs` is currently asserted after *all* instructions that involve the FPU. 
  - According to the ISA manual, it should only assert in case the architecturally visible state (such as floating point registers) was modified.
  - For example, `mstatus.fs` currently indicates dirty for `fsd`, which is read-only with respect to the floating point registers.
  - Instead, `dirty_fp_state_o` in `commit_state.sv` should assert only for operations that write floating-point registers.
  
I will send a PR with proposed fixes for both issues shortly.

WorldofJARcraft:
This commit makes two changes to the commit stage:
- It fixes incorrect parenthesis, causing dirty_fp_state_o not to be set on floating point instructions such as fld.
- It adds a condition for dirty_fp_state_o that only asserts the flag for floating point instructions that change any FPU registers.

This PR fixes #2934

### DIFF
--- a/core/include/ariane_pkg.sv
+++ b/core/include/ariane_pkg.sv
@@ -608,13 +608,13 @@ package ariane_pkg;
   endfunction
 
   function automatic logic fd_changes_rd_state(input fu_op op);
-    unique case(op) inside
-      [FSD:FSB],  // stores
-      FCVT_F2I,   // conversion to int
-      FMV_F2X,    // move as-is to int
-      FCLASS:     // classification (writes output to integer register)
-        return 1'b0; // floating-point registers are only read
-      default: return 1'b1; // other ops - floating-point registers are written as well
+    unique case (op) inside
+      [FSD : FSB],  // stores
+      FCVT_F2I,  // conversion to int
+      FMV_F2X,  // move as-is to int
+      FCLASS:  // classification (writes output to integer register)
+      return 1'b0;  // floating-point registers are only read
+      default: return 1'b1;  // other ops - floating-point registers are written as well
     endcase
   endfunction
 
--- a/core/include/ariane_pkg.sv
+++ b/core/include/ariane_pkg.sv
@@ -607,6 +607,17 @@ package ariane_pkg;
     endcase
   endfunction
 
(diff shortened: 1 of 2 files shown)


core/include/ariane_pkg.sv
```diff
@@ -574,6 +574,17 @@ package ariane_pkg;
     endcase
   endfunction
 
+  function automatic logic fd_changes_rd_state(input fu_op op);
+    unique case (op) inside
+      [FSD : FSB],  // stores
```
Gchauvon, line 579: Same behaviour but a bit clearer as there is no need to check enum type definition to see if FSW and FSH are not missing.
```suggestion
      FSD, FSW, FSH, FSB,  // stores
```

Gchauvon (approved):
Thank you for this nice contribution !

WorldofJARcraft:
Rebased to master.

JeanRochCoulon:
Thanks for the contribution.
Can you rebase ?  I have no right on Cispa Github domain. This would more convenient to submit your PR from personal repository. In that way, I could rebase by myself. Otherwise I have to ask to do so.

WorldofJARcraft:
@JeanRochCoulon I have rebased the branch.
I will leave the PR as-is for now and submit from a personal fork in the future.

JeanRochCoulon:
Can you rebase (again) ?

PR #2938, opened 2025-04-16
Fix URLs to point to CV32A60X-specific files on RTDs.

MikeOpenHWGroup:
This fixes the paths for the CV32A60X-specific documentation (from the `cv32a60x` branch).  Whenever the `cv32a60x` branch is updated, the documentation will be regenerated by RTD.


JeanRochCoulon:
Nice, it works !

PR #2939, opened 2025-04-16
[DO NOT MERGE] docs: Review of the current cv32a60x documentation

cairo-caplan:
DO NOT MERGE!

Hi @ASintzoff and @MikeOpenHWGroup ,

I analysed the docs under https://cva6-cv32a60x.readthedocs.io/en/latest/07_cv32a60x/, the parts that are specific to the CV32A60X, I would like to point out some questions and suggestions:

* Line 1345 of [machine.adoc](https://github.com/openhwgroup/cva6/compare/cv32a60x...cairo-caplan:cva6:cv32a60x?expand=1#diff-696b2e33a1189641c377671cf26cf7eebce4587a2ba0aa33dd7dfa0b40fc006e) , on section "3.1.6.7. Extension Context Status in `mstatus` Register", the paragraph "[CV32A60X] When an extension’s status is set to Off, any instruction that attempts to read or write the corresponding state will cause an illegal-instruction exception.". It starts with the "[CV32A60X]" ([{ohg-config}]), but I don't see this paragraph specific to the [CV32A60X].
* Line 1649 of [machine.adoc](https://github.com/openhwgroup/cva6/compare/cv32a60x...cairo-caplan:cva6:cv32a60x?expand=1#diff-696b2e33a1189641c377671cf26cf7eebce4587a2ba0aa33dd7dfa0b40fc006e): Mode => MODE, as this is what is used on other parts of the text and on Figure 9.
* Many times an instruction is listed with a \`\` syntax, others they are formatted as plain text. I am setting them as plain text (without \`\`). Feel free to completely ignore this though.
* In section 3.1.10. Hardware Performance Monitor, on the part "[CV32A60X] The hardware performance monitor includes 29 additional 64-bit event counters". I see the standard amount of Performance counters is 6, as in `/core/csr_regfile.sv` line 589: `localparam int unsigned MHPMCounterNum = 6;`. Should it be mentioned here that the amount of performance counters is configurable, with a maximum amount of 29?
* On section "3.1.19. Machine Security Configuration (mseccfg) Register", it is written that "As XLEN=32, mseccfgh is a 32-bit read/write register that aliases bits 63:32 of mseccfg". At the same time, the next paragraph affirms that "[CV32A60X] As Zkr, Smepmp, and Smmpm extensions are not implemented, mseccfg and mseccfgh do not exist.". Only the last one should exist in my opinion.

### DIFF
--- a/docs/riscv-isa/src/colophon.adoc
+++ b/docs/riscv-isa/src/colophon.adoc
@@ -276,15 +276,15 @@ and this and future versions of this document will be released under the
 same license.
 * Rearranged chapters to put all extensions first in canonical order.
 * Improvements to the description and commentary.
-* Modified implicit hinting suggestion on `JALR` to support more efficient
-macro-op fusion of `LUI/JALR` and `AUIPC/JALR` pairs.
+* Modified implicit hinting suggestion on JALR to support more efficient
+macro-op fusion of LUI/JALR and AUIPC/JALR pairs.
 * Clarification of constraints on load-reserved/store-conditional
 sequences.
 * A new table of control and status register (CSR) mappings.
 * Clarified purpose and behavior of high-order bits of `fcsr`.
-* Corrected the description of the `FNMADD`._fmt_ and `FNMSUB`._fmt_
+* Corrected the description of the FNMADD._fmt_ and FNMSUB._fmt_
 instructions, which had suggested the incorrect sign of a zero result.
-* Instructions `FMV.S.X` and `FMV.X.S` were renamed to `FMV.W.X` and `FMV.X.W`
+* Instructions FMV.S.X and FMV.X.S were renamed to `FMV.W.X` and `FMV.X.W`
 respectively to be more consistent with their semantics, which did not
 change. The old names will continue to be supported in the tools.
 * Specified behavior of narrower (latexmath:[$<$]FLEN) floating-point
@@ -317,20 +317,20 @@ avoid moving the _rd_ specifier in very long instruction formats.
 the counter registers are introduced, as opposed to only being
 introduced later in the floating-point section (and the companion
 privileged architecture manual).
-* The SCALL and SBREAK instructions have been renamed to `ECALL` and
-`EBREAK`, respectively. Their encoding and functionality are unchanged.
+* The SCALL and SBREAK instructions have been renamed to ECALL and
+EBREAK, respectively. Their encoding and functionality are unchanged.
 * Clarification of floating-point NaN handling, and a new canonical NaN
 value.
 * Clarification of values returned by floating-point to integer
 conversions that overflow.
-* Clarification of `LR/SC` allowed successes and required failures,
+* Clarification of LR/SC allowed successes and required failures,
 including use of compressed instructions in the sequence.
 * A new `RV32E` base ISA proposal for reduced integer register counts,
 supports `MAC` extensions.
 * A revised calling convention.
 * Relaxed stack alignment for soft-float calling convention, and
 description of the RV32E calling convention.
-* A revised proposal for the `C` compressed extension, version 1.9 .
+* A revised proposal for the `C` compressed extension, version 1.9.
 
 [.big]*_Preface to Version 2.0_*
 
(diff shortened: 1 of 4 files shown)


ASintzoff:
Thanks for the review.

A comment to explain how the RISC-V ISA manual tailored for CV32A60X is generated.
The source files from [riscv-isa-manual](https://github.com/riscv/riscv-isa-manual) to be tailored are copied and modified as little as possible. For instance, when I discovered errors I created pull requests on that repository. Therefore, for instance, we do not change how backticks are used or change word capitalization.

3.1.6.7: you are right that the paragraph is not specific to CV32A60X but IMO, as the documentation is only for this configuration, it is not a real issue.

3.1.10: to be compliant with RISC-V ISA, the `mphcounter[i] (3<=i<=31)` must exist. In CV32A60X, there is no usable performance counter so they are always equal to zero. This is not configurable.  

3.1.19: good catch, only the last paragraph should be kept to avoid any confusion.

docs/riscv-isa/src/colophon.adoc
```diff
@@ -317,20 +317,20 @@ avoid moving the _rd_ specifier in very long instruction formats.
...
-* Clarification of `LR/SC` allowed successes and required failures,
+* Clarification of LR/SC allowed successes and required failures,
 including use of compressed instructions in the sequence.
 * A new `RV32E` base ISA proposal for reduced integer register counts,
 supports `MAC` extensions.
 * A revised calling convention.
 * Relaxed stack alignment for soft-float calling convention, and
 description of the RV32E calling convention.
-* A revised proposal for the `C` compressed extension, version 1.9 .
+* A revised proposal for the `C` compressed extension, version 1.9.
```
cairo-caplan, line 333: Extra empty space removed

docs/riscv-isa/src/machine.adoc
```diff
@@ -94,10 +94,10 @@ endif::[]
 ifdef::archi-CVA6[]
 [CVA6] The Extensions field encodes the presence of the standard extensions,
 with a single bit per letter of the alphabet (bit 0 encodes presence of
-extension "A" , bit 1 encodes presence of extension "B", through to
+extension "A", bit 1 encodes presence of extension "B", through to
 bit 25 which encodes "Z"). The "I" bit will be set for RV32I, RV64I,
 and RV128I base ISAs, and the "E" bit will be set for RV32E and RV64E.
-In CVA6, the Extensions field is not writeable, the presence of standard
+In CVA6, the Extensions field is not writable, the presence of standard
```
cairo-caplan, line 100: Fix writable => writeable

docs/riscv-isa/src/machine.adoc
```diff
@@ -400,7 +407,8 @@ endif::[]
 
 ifdef::archi-CV32A60X,archi-CV32A65X[]
 The `mimpid` CSR provides a unique encoding of the version of the
-processor implementation.
+processor implementation. The Implementation value should reflect the
```
cairo-caplan, line 410: The part:
"The Implementation value should reflect the design of the RISC-V processor itself and not any surrounding system." here and on line 420, from the default documentation, still has some information value for `mimpid`, but I also think it is all right to be kept as it is.

docs/riscv-isa/src/machine.adoc
```diff
@@ -515,7 +523,7 @@ endif::[]
 
 ifdef::archi-default,XLEN-32[]
 [[mstatushreg]]
-.Additional machine-mode status (`mstatush`) register  for RV32.
+.Additional machine-mode status (`mstatush`) register for RV32.
```
cairo-caplan, line 526: Empty space

docs/riscv-isa/src/machine.adoc
```diff
@@ -1297,7 +1305,7 @@ read-only zero, then SD is also always zero.
 endif::[]
 
 ifdef::archi-CVA6[]
-[{ohg-config}] The SD bit is a read-only bit that summarizes whether either the FS, VS,
+The SD bit is a read-only bit that summarizes whether either the FS, VS,
```
cairo-caplan, line 1308: This part is just a description on the SD bit, which is not specific only to the particular CVA6 configuration, so I think it is worth to just remove the `[{ohg-config}].`

docs/riscv-isa/src/machine.adoc
```diff
@@ -2221,7 +2229,7 @@ do not exist.
 endif::[]
 
 ifdef::archi-CV32A60X,archi-CV32A65X[]
-As the Sscofpmf extension is not implemented, the `mhpmevent__n__h` CSRs
+[{ohg-config}] As the Sscofpmf extension is not implemented, the `mhpmevent__n__h` CSRs
```
cairo-caplan, line 2232: The missing Sscofpmf extension is particular to the CV32A60X configuration, as shown in line 2124. So I think it should also add `[{ohg-config}]` on the beginning of the paragraph

cairo-caplan (commented):
Thank you, @ASintzoff, for answering and reviewing my review. 

I commented some lines I changed on this PR that could still be considered after all.

cairo-caplan:
Hi @jquevremont, no worries.

Yes, I just checked and can confirm that [my last comments](https://github.com/openhwgroup/cva6/pull/2939#pullrequestreview-2776111229) would still be relevant to be checked by @ASintzoff.

Thank you, @ASintzoff and @jquevremont.

PR #2951, opened 2025-05-03
Add ALU to ALU bypass in superscalar mode

ricted98:


ci/install-prereq.sh
cathales, line 1: Thanks for this fix 🙏
It is not related to this PR, so I think it should be split to another PR.

It would allow maintainers to merge it earlier, before the review of ALU-ALU bypass.

ricted98: Sure, take a look at: https://github.com/openhwgroup/cva6/pull/2953!

ricted98:
- [ ]  Make the ALU-ALU forwarding parametric
- [ ] Rebase on top of latest master branch

core/include/ariane_pkg.sv
```diff
@@ -532,6 +532,11 @@ package ariane_pkg;
     SHA512SUM1
   } fu_op;
 
+  typedef struct packed {
+    logic rs1_into_rd;
+    logic rs2_into_rd;
```
cathales, line 537: Isn’t it `rsX_from_rd` instead of `into` ?

core/issue_read_operands.sv
```diff
@@ -359,11 +365,24 @@ module issue_read_operands
         end
         ALU: begin
           if (CVA6Cfg.SuperscalarEn && !fus_busy[0].alu2) begin
-            fus_busy[1].alu2 = 1'b1;
```
cathales, line 362: I do not see why these changes should affect structural hazards.
I would expect them to simply bypassing RAW stalls is some cases.

core/issue_read_operands.sv
```diff
@@ -591,15 +612,32 @@ module issue_read_operands
...
-        stall_raw[1] = 1'b1;
+        // Check if the operations are ALU -> ALU, if not, stall.
+        if (!(issue_instr_i[0].fu == ALU && issue_instr_i[1].fu == ALU)) begin
+          stall_raw[1] = 1'b1;  // RS1[1] NEEDS RD[0]
+          // If it is a ALU -> ALU, we can fuse all operation beside CPOP (maybe can be optimized OP -> CPOP, to explore).
+        end else if ( ((issue_instr_i[0].op inside {CPOP, CPOPW}) || (issue_instr_i[1].op inside {CPOP, CPOPW}) )) begin
+          stall_raw[1] = 1'b1;  // RS1[1] NEEDS RD[0]
+        end else begin
+          // We cannot fuse if we have the same destination register, it will create a forwarding inconsistency.
+          alu_bypass.rs1_into_rd = issue_instr_i[0].rd != issue_instr_i[1].rd;
```
cathales, line 623: Is it still needed since a renaming-like feature was added?

core/issue_read_operands.sv
```diff
@@ -591,15 +612,32 @@ module issue_read_operands
...
       if ((!CVA6Cfg.FpPresent || (is_rs2_fpr(
               issue_instr_i[1].op
           ) == is_rd_fpr(
               issue_instr_i[0].op
           ))) && issue_instr_i[1].rs2 == issue_instr_i[0].rd && issue_instr_i[1].rs2 != '0) begin
-        stall_raw[1] = 1'b1;
+        if (!(issue_instr_i[0].fu == ALU && issue_instr_i[1].fu == ALU)) begin
+          stall_raw[1] = 1'b1;  // RS2[1] NEEDS RD[0]
+          // If it is a ALU -> ALU, we can fuse all operation beside CPOP (maybe can be optimized OP -> CPOP, to explore).
+        end else if ( ((issue_instr_i[0].op inside {CPOP, CPOPW}) || (issue_instr_i[1].op inside {CPOP, CPOPW}) )) begin
```
cathales, line 635: These conditions seem repeated from above, could we use an intermediate signal as a single condition instead?

We have the two cases of "not ALU->ALU" which imply the standard `stall_raw()`, then the case "ALU->ALU" which implies the specific case added in this PR.

core/issue_read_operands.sv
```diff
@@ -813,6 +851,14 @@ module issue_read_operands
       if (issue_instr_valid_i[i] && !fu_busy[i]) begin
         if (!stall_raw[i]) begin
           issue_ack[i] = 1'b1;
+          // For now keep this as is, we cannot do inter dual-issue forwarding
+          if (i > 0) begin
+            if ((issue_instr_i[i].rd[4:0] == issue_instr_i[i-1].rd[4:0]) && (issue_instr_i[i].rd[4:0] != '0)) begin
```
cathales, line 856: Are these `[4:0]` slicing needed?

Shouldn’t this block of conditions be already managed inside `stall_raw[1]` instead of here?

core/issue_read_operands.sv
```diff
@@ -813,6 +851,14 @@ module issue_read_operands
       if (issue_instr_valid_i[i] && !fu_busy[i]) begin
         if (!stall_raw[i]) begin
           issue_ack[i] = 1'b1;
+          // For now keep this as is, we cannot do inter dual-issue forwarding
```
cathales, line 854: Isn’t it the feature added by this PR?

core/alu_wrapper.sv
```diff
@@ -0,0 +1,115 @@
...
+  import ariane_pkg::*;
+#(
+    parameter config_pkg::cva6_cfg_t CVA6Cfg = config_pkg::cva6_cfg_empty,
+    parameter type fu_data_t = logic
+) (
+    input  logic                                clk_i,
+    input  logic                                rst_ni,
+    input  alu_bypass_t                         alu_bypass_i,
+    input  fu_data_t    [1:0]                   fu_data_i,
+    output logic        [1:0][CVA6Cfg.XLEN-1:0] result_o,
```
cathales, line 25: This should probably be configurable instead of assigning a lot of wires to 0

core/alu.sv
```diff
@@ -204,13 +204,19 @@ module alu
 
   if (CVA6Cfg.RVB) begin : gen_bitmanip
     // Count Population + Count population Word
+    if (CVA6Cfg.SuperscalarEn) begin
```
cathales, line 207: So we drop support for pop count as soon as the processor is superscalar?

cathales (changes requested):

core/alu.sv
ricted98: The CPOP operation is moved outside of the ALUs to move it outside of the combinational path with the two ALUs chained.

core/issue_read_operands.sv
ricted98: Don't we need to make sure the ALU-ALU instructions with RAW hazards use the correct ALUs? i.e. the older instruction uses `alu`, the younger instruction uses `alu2`

core/include/ariane_pkg.sv
ricted98: Will do!

core/alu_wrapper.sv
ricted98: Will do!

core/issue_read_operands.sv
ricted98: Unclear comment that I will remove. There is later a check for the specific case of ALU-ALU bypass to allow it.

https://github.com/openhwgroup/cva6/blob/bee8bbcb75b4a0b80fa16267c50dfa9cae45c600/core/issue_read_operands.sv#L857

core/issue_read_operands.sv
ricted98: Nice catch, I would say not.

core/issue_read_operands.sv
cathales: The initial version of superscalar does the opposite: prefer using `alu2` when it is possible, use `alu` for instructions where `alu2` cannot be used (eg. branch) or when `alu2` is already busy.

So your ALU-ALU forwarding could be only alu2->alu, and the issue stage should always issue in this order.

core/alu.sv
cathales: Ah I didn’t see it, sorry. Thanks for the pointer!

I was also surprised that the condition was only `SuperscalarEn`, IIUC you will add another parameter to conditionally enable ALU-ALU forwarding, right?

ricted98: Exactly. I was unsure if we wanted it "baked-in" with the superscalar or not (e.g. how renaming was upstreamed). Also @JeanRochCoulon asked me to parametrize this feature, I am doing it.

core/issue_read_operands.sv
```diff
@@ -586,20 +607,34 @@ module issue_read_operands
     end
 
     if (CVA6Cfg.SuperscalarEn) begin
+
+      automatic logic is_alu_bypass;
```
cathales, line 611: Could we use something more standard here?
I don’t know how the tools will handle this case.

For instance:

```systemverilog
logic is_alu_bypass;

if (...) begin
  assign is_alu_bypass = ...;
end else begin
  assign is_alu_bypass = '0;
end
```

ricted98: Will do!

ricted98:
@JeanRochCoulon would it be possible to understand what is failing in the FPGA flow in the CI? I tried to run an FPGA synth locally and it arrives at the end, but I cannot reproduce the CI setup one-to-one.

JeanRochCoulon:
A file is not available, this causes the FPGA build fail... I am looking at it.

JeanRochCoulon:
Before investing too much engineering into this PR, I would like to double check the performance gain. On Dhrystone, the performance gain is 1%, quite low. And what about the Coremark?

ricted98:
I tried to run CoreMark with the same setup as the CI with and without ALU bypass.
Without ALU bypass:
```
CoreMark/MHz 1.0 : 4.07 / GCC13.2.0 -O3 -g -fno-tree-loop-distribute-patterns -static -mcmodel=medany -fvisibility=hidden -nostdlib -nostartfiles -lgcc -funroll-all-loops -ffunction-sections -fdata-sections -Wl,-gc-sections -falign-jumps=4 -falign-functions=16 / UNSPECIFIED() RATIOS:1
```
With ALU bypass:
```
CoreMark/MHz 1.0 : 4.24 / GCC13.2.0 -O3 -g -fno-tree-loop-distribute-patterns -static -mcmodel=medany -fvisibility=hidden -nostdlib -nostartfiles -lgcc -funroll-all-loops -ffunction-sections -fdata-sections -Wl,-gc-sections -falign-jumps=4 -falign-functions=16 / UNSPECIFIED() RATIOS:1
```

Thus, there is a CoreMark improvement of ~4.17%.

JeanRochCoulon:
Thank you, we see Coremark results are highly better than Dhrystone.
But I again fear that the RTL modification is big in relation to the performance increase. With @cathales we tried to find solution to increase again the performance based on your PR. Do you thing the coverage of the fusion of ALU+BRANCH would be possible, this PR should not be far from supporting this.

ricted98:
Thank you for the feedback. For the ALU+BRANCH the fusion is possible, but it would likely lead to a long combinational path ALU2 (not branch capable) ->ALU (branch capable) -> BRANCH UNIT. I will need to run some synthesis if we want to go down this path. At the moment the only legal chaining is ALU->ALU2, so this path would not pop up.

JeanRochCoulon:
If ALU2 + ALU + BRANCH does not impact critical path, this PR would be a great improvement for cva6 (to be confirmed on Coremark). If not, let's discuss.

Today code is fully verified in cv32a60x configuration (100% code and functional coverages). Modifying lines impact maturity status, and constraint verification team to continue their work. But if performance is at top, everything is possible !

ricted98:
Thanks for the explanation. I will look into this change and let you know how the PPA is affected.

ricted98:
Hi @JeanRochCoulon, @cathales. Sorry for the delay but I finally had some time to look into the suggested change to allow ALU->BRANCH UNIT.

The PR as of now only allows ALU->ALU forwarding, with a somewhat more complex issue logic to reorder instructions on the two alus to allow forwarding. The first ALU is the branch capable one, while the second one is the non-branch capable one.

The changes to allow ALU->BRANCH UNIT forwarding can be found here: https://github.com/ricted98/cva6/commit/c5d9dd03fcd25ebe1bf02f415ee807b107b216f2. I am excluding JALR for the moment, as it would require additional muxing.

On CoreMark, I have observed a 1% improvement over the plain ALU-ALU bypass. However, the critical path is now on the ALU->ALU->BRANCH UNIT ->FLUSH path. I can give more detailed synthesis results in the upcoming hours.

Should I add the ALU->BRANCH UNIT commit to this PR as well nevertheless?

ricted98:
To complement what was previously said, if we keep the ```LoadPipeReg``` enabled the maximum frequency drops by ~15%.

core/issue_read_operands.sv
```diff
@@ -289,6 +297,27 @@ module issue_read_operands
...
+    assign is_alu_bypass =
+      (issue_instr_i[0].fu == ALU && issue_instr_i[1].fu == ALU) &&
+      !((issue_instr_i[0].op inside {CPOP, CPOPW}) || (issue_instr_i[1].op inside {CPOP, CPOPW}));
+  end else begin
+    assign is_alu_bypass = 1'b0;
+  end
+
+  if (CVA6Cfg.SuperscalarEn) begin
+    // When a bypass is possible, an instruction uses `alu2` only when `alu` is already busy,
+    // in all other scenarios `alu2` is preferred over `alu, unless it is busy
```
cathales: `` `alu `` missing backtick here

JeanRochCoulon:
Great job Riccardo, the final code is very nice !

PR #2958, opened 2025-05-06
Bugfix/conflicting declaration read elf

OlivierBetschi:
Use define to constraint function declaration from DPI-C. This was already done in ariane_tb.sv and rvfi_tracer.sv
This prevent an error with xcelium reporting that functions such as read_elf have multiple definitions


JeanRochCoulon:
Can you rebase...

PR #2959, opened 2025-05-06
Add missing value to return

OlivierBetschi:
return requires a value to be returned (detected as an xcelium compilation error). Set a default value of 0.


JeanRochCoulon:
Hello @OlivierBetschi 
The PR need to be rebased before merging. I cannot do it because I have no right on boschglobal GitHub domain, could you rebase ?
For the coming PRs, would it be possible to propose a branch with write permissions (personal branch for instance) ?

PR #2968, opened 2025-05-17
Hybrid cache clean

ibooga:


JeanRochCoulon:
Hi @ibooga 
Thank you for the interest in CVA6.
For information, the cva6 community plans to remove WT and WB caches in short term and promotes the HPDCache. Your PR do not take this direction. If you would like to present your PR, I propose you to attend and present during one cva6 weekly (on Tuesday). @jquevremont moderates the weekly. @cfuguet designed the HPDCache.
Regards

cfuguet (commented):
This looks like an interesting work.

However, as @JeanRochCoulon was saying, before the CVA6 team can accept such a "big" contribution (many modifications), it would be better to have a presentation in one of the CVA6 weekly meetings.

In addition to that, we decided a long time ago to keep the HPDcache and the CVFPU as submodules. This is because those are independent and active projects that are used beyond the CVA6 only. The submodule approach allows to more easily track changes. Importing them as subtrees requires a huge effort to both pull and push contributions in the original repositories.

PR #2984, opened 2025-05-29
Fix makefile (bjdooks/fix makefile cc)

bjdooks-ct:
It looks like commit 92b2d0b73b3ab825cbffcd4cc1430949b7d3f65d didnt quite work as intended, so fix the setting of CC


JeanRochCoulon:
@math-gout Can you approve this PR ?

PR #2992, opened 2025-06-03
Prevent verif/regress scripts from failing silently

LawrenceHunter:
Simple addition to prevent errors in the smoke test scripts not being reported.

### DIFF
--- a/verif/regress/cv32a6_tests.sh
+++ b/verif/regress/cv32a6_tests.sh
@@ -8,6 +8,8 @@
 #
 # Original Author: Jean-Roch COULON - Thales
 
+set -exo pipefail
+
 # where are the tools
 if ! [ -n "$RISCV" ]; then
   echo "Error: RISCV variable undefined"
@@ -54,6 +56,9 @@ riscv_tests_list=(
   rv32ui-p-beq
   rv32ui-p-jal
 )
+
+set +exo pipefail
+
 for t in ${riscv_tests_list[@]} ; do
   python3 cva6.py --testlist=../tests/testlist_riscv-tests-${DV_TARGET}-p.yaml --test $t --iss_yaml cva6.yaml --target ${DV_TARGET} --iss=$DV_SIMULATORS $DV_OPTS
   [[ $? > 0 ]] && ((errors++))
@@ -62,6 +67,8 @@ done
 python3 cva6.py --target ${DV_TARGET} --iss=$DV_SIMULATORS --iss_yaml=cva6.yaml --c_tests ../tests/custom/hello_world/hello_world.c --linker=../../config/gen_from_riscv_config/linker/link.ld\
   --gcc_opts="-static -mcmodel=medany -fvisibility=hidden -nostdlib -nostartfiles -g ../tests/custom/common/syscalls.c ../tests/custom/common/crt.S -lgcc -I../tests/custom/env -I../tests/custom/common" $DV_OPTS
 [[ $? > 0 ]] && ((errors++))
+
+set -exo pipefail
 
 make -C ../.. clean
 make clean_all
--- a/verif/regress/cv64a6_imafdc_tests.sh
+++ b/verif/regress/cv64a6_imafdc_tests.sh
@@ -11,6 +11,8 @@
 # Contributor: Cesar Fuguet, CEA List
 # RISC-V tests for 64-bit configurations implementing MMU
 
+set -exo pipefail
+
 # where are the tools
 if ! [ -n "$RISCV" ]; then
   echo "Error: RISCV variable undefined"
@@ -57,6 +59,9 @@ riscv_tests_list=(
   rv64ui-v-beq
   rv64ui-v-jal
 )
+
+set +exo pipefail
+
 for t in ${riscv_tests_list[@]} ; do
   python3 cva6.py --testlist=../tests/testlist_riscv-tests-cv64a6_imafdc_sv39-v.yaml --test $t --iss_yaml cva6.yaml --target ${DV_TARGET} --iss=$DV_SIMULATORS $DV_OPTS
   [[ $? > 0 ]] && ((errors++))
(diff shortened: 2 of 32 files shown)


JeanRochCoulon:
Thanks @LawrenceHunter for your suggestion. But I would be a little bit reluctant to merge. This is the reason: our CI executes jobs, we would like to give the opportunity to have some fails into the jobs: for instance, tests from the the whole arc-test suite can be failed. At the end we report the final result (for instance 3 fails over 123 tests).

LawrenceHunter:
I believe that is all done in the cva6.py file so would be unaffected by this change as that python file will handle its own errors; instead here this would prevent the running of the cva6.py file is the previous `make clean_all` command failed

JeanRochCoulon:
@cathales What do you think ?

cathales:
What would you think of this?

```bash
error=0
python3 cva6.py ... || error=$?
python3 cva6.py ... || error=$?
...
exit $error
```

LawrenceHunter:
I think that one is a little misleading as it will only return the true exit status of the final command that failed but does still solve my original concern, so happy if that approach is more acceptable

JeanRochCoulon:
Who will modify the scripts :-)

LawrenceHunter:
It should remain, the intention is that if for example 'make clean' errors the script won't continue in an invalid state, but instead the discussion above means that for the cva6.py testing steps all of them are executed, and the failure reported as a whole. I have actually missed setting the fail in one of the files and will fix that

cathales (approved):

LawrenceHunter:
Okay, thanks for the clarification I will instead update them all to get rid of this useless step

LawrenceHunter:
Hi, I've obviously messed up one of the scripts https://riscv-ci.pages.thales-invia.fr/dashboard/dashboard_cva6_2992.html but I can't really see what the bot does can you hint me to what script is failing?

verif/regress/cv32a6_tests.sh
```diff
@@ -54,6 +56,9 @@ riscv_tests_list=(
   rv32ui-p-beq
   rv32ui-p-jal
 )
+
+set +exo pipefail
```
cathales, line 60: Why is it set twice?

verif/regress/cv32a6_tests.sh
```diff
@@ -63,6 +68,8 @@ python3 cva6.py --target ${DV_TARGET} --iss=$DV_SIMULATORS --iss_yaml=cva6.yaml
   --gcc_opts="-static -mcmodel=medany -fvisibility=hidden -nostdlib -nostartfiles -g ../tests/custom/common/syscalls.c ../tests/custom/common/crt.S -lgcc -I../tests/custom/env -I../tests/custom/common" $DV_OPTS
 [[ $? > 0 ]] && ((errors++))
 
+set -exo pipefail
```
cathales, line 71: Three times?

verif/regress/cv64a6_imafdc_tests.sh
```diff
@@ -57,6 +59,9 @@ riscv_tests_list=(
   rv64ui-v-beq
   rv64ui-v-jal
 )
+
+set +exo pipefail
```
cathales, line 63: Same

verif/regress/cv64a6_imafdc_tests.sh
cathales, line 1: Smoke test questa-testharness cv64a6_imafdc_sv39 is this one

verif/regress/hwconfig_tests.sh
cathales, line 1: HW config vcs-uvm cv32a65x is this one

verif/regress/cv32a6_tests.sh
LawrenceHunter: The + means to disable the setting at the top of the file. This is done intentionally to allow the script's pre-existing error handling to take over without modification which is then re-enabled (the third time), the intention was minimal change if your opinion is that this is messy, which I agree, and would prefer direct modification of the scripts' error handling I can update for that, but my assumption was this approach would be more palatable

verif/regress/cv64a6_imafdc_tests.sh
LawrenceHunter: Are there any logs available for the bot job? I can't see why this test would fail

verif/regress/cv32a6_tests.sh
cathales: Ah sorry I misread (I was in a hurry for the week-end) 🤦

I vote for having the same error handling mechanism everywhere.

verif/regress/cv64a6_imafdc_tests.sh
cathales: I cannot send you the logs, but this one fails because of mismatches between the RTL and Spike.

verif/regress/hwconfig_tests.sh
cathales: For this one it succeeds, but the issue is that in the CI it is `source`’d so the `exit` statement prevents from running the "report success" step here.
(By default we first generate an error report to avoid jobs silently failing. This report should be erased with another report, but in this case it doesn’t.)

https://github.com/openhwgroup/cva6/blob/7882136f84926f56e39794872c0788acb690ec71/.gitlab-ci.yml#L208-L209

LawrenceHunter: I'm not sure, I follow here, if the exit step is causing the python command to not run that means it exited with non-zero and therefore the test failed?

verif/regress/cv64a6_imafdc_tests.sh
LawrenceHunter: I'm having problems running this one locally due to VCS issues '/bin/sh: 2: vlogan: not found' but I have no idea how my changes would have caused a regression in RTL and Spike

verif/regress/hwconfig_tests.sh
cathales: No: it exited with zero; yet it exited anyway and skipped the `python3` line.
The issue is that `source` is used instead of `bash`.
Without `source` it would have exited the `hwconfig_tests.sh` script only (expected), not the whole job (currently observed).

verif/regress/cv64a6_imafdc_tests.sh
cathales: Agreed. The issue might come from the version of `master` branch this PR is based on.

JeanRochCoulon:
Thanks @LawrenceHunter for your interest in cva6 and your PR submission. But as this modification would impact CI good execution, I am constrained to close the PR.

PR #3013, opened 2025-06-13
Bjdooks/update mimpid post fs fix2

bjdooks-ct:
Using the MMMPID to show version of the core code is useful


bjdooks-ct:
Is there anything I need to do here, don't  understand the failures in the pipeline ?

Gchauvon:
Failure of the pipeline are not related to this PR.

bjdooks-ct:
> I'm not sure how we should define `MIMPID` csr. Setting it to 1 after a fix of the FPU seems _rushed_ to me.

I was thinking this is a good idea to increment this each time any serious bug is sorted out. It gives you an idea of the version of the core was used, so this could also increment with new releases, etc.

JeanRochCoulon:
Hello @bjdooks-ct 
Thanks again for contributing.
Your PR raised a discussion topic in OpenHW Mattermost channel (are you in?).
We need definitely to improve the MIMPID strategy, but setting it at 1 is not the wanted strategy. I let you know.

bjdooks-ct:
Since the MMPID changes for bugfixes seems to be still debatable, should we just merged the configuration update to allow this to be changed from the config or drop the whole lot?

JeanRochCoulon:
A trade-off to prepare the future is to merge your PR, but setting MIMPID to 0.
For information, the MIMPID value will be discussed by end of August

jjardon:
@bjdooks-ct @nazar-codethink is there any further updates for this PR?

core/include/ariane_pkg.sv
```diff
@@ -54,6 +54,8 @@ package ariane_pkg;
 
   localparam logic [31:0] OPENHWGROUP_MVENDORID = 32'h0602;
   localparam logic [31:0] ARIANE_MARCHID = 32'd3;
+  // Use IMPID 1 to denote the fpu state issue has been fixed
+  localparam logic [31:0] ARIANE_MIMPID = 32'd1;
```
JeanRochCoulon: Not possible Today to change MIMPID from 0 to 1. 
Unfortunately I can only propose these two options:
- You modify with ARIANE_MIMPID = 0, or
- You close the PR
Thanks

bjdooks-ct: ok, pushed a new version without setting the value to 1

PR #3025, opened 2025-06-24
cva6_tlb: fix wrong indexing of `vaddr_vpn_match` when `RVH` is set

ricted98:
I tried using the CVA6 configuration `cv64a6_imafdch_sv39_wb` with QuestaSim. The following error is returned.

```
'vaddr_vpn_match[i][1][0]' is driven by more than one continuous assignment
```
where i ranges on all the values of the following loop.

https://github.com/openhwgroup/cva6/blob/301d11ceb88c1169f75e9dea415e4bff4eb29888/core/cva6_mmu/cva6_tlb.sv#L102-L112

My educated guess for the fix is the following:
https://github.com/ricted98/cva6/blob/ef6cc993c8b45b3402587c715ab65facc2a06f24/core/cva6_mmu/cva6_tlb.sv#L102-L112

Where the last index is not tied to `0` but rather it varies with `x`, however I would like someone more knowledgeable to review this change.


niwis:
LGTM. @JeanRochCoulon @ninolomata any update on this?

JeanRochCoulon:
Thanks @ricted98

PR #3030, opened 2025-06-27
instr_tracer: add fixes for superscalar configuration

ricted98:
When doing ```diff``` between a trace generated by the scalar and superscalar versions of the core, I noticed some leftover divergences.

- Wrong ```wdata``` due to missing index in padding
- The tracer has no visibility on dropped commits, so it wrongly uses ```sbe``` entries that should be discarded during the trace printing
- The source operands are printed from the shadow regfiles. To ensure they are always up to date, even when forwarding happened, the tracer should first commit on port 0, update the shadow regfiles and then process port 1.

I tested the changes on some embench-iot workloads and there is no divergence on the traces (once we omit the timing information, as scalar and superscalar will have different timing, and some branches being predicted differently due to different fetch window widths).
The command used is
```bash
diff <(cut -c24- trace1) <(cut -c24- trace2)
```


JeanRochCoulon:
Thanks @ricted98

PR #3032, opened 2025-06-29
Fix RAW hazard on rs3 for 3-op CV-X-IF usage

bugavix:
Adds signal in IRO to detect RAW data hazards when rs3 is used as the third operand in CV-X-IF.

### DIFF
--- a/core/issue_read_operands.sv
+++ b/core/issue_read_operands.sv
@@ -184,6 +184,7 @@ module issue_read_operands
   logic [ CVA6Cfg.NrIssuePorts-1:0]                                rs3_raw_check;
   logic [ CVA6Cfg.NrIssuePorts-1:0]                                rs3_has_raw;
   logic [ CVA6Cfg.NrIssuePorts-1:0]                                rs3_fpr;
+  logic [ CVA6Cfg.NrIssuePorts-1:0]                                rs3_gpr_cvxif;
 
 
   logic [CVA6Cfg.NR_SB_ENTRIES-1:0][ariane_pkg::REG_ADDR_SIZE-1:0] rd_list;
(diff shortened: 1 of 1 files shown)


core/issue_read_operands.sv
```diff
@@ -484,7 +485,12 @@ module issue_read_operands
         .idx_o(idx_hzd_rs3[i]),
         .valid_o(rs3_raw_check[i])
     );
-    assign rs3_has_raw[i] = rs3_raw_check[i] && rs3_fpr[i];
+    assign rs3_has_raw[i] = rs3_raw_check[i] && (rs3_fpr[i] || rs3_gpr_cvxif[i]);
+
+    // Check if the instruction is going to be offloaded to cvxif and
+    // the interface uses 3 operands.
+    assign rs3_gpr_cvxif[i] = (CVA6Cfg.CvxifEn && (OPERANDS_PER_INSTR == 3)
+        && issue_instr_i[i].op == OFFLOAD);
```
cathales: Some braces are useless here.

```suggestion
    assign rs3_gpr_cvxif[i] = CVA6Cfg.CvxifEn && (OPERANDS_PER_INSTR == 3)
        && issue_instr_i[i].op == OFFLOAD;
```

Also, have we considered moving this assignment into the `for` loop line 430, just after where `rs3_fpr` is assigned?

Gchauvon:
> Also, have we considered moving this assignment into the for loop line 430, just after where rs3_fpr is assigned?
Indeed it could be moved into this for loop

> Open question: should we add regression tests in the CI?
It would require to add NrRgprPorts as a CVA6 parameters and add a testbench with a copro supporting 3 operands.
The copro_example can be used to do that since it supports 3 operands. 
A directed test on a specific configuration could do the job.

cathales (approved):

JeanRochCoulon:
Great !   I merge

PR #3035, opened 2025-06-30
Correct logic for sign bit in load_unit.sv

JosephECSE1:
Without this change the logic is ambiguous.

### DIFF
--- a/core/load_unit.sv
+++ b/core/load_unit.sv
@@ -505,7 +505,7 @@ module load_unit
 
   // select correct sign bit in parallel to result shifter above
   // pull to 0 if unsigned
-  assign rdata_sign_bit = rdata_is_signed & rdata_sign_bits[rdata_offset] | (CVA6Cfg.FpPresent && rdata_is_fp_signed);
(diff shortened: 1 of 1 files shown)


JeanRochCoulon:
@cathales can you approve ?

cathales:
- [x] :+1: Switch from bitwise to to boolean operators makes the intent more explicit as type is actually `logic` (so this is equivalent to previous code)
- [x] Adding braces is actually unneeded but we can add them if you want.

Note: `&&` has precedence over `||` and `&` has precedence over `|`, this is standard logic, whatever the language with infix notations; just like `*` and `+` in mathematics notations.

cathales (approved):

JosephECSE1:
> Thanks @JosephECSE1 If you would like to present your cva6 project at OpenHW cva6 weekly, tell me.

Thankyou @JeanRochCoulon, yes that sounds like a fantastic opportunity. 

For some context, we've been working with RISC-V Big Endian for the last few months, here's is a link showing some previous work running Big Endian Linux on QEMU:
[](https://www.codethink.co.uk/articles/risc-v-big-endian-support-runtime-testing/)

JosephECSE1:
[https://www.codethink.co.uk/articles/risc-v-big-endian-support-runtime-testing/](url)

PR #3037, opened 2025-07-01
Updated cv64a60ax configuration

khandelwaltanuj:


khandelwaltanuj:
Hi @JeanRochCoulon 

I thought that utils.py and mmu.sv were already merged. 

Yeah, utils.py is related the issue regarding the reset values. Earlier it would just give some python errors, stops. Now it prints a message saying "reg_name: reset value is not specified" 

for mmu.sv yeah its related to misaligned load/store transaction. @fatimasaleem and @AngelaGonzalezMarino were ok that we fixe this issue during a discussion on this issue in github. 

Thanks and Regards
Tanuj

PR #3041, opened 2025-07-04
Fix typos

nazar-codethink:
Fixed some typos I noticed while reading the code

### DIFF
--- a/core/amo_buffer.sv
+++ b/core/amo_buffer.sv
@@ -11,7 +11,7 @@
 // Author: Florian Zaruba, ETH Zurich
 // Date: 20.09.2018
 // Description: Buffers AMO requests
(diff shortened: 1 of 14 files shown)


nazar-codethink:
> We advise you to submit your PR from a repository with access right. It allows me to rebase before merge.

I can't figure out how this should work, do you know how to do it?

nazar-codethink:
> The domain codethinklabs is protected, that is why I cannot rebase. One solution can be to open rigths on codethinklabs, but maybe it is not wanted by your company. Another solution is to use your personal Github branch, I think that by default rights are open.

It's apparently GitHub limitation https://github.com/orgs/community/discussions/5634
I'll keep that in mind for future

PR #3069, opened 2025-08-01
Minor coding fixes

oliverbm67:
Minor RTL rewrites to extract static condition for dynamic ones to help some tool optimization.
This should be formally equivalent to previous code.

### DIFF
--- a/core/cva6.sv
+++ b/core/cva6.sv
@@ -306,7 +306,7 @@ module cva6
     parameter type x_commit_t = `X_COMMIT_T(CVA6Cfg, hartid_t, id_t),
     parameter type x_result_t = `X_RESULT_T(CVA6Cfg, hartid_t, id_t, writeregflags_t),
     parameter type cvxif_req_t =
-    `CVXIF_REQ_T(CVA6Cfg, x_compressed_req_t, x_issue_req_t, x_register_req_t, x_commit_t),
+    `CVXIF_REQ_T(CVA6Cfg, x_compressed_req_t, x_issue_req_t, x_register_t, x_commit_t),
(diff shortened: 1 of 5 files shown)


core/include/ariane_pkg.sv
```diff
@@ -856,7 +856,7 @@ package ariane_pkg;
   // ----------------------
   // Avoid negative array slices when defining parametrized sizes
   function automatic int unsigned avoid_neg(int n);
-    return (n < 0) ? 0 : n;
+    return (n < 0) ? 32'd0 : n;
```
cathales: I think 0 should be kept here as the output is an int, not a bit vector.

core/alu.sv
```diff
@@ -409,8 +409,10 @@ module alu
         result_o = {
           {32{fu_data_i.operand_b[15]}}, {fu_data_i.operand_b[15:0]}, {fu_data_i.operand_a[15:0]}
         };
-      if (fu_data_i.operation == UNZIP && CVA6Cfg.IS_XLEN32) result_o = unzip_gen;
-      if (fu_data_i.operation == ZIP && CVA6Cfg.IS_XLEN32) result_o = zip_gen;
+      if (CVA6Cfg.IS_XLEN32) begin
```
cathales, line 417: I indeed prefer this 👍

core/include/ariane_pkg.sv
oliverbm67: I rollback this change

oliverbm67:
Would it  be possible to get the FPGA boot logs to see what could be going wrong ?

Issue #3075, opened 2025-08-14
[hpdcache subsystem] Fetch issue when CVA6Cfg.AxiDataWidth < CVA6Cfg.ICACHE_LINE_WIDTH


khandelwaltanuj:
### Is there an existing CVA6 bug for this?

- [x] I have searched the existing bug issues

### Bug Description

Hello @cfuguet 

I am using following parameters: 

  localparam CVA6ConfigAxiDataWidth = 128;
  IcacheByteSize: unsigned'(32768),
  IcacheSetAssoc: unsigned'(8),
  IcacheLineWidth: unsigned'(512),

I have following scenario. 

I have a bootrom of word size 64 bits. It is connected to axi2rom which translate words from 64 bits -> 64 bits. 

After the data goes into an industrial NOC from 64 bits to 128 bits. Here NoC alignes the data according to the address. For ex in the case of addr 0x0, data is in 64 LSBits and in the case of addr 0x8 data is aligned to the MSB (127: 64 bits). 

When the data arrives in hpdcache subsystem. The entry data is 128 bits. Cache line is 512 bits.  It goes in the following code. 

https://github.com/openhwgroup/cva6/blob/bcb0f7de8654c8ad0b260d53a2ce070152d10c9f/core/cache_subsystem/cva6_hpdcache_subsystem_axi_arbiter.sv#L170C9-L170C57

At this line, I feel there is an issue, it doesnt chose anyword and I get instruction fault: 
https://github.com/openhwgroup/cva6/blob/bcb0f7de8654c8ad0b260d53a2ce070152d10c9f/core/cache_subsystem/cva6_hpdcache_subsystem_axi_arbiter.sv#L214

I was expecting a code something like this: 
https://github.com/openhwgroup/cva6/blob/bcb0f7de8654c8ad0b260d53a2ce070152d10c9f/core/cache_subsystem/cva6_hpdcache_subsystem_axi_arbiter.sv#L224

For the case of CVA6ConfigAxiDataWidth = 128, I have made following changes at line 214 (cva6_hpdcache_subsystem_axi_arbiter.sv) and it seems to work. 

      always_comb begin : icache_miss_resp_data_comb
        if (!icache_miss_req_rdata.mem_req_cacheable) begin
          automatic logic [ICACHE_CL_WORD_INDEX - 1:0] icache_miss_word_index;
          automatic logic [63:0] icache_miss_word;
[...]

PR #3082, opened 2025-08-21
Fix: right-align uncacheable reads from the Icache when AxiDataWidth > 64

cfuguet:
This PR fixes an issue when AxiDataWidth is greater than 64 bits but smaller than Icache linewidth on the cva6_hpdcache_cachesubsystem.

In such cases, when issuing uncacheable reads (e.g. read a ROM), the Icache expects the 8 bytes word to be right-aligned on the response interface.

It addresses issue #3075.


khandelwaltanuj:
Hi @cfuguet 

Thanks for the fix. I will try this next week and let you know. 

Regards
Tanuj

khandelwaltanuj (approved):
Hi César, 

I can confirm you that the fix has worked. 

Regards
Tanuj Khandelwal

PR #3098, opened 2025-09-24
Deprecate unsupported packages

oliverbm67:
Per discussion during CVA6 meetings, it has been agreed to clean up some cva6 configuration packages that are not supported. As a first step, the selected configurations are moved into a dedicated directory. They should be completely removed later.


JeanRochCoulon:
LGTM, I will merge tomorrow if no negative feedbacks

oliverbm67:
I thought that too, but it is included in the gitlab CI. I don't know how we missed that during the CVA6 meeting. That is why I kept cv32_imac_sv32 outside the deprecated list.

PR #3101, opened 2025-09-26
[DRAFT] Remove unsupported config from the CI

oliverbm67:
Following discussion on the CVA6 configuration, this PR start the discussion on how to update both the gitlab and the github CI. cv32a60x and cv32a65x are kept, the 64bits configuration  are switched to cv64a60ax and cv64a6_imafdc_sv39_hpdcache_wb (to also check the writeback version of hpdcache).
List of open points (so far) :

- What to do with config cv32a6_imac_sv32 ? Could it be simply removed ? Are we at risk of letting bugs go unnoticed on the sv32 mode of the MMU ?
- Cache selection : if only the HPD/Icache is kept, is it better to check static WT and WB configuration or a single config that can configure the cache in either mode ? @cfuguet  what is your opinion on this ?
- Any config that should get added to the CI ? Openpiton ? a configuration with hypervisor enabled ?

All inputs are welcome

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,12 +62,12 @@ jobs:
       SPIKE_TANDEM: 1
     strategy:
       matrix:
-        testcase:  [ cv64a6_imafdc_tests ]
-        config:    [ cv64a6_imafdc_sv39_hpdcache, cv64a6_imafdc_sv39_hpdcache_wb, cv64a6_imafdc_sv39_wb, cv64a6_imafdc_sv39 ]
+        testcase:  [ cv64a60ax_tests ]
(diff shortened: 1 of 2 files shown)


cfuguet:
Hi @oliverbm67,

> Cache selection : if only the HPD/Icache is kept, is it better to check static WT and WB configuration or a single config that can configure the cache in either mode ? @cfuguet what is your opinion on this ?

The easier thing is to statically check WT and WB separately. If you want with a single configuration check both dynamically, we need first to integrate in the CVA6 platform a way to tag memory segments as write-through or write-back (e.g. a PMA table) and forward that information to the cache when decoding the address.

PR #3109, opened 2025-10-05
ZED working at 57600 baud rate on UART

marcadrian20:


marcadrian20:
Accidental PR, please ignore

PR #3150, opened 2025-11-28
Agilex hps load image from SD Card

AngelaGonzalezMarino:
Modify bootrom to load FW PAYLOAD from SD card in Agilex7
Add IRQ for SD card in altera APU


J-Mallet:
There is a lot of dead code. We should trim all unused functions from the bootrom.

PR #3154, opened 2025-12-03
Adding core config pkg for cvfpu-uvm multi-tool workflow with Bender

rgiunti:


core/include/cv64a60ax_config_pkg_cvfpu-uvm.sv
```diff
@@ -0,0 +1,124 @@
+/*
+ *  Copyright (c) 2025 CEA*
```
JeanRochCoulon, line 2: To fix Verible issues, you have to run the Verible command specified into CONTRIBUTING.md files !

core/include/cv64a60ax_config_pkg_cvfpu-uvm.sv
```diff
@@ -0,0 +1,124 @@
+/*
```
JeanRochCoulon, line 1: Can you provide information about this new configuration. Why dont you reuse the CEA configuration ? What is the objective to define it ?

JeanRochCoulon:
TO fix the GitHub Actions CI, Eclipse agreement need to be signed-up.

core/include/cv64a60ax_config_pkg_cvfpu-uvm.sv
rgiunti: Hi @JeanRochCoulon,
I used that configuration for the **cvfpu-uvm** project [https://github.com/openhwgroup/cvfpu-uvm.git](url) for which I developed a PR. Originally the configuration file was located in the /fpu_common directory of the cvfpu-uvm project and was named **cv64a60ax_config_pkg.sv** (same as the one in cva6/core/include). To be coherent with the modifications introduced by my PR, I needed to take that file no more from /fpu_common but directly from cva6/core/include directory.
However while doing that I noticed that the two files (the one previously used in /fpu_common and the one in /cva6/core/include) were different:
image
image
That's the reason why I preferred to introduce a copy of that configuration directly in /cva6/core/include with the slight modifications needed for the cvfpu-uvm project, in which cva6 is a wrapper for the testing of fpnew.

PR #3155, opened 2025-12-05
Updating CVA6 for OpenPiton and add support for HPDC

ArnauBigas:
This Pull Request adds support for using the HPDC in OpenPiton. To do so, I have added and updated the L1.5 adapters and hpdcache cache subsystems.

Because previously the L1.5 interface types & enums were in the wt_cache pkg and adapter files, I have moved the typedefs to a l15_pkg provided by openpiton. This allows any cache (and any core) to interface with the L1.5.

The changes are working as tested in OpenPiton.


ArnauBigas:
Hi, this PR is good to go from my side. Are there any other blocking issues?

core/cache_subsystem/cva6_hpdcache_subsystem.sv
```diff
@@ -211,8 +211,14 @@ module cva6_hpdcache_subsystem
     userCfg.mshrSetsPerRam = CVA6Cfg.NrLoadBufEntries < 16 ? 1 : CVA6Cfg.NrLoadBufEntries / 2;
     userCfg.mshrRamByteEnable = 1'b1;
     userCfg.mshrUseRegbank = (CVA6Cfg.NrLoadBufEntries < 16);
+    userCfg.cbufEntries = 4;
```
cfuguet: Hello Arnau, here you can use the WbufDepth parameter.

Maybe later, can add some additional parameters to the CVA6 but in the meanwhile, WbufDepth is the more closely related parameter.

cfuguet:
> Questa tests are failed. And one test raises an assertion error: image

I think that error is the one that has prevented me to merge the latest versions of the HPDcache into the CVA6. As I do not have VCS, I cannot reproduce the error

cfuguet (changes requested):
Besides the change of the cbufDepth parameter, all other changes look good to me.

When all tests pass, IMHO this can be merged

ArnauBigas:
Hi @JeanRochCoulon I have looked at the waves and, with the help of @cfuguet , we have determined that the issue lies in the fact that the assertion which is failing ignores the strb AXI signal, meaning that it is not allowing Xs in non-valid bits of the data bus, something which should be allowed. This is failing with newer HPDC versions because Cesar optimized the write buffer in the HPDC to use flip-flops without reset instead of initializing them to 0.

For context, this is the assert in question:
```
   property AXI4_WDATA_X;
      @(posedge clk) disable iff (!rst_n) axi_assert.psv_axi_cb.w_valid |-> (!$isunknown(axi_assert.psv_axi_cb.w_data));
   endproperty
```

cfuguet:
Thanks, @MikeOpenHWGroup. And it goes without saying that, with all your experience, your opinion is as important as anyone else's -- or even more so.

May I assume that you will consider to accept the PR on the core-v-verif repository ?

Flist.ariane
```diff
@@ -20,24 +20,40 @@
 +incdir+common/local/util/
 +incdir+corev_apu/register_interface/include/
 
+core/include/config_pkg.sv
 core/include/cv64a6_imafdc_sv39_openpiton_config_pkg.sv
 core/include/riscv_pkg.sv
 corev_apu/riscv-dbg/src/dm_pkg.sv
-vendor/openhwgroup/cvfpu/src/fpnew_pkg.sv
+core/cvfpu/src/fpnew_pkg.sv
```
MikeOpenHWGroup, line 27: Does this 'hardcode' a specific version of the CVFPU into CVA6?

Flist.ariane
```diff
@@ -142,25 +157,38 @@ corev_apu/fpga/src/axi_slice/src/axi_r_buffer.sv
...
-vendor/openhwgroup/cvfpu/src/fpnew_divsqrt_multi.sv
-vendor/openhwgroup/cvfpu/src/fpnew_fma_multi.sv
-vendor/openhwgroup/cvfpu/src/fpnew_fma.sv
-vendor/openhwgroup/cvfpu/src/fpnew_noncomp.sv
-vendor/openhwgroup/cvfpu/src/fpnew_opgroup_block.sv
-vendor/openhwgroup/cvfpu/src/fpnew_opgroup_fmt_slice.sv
-vendor/openhwgroup/cvfpu/src/fpnew_opgroup_multifmt_slice.sv
-vendor/openhwgroup/cvfpu/src/fpnew_rounding.sv
-vendor/openhwgroup/cvfpu/src/fpnew_top.sv
+core/cvfpu/src/fpu_div_sqrt_mvp/hdl/defs_div_sqrt_mvp.sv
```
MikeOpenHWGroup, line 160: Same question: does this 'hardcode' a specific version of the CVFPU into CVA6?

Flist.ariane
cfuguet: Hi @MikeOpenHWGroup, I verified now, the Flist.cva6 file refers to the same file.

Flist.ariane
cfuguet: Same here

Flist.ariane
cfuguet: The vendor/openhwgroup/cvfpu does not exist anymore

cfuguet: @ArnauBigas is not possible to use directly the core/Flist.cva6 ? The Flist.ariane is an old manifest file. This could be the opportunity to remove it. I know that the Flist.ariane is flatten while the Flist.cva6 uses environment variables and other Flist inclusion. If that is the issue, you can use the utils/flist_flattener.py script

Jbalkind: I would really rather not. People break our build all the time when we use the Flist.cva6

cfuguet: Ok, but at the same time, if people changes the Flist.cva6 it is because there are new dependencies. Hence, a modification of the Flist.ariane is required.

Jbalkind: Yes I'm aware. I would still prefer we don't switch

cfuguet (approved):

cainria:
We have no artifacts from GitHub CI. Thales CI usually provides more artifact for debugging but it seems currently down

ArnauBigas:
So taking a look at the logs I've seen that some errors are due to the missing `_zifencei` isa string on `cva32` targets, so I guess I'll add that to all targets.

However, I've seen that in tests that were previously passing in this merge request, after updating spike no longer pass. This is due to a mismatch in which spike is raising an exception while the RTL isn't. Here is an example:

```
mismatches:
- 0: null
  core:
    insn: 000000000000e398
    insn_disasm: c.sd    a4, 0(a5)
    mode: 3
    pc_rdata: 0000000080003ab4
    pc_wdata: 0
    rd1_addr: 0
    rd1_rdata: 0
    rs1_addr: 000000000000000f
    rs1_rdata: 0000000080006000
    rs2_addr: 000000000000000e
    rs2_rdata: 0000000020001c01
    trap: 0
  reference_model:
    insn: 000000000000e398
    insn_disasm: c.sd    a4, 0(a5)
    mode: 3
    pc_rdata: 0000000080003ab4
    pc_wdata: 0
    rd1_addr: 0
    rd1_rdata: 0
    rs1_addr: 1
    rs1_rdata: 0000000080006000
    rs2_addr: 0
    rs2_rdata: 0000000020001c01
    trap: 1
```

Looking at other logs of the same test, it seems like spike is throwing an illegal instruction exception. I am not sure what could be causing this...

AyoubJalali:
> So taking a look at the logs I've seen that some errors are due to the missing `_zifencei` isa string on `cva32` targets, so I guess I'll add that to all targets.
> 
> However, I've seen that in tests that were previously passing in this merge request, after updating spike no longer pass. This is due to a mismatch in which spike is raising an exception while the RTL isn't. Here is an example:
> 
> ```
> mismatches:
> - 0: null
>   core:
>     insn: 000000000000e398
>     insn_disasm: c.sd    a4, 0(a5)
>     mode: 3
>     pc_rdata: 0000000080003ab4
>     pc_wdata: 0
>     rd1_addr: 0
>     rd1_rdata: 0
>     rs1_addr: 000000000000000f
>     rs1_rdata: 0000000080006000
>     rs2_addr: 000000000000000e
>     rs2_rdata: 0000000020001c01
>     trap: 0
>   reference_model:
>     insn: 000000000000e398
>     insn_disasm: c.sd    a4, 0(a5)
>     mode: 3
>     pc_rdata: 0000000080003ab4
>     pc_wdata: 0
>     rd1_addr: 0
>     rd1_rdata: 0
>     rs1_addr: 1
>     rs1_rdata: 0000000080006000
>     rs2_addr: 0
>     rs2_rdata: 0000000020001c01
>     trap: 1
> ```
> 
> Looking at other logs of the same test, it seems like spike is throwing an illegal instruction exception. I am not sure what could be causing this...

Hey before that I see an AXI assertion failed here :  

image

And to fix the mismatch problem between Spike & RTL, you should add _zifencei in the isa string in [Spike.yaml](https://github.com/openhwgroup/cva6/blob/0e04a86b1316dfcd59b5f17298260cc5ea10af3e/config/gen_from_riscv_config/cv32a65x/spike/spike.yaml#L15-L16) for target related

ArnauBigas:
Hi, I was taking a look at some of the tests and I found that the core is setting the `valid` to `0` before receiving the `ready`, something which is not supported in the HPDC. Seems to happen when there is a transition in the load unit to the `WAIT_FLUSH` state.

imatge

This triggers some asserts in the HPDC, but the test passes anyway. I'm not sure if it's related to the tests which are failing.

ArnauBigas:
Also I have to say I'm still having a lot of trouble replicating the problems in the CI, I see different errors/messages (or none at all)... And the tests that worry me, with the cv64 target, don't have any assert issues, just the mismatch with the spike.

I would really appreciate some help.

JeanRochCoulon:
By analysing the logs of 64bit tests, I understand that the test terminates well but the simulation does not stop. Then a trap triggers which creates discrepancy between RTL and Spike. You could look at rvfi_tracer.sv in the termination condition. 

image

ArnauBigas:
Hi, sorry, I've been taking a look for a while but I'm unable to find or understand the issue...

cfuguet:
> Hi, sorry, I've been taking a look for a while but I'm unable to find or understand the issue...

I'm in the same position here...

I propose to de-correlate the issues. It looks like the problem comes from an incompatibility of the latest version of the core-v-verif submodule with the CV64A6 (the RV32 variant does not seem to have an issue).

I'm opening a new PR #3248, which only tries to update the core-v-verif. Once we get that solve, normally we will be able to rebase this PR on it. Otherwise, this PR is trying to solve to many unrelated issues...

ArnauBigas:
Hi, I have rebased and pushed the branch now that #3248 has been merged, there should be no issues in the CIs. Could you run the CIs for me? Thanks.

cfuguet:
Hi @ArnauBigas, I opened another PR #3277 to update the HPDcache submodule. There are still some issues being solved. When that will be done, you can rebase.

Normally after that, this PR will only add OpenPiton related modifications.

cfuguet:
Hi @ArnauBigas, it should be good now to rebase and try again. The HPDcache submodule has been updated

cfuguet:
Hi @JeanRochCoulon,

This looks good. From my side, this can be merged.

Thanks

JeanRochCoulon:
Long job @ArnauBigas and @cfuguet, but Now it works. Great ! I merge !

cfuguet:
Great ! Thanks :) :tada:

PR #3160, opened 2025-12-11
Refresh cva6_requirements_specification.rst

jquevremont:
Refresh this document with several changes:

HPDCache superseding legacy caches
New supported extensions
Transition to OpenHW Foundation

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -271,7 +271,7 @@ spyglass:
     - cp spyglass/reference_summary.rpt artifacts/lint_reports
     - python3 .gitlab-ci/scripts/report_spyglass_lint.py spyglass/reference_summary.rpt artifacts/lint_reports/cva6_lint_lint_rtl/summary.rpt
 
-cvxif-regression:
+.cvxif-regression:
   extends:
     - .synthesis_test
   variables:
--- a/core/decoder.sv
+++ b/core/decoder.sv
@@ -1258,9 +1258,9 @@ module decoder
               unique case (instr.itype.funct3)
                 3'b001: begin
                   if (instr.instr[31:25] == 7'b0110000) begin
-                    if (instr.instr[21:20] == 2'b10) instruction_o.op = ariane_pkg::CPOPW;
-                    else if (instr.instr[21:20] == 2'b00) instruction_o.op = ariane_pkg::CLZW;
-                    else if (instr.instr[21:20] == 2'b01) instruction_o.op = ariane_pkg::CTZW;
+                    if (instr.instr[24:20] == 5'b00010) instruction_o.op = ariane_pkg::CPOPW;
+                    else if (instr.instr[24:20] == 5'b00000) instruction_o.op = ariane_pkg::CLZW;
+                    else if (instr.instr[24:20] == 5'b00001) instruction_o.op = ariane_pkg::CTZW;
                     else illegal_instr_bm = 1'b1;
                   end else if (instr.instr[31:26] == 6'b000010) begin
                     instruction_o.op = ariane_pkg::SLLIUW; // Shift Left Logic by Immediate (Unsigned Word)
--- a/core/issue_read_operands.sv
+++ b/core/issue_read_operands.sv
@@ -538,7 +538,7 @@ module issue_read_operands
   always_comb begin
     for (int unsigned i = 0; i < CVA6Cfg.NR_SB_ENTRIES; i++) begin
       fwd_res[i] = fwd_i.sbe[i].result;
-      fwd_res_valid[i] = fwd_i.sbe[i].valid;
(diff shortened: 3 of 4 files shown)


docs/02_cva6_requirements/cva6_requirements_specification.rst
```diff
@@ -559,10 +564,14 @@ size/ways configurations may be implemented in the design.
...
 
 These L1WTD requirements apply to the legacy WT cache from PULP.
-They also apply to the HPDCache that will supersede the legacy PULP caches.
+
+They also apply to the HPDCache, except L1W‑80, that is superseding the legacy PULP caches.
 Note that the HPDCache has more features, such as write-back options, that
 are beyond the scope of this specification.
 
+.. attention::
+   L1W-80 (replacement policy) does not apply to HPDCache.
```
cfuguet: The HPDcache supports either LFSR or Pseudo-LRU. So L1W-80 could apply to the HPDcache I guess.

docs/02_cva6_requirements/cva6_requirements_specification.rst
```diff
@@ -559,10 +564,14 @@ size/ways configurations may be implemented in the design.
 The design will support one replacement policy allowed by L1W-80.
 
 These L1WTD requirements apply to the legacy WT cache from PULP.
-They also apply to the HPDCache that will supersede the legacy PULP caches.
+
+They also apply to the HPDCache, except L1W‑80, that is superseding the legacy PULP caches.
```
cfuguet: Reading the requirements, I have a question about the L1W‑20. It says: L1WTD shall not change the order of write accesses to the external memory with respect to the order of write accesses (stores) received from the CVA6 core. For this to be true, it means that we want to implement a TSO (Total Store Order) consistency model, which is supported by RISC-V, but it is not the default one, which is Weak Ordering.

In the weak ordering, stores on non-overlapping addresses can be forwarded to the memory in a different order than the program order. Currently the HPDcache (in WT mode), and the PULP WT cache implement the weak ordering.

Do you really want to support TSO ? Or did you mean that the writes in overlapping addresses need to be issued to the memory in order ? If it is the second case, we need just to add the overlapping notion in the requirement. If it is the first case, we need to make some RTL :)

Note: in the weak ordering, if the software needs to make stores to be issued to the memory in program order, it needs to enforce that explicitly using fences. The HPDcache supports fences.

docs/02_cva6_requirements/cva6_requirements_specification.rst
```diff
@@ -621,6 +630,10 @@ size/ways configurations may be implemented in the design.
 
 The design will support one replacement policy allowed by L1I-50.
 
+These requirement to the HPICache, a modified verision of HPICache, that
```
cfuguet: Typos correction: These « requirementS » to the HPICache, a modified « version » of the « HPDcache »

docs/02_cva6_requirements/cva6_requirements_specification.rst
jquevremont: With L1W-10 and L1W-20, I have clumsily tried to capture requirements to express that we need a write-through cache. Weak ordering is OK, we do not need TSO.

docs/02_cva6_requirements/cva6_requirements_specification.rst
jquevremont: Excellent! Removing this comment...

docs/02_cva6_requirements/cva6_requirements_specification.rst
jquevremont: Thanks for spotting this typo. I have also removed the caveat about the replacement policy, as for the data cache.

docs/02_cva6_requirements/cva6_requirements_specification.rst
jquevremont: @cfuguet, what about rewording L1W-20 with "L1WTD shall support RISC-V Weak Memory Ordering (RVWMO)"?

cfuguet: Yes !

docs/02_cva6_requirements/cva6_requirements_specification.rst
```diff
@@ -621,6 +627,9 @@ size/ways configurations may be implemented in the design.
 
 The design will support one replacement policy allowed by L1I-50.
 
+These requirements to the HPICache, a modified verision of HPICache, that
```
cfuguet: These requirements "apply" to the HPIcache...

cfuguet: a modified "verision -> version" of the "HPIcache -> HPDcache"

cfuguet (changes requested):

Issue #3167, opened 2025-12-22
[BUG] <title>Core freezing when enabling CV-X-IF without RVC extension


bugavix:
### Is there an existing CVA6 bug for this?

- [x] I have searched the existing bug issues

### Bug Description

It appears that when enabling the CV-X-IF interface while disabling the RVC extension, the core freezes during the initialization phase of the debug sequence in the testbench.

I managed to find out that this was caused by a bug inside the Decode stage: the compressed‑instruction channel driver of the CV-X-IF inside the Decode stage is only generated if the RVC extension is enabled. However, this module still drives signals inside the stage even when RVC is off. I proposed a fix in the PR below, although I am not sure it is the cleanest way to address the issue.

cainria:
It seems this issue has been fixed by #3168 
Can we close this issue?

PR #3168, opened 2025-12-22
Fix a bug that causes the core to freeze when enabling CV-X-IF while disabling RVC extension

bugavix:
This is an attempt to fix the bug described in #3167.

It appears that when enabling the CV-X-IF interface while disabling the RVC extension, the core freezes during the initialization phase of the debug sequence in the testbench.

I managed to find out that this was caused by a bug inside the Decode stage: the compressed‑instruction channel driver of the CV-X-IF inside the Decode stage is only generated if the RVC extension is enabled. However, this module still drives signals inside the stage even when RVC is off. I proposed a fix in the PR below, although I am not sure it is the cleanest way to address the issue.


core/id_stage.sv
```diff
@@ -289,14 +289,16 @@ module id_stage #(
...
     is_compressed_deco = is_compressed_rvc;
-    if (CVA6Cfg.CvxifEn) begin
-      is_illegal_deco[0]    = is_illegal_cvxif_o;
-      instruction_deco[0]   = instruction_cvxif_o;
-      is_compressed_deco[0] = is_compressed_cvxif_o;
-    end else if (!CVA6Cfg.CvxifEn && (CVA6Cfg.RVZCMP || CVA6Cfg.RVZCMT)) begin
-      is_illegal_deco[0]    = is_illegal_cvxif_i;
-      instruction_deco[0]   = instruction_cvxif_i;
-      is_compressed_deco[0] = is_compressed_cvxif_i;
+    if (CVA6Cfg.RVC) begin
```
bugavix, line 304: In order to have a cleaner code, I think these assignments should be made inside the previous if block (lines 156-281). This is only a quick fix.

PR #3178, opened 2026-01-06
Support for Cache Management Operations (Zicbom) Operations with HPDCache: CBO.INVAL, CBO.CLEAN, CBO.FLUSH

WorldofJARcraft:
## Overview
This PR adds support for Zicbom (CBO.INVAL, CBO.CLEAN, CBO.FLUSH) as of [Version 1.0.0 of the CMO specification](https://docs.riscv.org/reference/isa/v20240411/unpriv/cmo.html).

## Changes
TL;DR the following changes are proposed:
- `decoder.sv`: decoding, generation of exceptions for CBO instructions currently supported 
- `csr_regfile.sv`: management of enable / configuration bits for CBO in `menvcfg`, `senvcfg`, `henvcfg` CSRs
- configuration system: new option `RVZiCbom`, only enabled for hpdcache configs currently
- execute stage / store buffer: CMOs are handled akin to `sb` stores, i.e., no misalignment exceptions but same handling of address translation and PMP checks; CMO operations commit in the store buffer as soon as accepted *and completed* in hpdcache. CMOs commit in order and one can use a `fence` instruction to ensure all outstanding CMOs have committed.
- cache interface: indicates CMO operation if needed
- `cva6_ptw.sv`: I encountered an issue where the page table walker would request a PTE, then encounter a flush and request a new PTE. This caused an issue in conjunction with the HPDcache in which `rvalid` was first signaled for the old PTE and then for the new PTE, causing the PTW to use the incorrect PTE for a translation. I fixed this by waiting for `rvalid` in case of a flush in the PTW, this is in a separate commit e1aeba12bba2eecac24f3a9f7343b467cfba8a88.

## Testing environment
I have tested these patches with a slightly older version of cva6 in conjunction with a Xilinx AXI Ethernet subsystem with DMA (non-cache coherent). I used CBO to invalidate and flush DMA descriptors and transferred buffers from both u-boot and Linux with no issue. Software and hardware of the testing environment is [open source](https://github.com/cispa/CVA6-Vivado-Project-with-Xilinx-AXI-Ethernet).

### DIFF
--- a/core/cva6_mmu/cva6_ptw.sv
+++ b/core/cva6_mmu/cva6_ptw.sv
@@ -109,6 +109,7 @@ module cva6_ptw
     WAIT_RVALID,
     PROPAGATE_ERROR,
     PROPAGATE_ACCESS_ERROR,
+    KILL_REQ,
     LATENCY
   }
       state_q, state_d;
@@ -133,6 +134,8 @@ module cva6_ptw
   logic global_mapping_q, global_mapping_n;
   // latched tag signal
   logic tag_valid_n, tag_valid_q;
+  // latched kill signal
+  logic kill_req_n, kill_req_q;
   // register the ASID
   logic [CVA6Cfg.ASID_WIDTH-1:0] tlb_update_asid_q, tlb_update_asid_n;
   // register the VMID
@@ -155,8 +158,6 @@ module cva6_ptw
   // directly output the correct physical address
   assign req_port_o.address_index = ptw_pptr_q[CVA6Cfg.DCACHE_INDEX_WIDTH-1:0];
   assign req_port_o.address_tag   = ptw_pptr_q[CVA6Cfg.DCACHE_INDEX_WIDTH+CVA6Cfg.DCACHE_TAG_WIDTH-1:CVA6Cfg.DCACHE_INDEX_WIDTH];
-  // we are never going to kill this request
-  assign req_port_o.kill_req = '0;
   // we are never going to write with the HPTW
   assign req_port_o.data_wdata = '0;
   // we only issue one single request at a time
@@ -239,6 +240,7 @@ module cva6_ptw
   end
 
   assign req_port_o.tag_valid = tag_valid_q;
+  assign req_port_o.kill_req  = (state_q == KILL_REQ);
 
   logic allow_access;
 
@@ -294,6 +296,7 @@ module cva6_ptw
     // default assignments
     // PTW memory interface
     tag_valid_n             = 1'b0;
+    kill_req_n              = kill_req_q;
     req_port_o.data_req     = 1'b0;
     req_port_o.data_size    = 2'(CVA6Cfg.PtLevels);
     req_port_o.data_we      = 1'b0;
@@ -401,8 +404,15 @@ module cva6_ptw
         if (req_port_i.data_gnt) begin
           // send the tag valid signal one cycle later
           tag_valid_n = 1'b1;
-          state_d     = PTE_LOOKUP;
+          // if the request has been flushed, kill it one cycle later
+          state_d = (kill_req_q || flush_i) ? KILL_REQ : PTE_LOOKUP;
         end
+      end
+
+      KILL_REQ: begin
+        kill_req_n = 1'b0;
+        // even when killed, the request sends a response
+        state_d = WAIT_RVALID;
       end
 
       PTE_LOOKUP: begin
@@ -628,9 +638,10 @@ module cva6_ptw
       // 1. in the PTE Lookup check whether we still need to wait for an rvalid
       // 2. waiting for a grant, if so: wait for it
       // if not, go back to idle
-      if (((state_q inside {PTE_LOOKUP, WAIT_RVALID}) && !data_rvalid_q) || ((state_q == WAIT_GRANT) && req_port_i.data_gnt))
-        state_d = WAIT_RVALID;
-      else state_d = LATENCY;
+      if ((state_q inside {PTE_LOOKUP, WAIT_RVALID}) && !data_rvalid_q) state_d = WAIT_RVALID;
+      else if (state_q != WAIT_GRANT) state_d = LATENCY;
+
+      kill_req_n = (state_q == WAIT_GRANT);
     end
   end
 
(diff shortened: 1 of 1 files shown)


core/csr_regfile.sv
```diff
@@ -2667,6 +2763,10 @@ module csr_regfile
       icache_q        <= {{CVA6Cfg.XLEN - 1{1'b0}}, 1'b1};
       mcountinhibit_q <= '0;
       acc_cons_q      <= {{CVA6Cfg.XLEN - 1{1'b0}}, CVA6Cfg.EnableAccelerator};
+      if (CVA6Cfg.RVZiCbom) begin
+        mcbie_q  <= riscv::CBIE_INVAL;
```
cfuguet, line 2767: I did not find in the specification which should be the reset value for the cbie and cbcfe registers. Where did you find this information ? Or are we assuming here that the value is implementation-defined ?

core/csr_regfile.sv
```diff
@@ -2678,6 +2778,10 @@ module csr_regfile
         sscratch_q   <= {CVA6Cfg.XLEN{1'b0}};
         stval_q      <= {CVA6Cfg.XLEN{1'b0}};
         satp_q       <= {CVA6Cfg.XLEN{1'b0}};
+        if (CVA6Cfg.RVZiCbom) begin
+          scbie_q  <= riscv::CBIE_INVAL;
```
cfuguet, line 2782: same comment as above

core/csr_regfile.sv
```diff
@@ -2701,6 +2805,10 @@ module csr_regfile
         vstval_q                 <= {CVA6Cfg.XLEN{1'b0}};
         vsatp_q                  <= {CVA6Cfg.XLEN{1'b0}};
         en_ld_st_g_translation_q <= 1'b0;
+        if (CVA6Cfg.RVZiCbom) begin
+          hcbie_q  <= riscv::CBIE_INVAL;
```
cfuguet, line 2809: same comment as above

core/include/cv64a60ax_config_pkg.sv
cfuguet, line 1: Could you please revert the formatting modifications in this file as they are not related to the modifications for the  Zicbom extension ?

core/cva6_mmu/cva6_ptw.sv
cfuguet, line 1: This is interesting. If I'm not mistaken, this is a bugfix unrelated to Zicbom. It is related to the fact that the HPDcache does not allow to abandon an active but not granted request. If there is a PTE request valid but not yet granted, and the request is flushed, we need to wait anyway the grant signal, before abandoning the request.

Currently, you also wait for the rvalid signal, but we could actually kill the request on the next cycle (after the grant) using the kill signal in the dreq interface. This way we can actually avoid loading in the cache data related to a flushed request (reduce the surface for side-channel attacks...).

IMHO, I would do a separate PR for this modification. And if possible, I would use the kill signal to actually kill the PTE lookup sent in the previous cycle (when the request is flushed in the WAIT_GRANT state). What do you think ?

core/include/deprecated_packages/cv64a6_imadfcv_sv39_polara_config_pkg.sv
```diff
@@ -156,6 +157,7 @@ package cva6_config_pkg;
       DcacheSetAssoc: unsigned'(CVA6ConfigDcacheSetAssoc),
       DcacheLineWidth: unsigned'(CVA6ConfigDcacheLineWidth),
       DcacheFlushOnFence: bit'(CVA6ConfigDcacheFlushOnFence),
+      DcacheFlushOnFenceI: unsigned'(CVA6ConfigDcacheFlushOnFenceI),
```
cfuguet: cast to bit

core/include/ariane_pkg.sv.orig
cfuguet, line 1: Remove this file

core/commit_stage.sv.orig
cfuguet, line 1: Remove this file

cfuguet (changes requested):

core/csr_regfile.sv
WorldofJARcraft: In my understanding, the reset values for these registers are *implementation-specified* (more or less by omission: https://github.com/riscv/riscv-isa-manual/issues/820).
I picked defaults that made sense to me.
In practice, privileged software will / should initialize them early in the boot process.

core/include/cv64a60ax_config_pkg.sv
WorldofJARcraft: Done

core/include/ariane_pkg.sv.orig
WorldofJARcraft: Done

core/commit_stage.sv.orig
WorldofJARcraft: Done

core/cva6_mmu/cva6_ptw.sv
WorldofJARcraft: I have tried killing the request before, which did not resolve the issue.
However, I believe I did not wait for the grant before raising kill.
I will test whether this works.

I included the patch here, as my test system would otherwise not boot into Linux with hpdcache write-back, making it impossible to test my changes. I can remove the commit here and raise a separate pull request.

cfuguet: Yes please do a separate pull request. Thanks for spotting this issue. It is an actual bugfix but as it is unrelated to Zicbom, I think it is better to have it in a separate PR.

core/csr_regfile.sv
cfuguet: Understood. I agree, and your default values make also sense to me.

core/cva6_mmu/cva6_ptw.sv
WorldofJARcraft: I have removed the commit from this PR.
I will open a new PR for this fix.

WorldofJARcraft: New PR is here: https://github.com/openhwgroup/cva6/pull/3179

cfuguet (approved):
It is all good for me. In my opinion, this PR can be merged. @JeanRochCoulon, do you approve ?

PR #3179, opened 2026-01-07
When using HPDCache, kill cache request from CVA6 PTW in case of flush to prevent spurious rvalid

WorldofJARcraft:
I encountered an issue where the page table walker would request a PTE, then encounter a flush and request a new PTE. This caused an issue in conjunction with the HPDcache in which rvalid was first signaled for the old PTE and then for the new PTE, causing the PTW to use the incorrect PTE for a translation.

I did not fully investigate the cause of this issue. I assume that the root cause is first raising and then lowering the `data_req` flag without sending `tag_valid` or `kill_req` accordingly.

This PR makes the PTW send a kill request in that scenario, preventing the spurious `rvalid`.

### DIFF
--- a/core/cva6_mmu/cva6_ptw.sv
+++ b/core/cva6_mmu/cva6_ptw.sv
@@ -408,9 +408,9 @@ module cva6_ptw
       end
 
       KILL_REQ: begin
-          kill_req_n = 1'b0;
-          // even when killed, the request sends a response
-          state_d = WAIT_RVALID;
+        kill_req_n = 1'b0;
+        // even when killed, the request sends a response
+        state_d = WAIT_RVALID;
       end
 
       PTE_LOOKUP: begin
@@ -636,10 +636,8 @@ module cva6_ptw
       // 1. in the PTE Lookup check whether we still need to wait for an rvalid
       // 2. waiting for a grant, if so: wait for it
       // if not, go back to idle
-      if ((state_q inside {PTE_LOOKUP, WAIT_RVALID}) && !data_rvalid_q)
-        state_d = WAIT_RVALID;
-      else if (state_q != WAIT_GRANT)
-        state_d = LATENCY;
+      if ((state_q inside {PTE_LOOKUP, WAIT_RVALID}) && !data_rvalid_q) state_d = WAIT_RVALID;
+      else if (state_q != WAIT_GRANT) state_d = LATENCY;
 
       kill_req_n = (state_q == WAIT_GRANT);
     end


core/cva6_mmu/cva6_ptw.sv
```diff
@@ -392,14 +395,16 @@ module cva6_ptw
...
+      WAIT_GRANT, KILL_REQ: begin
         // send a request out
         req_port_o.data_req = 1'b1;
         // wait for the WAIT_GRANT
         if (req_port_i.data_gnt) begin
           // send the tag valid signal one cycle later
-          tag_valid_n = 1'b1;
-          state_d     = PTE_LOOKUP;
+          // exception: when killing the request, do not send tag valid
+          tag_valid_n = (state_q == PTE_LOOKUP);
```
cfuguet: This will never be true. Current state (state_q) is either WAIT_GRANT or KILL_REQ at this point

cfuguet: ```tag_valid_n = (state_q != KILL_REQ);```

core/cva6_mmu/cva6_ptw.sv
```diff
@@ -629,6 +634,14 @@ module cva6_ptw
       if (((state_q inside {PTE_LOOKUP, WAIT_RVALID}) && !data_rvalid_q) || ((state_q == WAIT_GRANT) && req_port_i.data_gnt))
         state_d = WAIT_RVALID;
       else state_d = LATENCY;
+
+      if(state_q == WAIT_GRANT && CVA6Cfg.DCacheType inside {
```
cfuguet: ```if(state_q == WAIT_GRANT && !req_port_i.data_gnt)```

cfuguet: Let's keep the old behavior: if the waiting for grant but at the same cycle, the FSM receives both grant and flush, then lets handle grant in priority. This is, in this case, do not try to kill the request. Only wait for RVALID: as handled in the lines above (lines 634 to 636):

```
if (((state_q inside {PTE_LOOKUP, WAIT_RVALID}) && !data_rvalid_q) ||
    ((state_q == WAIT_GRANT) && req_port_i.data_gnt))
        state_d = WAIT_RVALID;
```

core/cva6_mmu/cva6_ptw.sv
WorldofJARcraft: Oops, I missed that. I guess I was thinking about `state_d`. Curiously, it still worked like that in the configuration that I am testing on the FPGA - I will re-synthesize and test again.

WorldofJARcraft:
> I'm trying something on my end. Would you agree that I push into your branch when done ?

Yes, feel free to push your changes if GitHub will let you. Otherwise, you can open a new PR and close this one.

WorldofJARcraft:
@cfuguet I have tested your changes in three configurations:
- cva6 with this patch + the CBO PR + the AXI DMA
- original cva6 with cva6 SDK, WT-Cache (default)
- original cva6 with cva6 SDK, HPDCache-WB

The first and second configurations ran without issue.
For the third configuration, cva6 got stuck in u-boot while copying the boot image from SD card to memory (before Linux is loaded and before I experienced a crash before).

I will re-build the failing configuration with both my first commit from this branch and my original fix and see if the hang can be reproduced.

WorldofJARcraft:
@cfuguet I have been able to narrow down the problem.

First, a clarification on my testing environments:
- "original cva6" refers to the SoC from this repository
- the second testbed is a SoC variant with Xilinx AXI DMA and a few other differences

Here are the new insights:
- The problem does not seem to be related to this PR at all. Going back to the current master (aa4f5a548b33f7edff63483347fc8510aa9ad5fd) and building an image with HPDCache WB (`TARGET_CFG=cv64a6_imafdc_sv39_hpdcache_wb make fpga`) causes the same boot failure (hang while reading image from SD in u-boot).
- The problem goes away when merging the changes from https://github.com/openhwgroup/cva6/pull/3178 (i.e., I can get past the hang in u-boot and boot Linux with no issue).
- Merging the latest changes from this PR on top of https://github.com/openhwgroup/cva6/pull/3178 still results in an image that boots Linux without issue. Hence, the changes here are likely not to blame.

Here is my hypothesis: The only relevant change by https://github.com/openhwgroup/cva6/pull/3178 is disabling DcacheFlushOnFence in favor of CMO. In the current cva6-sdk configuration, neither u-boot nor Linux with the default LowRISC Ethernet ever use CMO (I patched this into the AXI DMA driver manually for testing). Thus, the remaining changes should not change the behavior of the CPU.
Looking at the u-boot driver, it issues a store fence before accessing the MMIO registers.
Thus, I suspect that the root cause of the observed hang is somewhere in the handling of fences / flushing the cache immediately before or after uncached reads or writes.

I normally test changes on my testbed with AXI DMA and use TFTP to boot Linux, rather than loading the image from SD.
This skips the step where cva6 hangs.
Thus, I only tested the combined changes for CMO and the PTW fix on the SoC contained in this repository. 
Thereby, I did not encounter this issue during development.

Issue #3185, opened 2026-01-13
[BUG] SystemVerilog port type compatibility issue: packet_type_o enum port declaration triggers DSim PortTypeMismatch


AlexChenIC:
### Is there an existing CVA6 bug for this?

- [x] I have searched the existing bug issues

### Bug Description

**Summary**

Altair DSim 2026.0 simulator elaboration fails due to a type mismatch on the `rv_tracer.packet_type_o` port connection in `ariane_testharness`.

**Observed Error (DSim)**
```
=E:[PortTypeMismatch]:
  ariane_testharness.sv:727 packet_type_o
  formal: enum logic[3:0]{...}[0:0]
  actual: enum logic[3:0]{...}
```

**Relevant Code**
```
corev_apu/instr_tracing/rv_tracer-main/rtl/rv_tracer.sv(line 60)
corev_apu/instr_tracing/rv_tracer-main/include/te_pkg.sv (line 172)
corev_apu/tb/ariane_testharness.sv (line 727)

```

**Root cause (based on code inspection)**

In `corev_apu/.../rv_tracer.sv` (line 60), the port is declared as an unpacked array of an enum type:

```
output te_pkg::it_packet_type_e [N-1:0] packet_type_o,
```

The instantiation parameter `N` is set to 1, while `te_pkg::it_packet_type_e` (in `te_pkg.sv`) is defined as an enumerated type with packed bit-width `[3:0]`. With `N=1`, the port becomes a single-element unpacked array (`[0:0]`) of a scalar enum, which is not type-compatible with a scalar enum connection under strict SystemVerilog typing. DSim therefore correctly flags this as a fatal type error.

Per Mike Thompson’s suggestion (based on `git blame`), could a maintainer please assign this issue to Jean Roch Coulon?

Thanks.

cainria:
It seems to be fixed by https://github.com/openhwgroup/cva6/pull/3186
Can we close this issue?

PR #3186, opened 2026-01-16
Fix packet_type_o port type mismatch in rv_tracer

AlexChenIC:
Ref: #3185

This PR removes the unpacked array dimension ([N-1:0], with N=1) from the
rv_tracer packet_type_o port to avoid a scalar enum vs [0:0] array type
mismatch under strict SystemVerilog typing (e.g. DSim).


MaxCThales:
Hi,
Indeed, @numero-744 summarized the issue well regarding the permanent removal of [N-1:0]. The current encoder is derived from https://github.com/pulp-platform/rv_tracer
and is designed to handle N “blocks.” However, in our modified version, we chose to focus on a block-by-block implementation (with N = 1). With the idea of eventually being able to use the encoder as-is (as a submodule), we wanted to preserve a similar structure, even if it is not currently used (i.e., handling the N > 1 cases).

Removing the [N-1:0] dimension would therefore break this indexed logic, e.g., at line 747 or line 853 of rv_tracer.sv

numero-744:
Have you tried the suggestion from my previous message? (Adding an `[N-1:0]` dimension to the `packet_type` signal at the instantiation place in `corev_apu/fpga/src/ariane_xilinx.sv` and `corev_apu/tb/ariane_testharness.sv`.)

This new dimension can be fixed as `[0:0]` as the instantiation specifies `.N(1)`. Then, it can be indexed as `packet_type[0]` where it is read.

To reduce the number of changed line, one could use an intermediary signal as show in the pseudocode below:

```systemverilog
packet_type_t [0:0] packet_types;
packet_type_t packet_type;

...
.N(1),
...
.packet_type_o (packet_types),
...

assign packet_type = packet_types[0];
```

JeanRochCoulon:
@AlexChenIC   The CI Thales was failed due to a bad synthesis kgates number result. I pushed a fix. I hope that is ok for you !

PR #3188, opened 2026-01-19
Add speculative loads support

LorenzoR01:
This PR introduces support for speculative loads, allowing dual issue of a control flow instruction and a load instruction.
Speculative loads improve Coremark performance by up to 7% with no impact on the CVA6 timing.
We have successfully tested the code on Linux and Splash3 benchmarks.

### DIFF
--- a/.gitlab-ci/scripts/report_benchmark.py
+++ b/.gitlab-ci/scripts/report_benchmark.py
@@ -19,9 +19,9 @@ iterations = None
 # Keep it up-to-date with compiler version and core performance improvements
 # Will fail if the number of cycles is different from this one
 valid_cycles = {
-    "dhrystone_dual": 16379,
+    "dhrystone_dual": 16282,
     "dhrystone_single": 22256,
-    "coremark_dual": 940954,
+    "coremark_dual": 865167,
     "coremark_single": 1294524,
     "dhrystone_cv32a65x": 27329,
     "dhrystone_cv32a60x": 36629,


core/commit_stage.sv
```diff
@@ -69,7 +69,7 @@ module commit_stage
     // Commit buffer of LSU is ready - EX_STAGE
     input logic commit_lsu_ready_i,
     // Transaction id of first commit port - ID_STAGE
```
numero-744, line 71: The comment does not match the declaration anymore, we should update it.

core/issue_stage.sv
```diff
@@ -171,7 +171,9 @@ module issue_stage
     // Information dedicated to RVFI - RVFI
     output logic [CVA6Cfg.NrIssuePorts-1:0][CVA6Cfg.XLEN-1:0] rvfi_rs2_o,
     // Original instruction bits for AES
-    output logic [5:0] orig_instr_aes_bits
+    output logic [5:0] orig_instr_aes_bits,
+    // Signals speculative loads to LSU for non-idempotent load handling - EX_STAGE
+    output logic speculative_load_o
```
numero-744: `speculative_load_o` should be inserted just after `lsu_valid_o` IMO

core/issue_stage.sv
```diff
@@ -312,7 +314,8 @@ module issue_stage
       .stall_issue_o,
       .rvfi_rs1_o              (rvfi_rs1_o),
       .rvfi_rs2_o              (rvfi_rs2_o),
-      .orig_instr_aes_bits     (orig_instr_aes_bits)
+      .orig_instr_aes_bits     (orig_instr_aes_bits),
+      .speculative_load_o
```
numero-744: `speculative_load_o` should be inserted just after `lsu_valid_o` IMO

core/cva6.sv
```diff
@@ -950,7 +953,8 @@ module cva6
       .rvfi_commit_pointer_o(rvfi_commit_pointer),
       .rvfi_rs1_o           (rvfi_rs1),
       .rvfi_rs2_o           (rvfi_rs2),
-      .orig_instr_aes_bits  (orig_instr_aes)
+      .orig_instr_aes_bits  (orig_instr_aes),
+      .speculative_load_o   (speculative_load)
```
numero-744: `speculative_load_o` should be inserted just after `lsu_valid_o` IMO

core/cva6.sv
```diff
@@ -1025,6 +1029,7 @@ module cva6
       .lsu_commit_i            (lsu_commit_commit_ex),           // from commit
       .lsu_commit_ready_o      (lsu_commit_ready_ex_commit),     // to commit
       .commit_tran_id_i        (lsu_commit_trans_id),            // from commit
+      .speculative_load_i      (speculative_load),
```
numero-744: `speculative_load_i` should be inserted just after `lsu_valid_i` IMO

It is connected to issue stage, not commit stage, right?

core/cva6.sv
```diff
@@ -463,6 +465,7 @@ module cva6
   logic [CVA6Cfg.XLEN-1:0] load_result_ex_id;
   logic load_valid_ex_id;
   exception_t load_exception_ex_id;
+  logic speculative_load;
```
numero-744:
```suggestion
  logic speculative_load_id_ex;
```

`speculative_load_id_ex` should be inserted just after `lsu_valid_id_ex` IMO

core/ex_stage.sv
```diff
@@ -113,7 +113,9 @@ module ex_stage
     // Commit queue ready to accept another commit request - COMMIT_STAGE
     output logic lsu_commit_ready_o,
     // Commit transaction ID - COMMIT_STAGE
-    input logic [CVA6Cfg.TRANS_ID_BITS-1:0] commit_tran_id_i,
+    input logic [CVA6Cfg.NrCommitPorts-1:0][CVA6Cfg.TRANS_ID_BITS-1:0] commit_tran_id_i,
+    // Speculative load signal - ISSUE_STAGE
```
numero-744: This comment just repeats the signal name (and the fact that it is a signal).
We should probably stick to the description from issue stage


```suggestion
    // Signals speculative loads to LSU for non-idempotent load handling - EX_STAGE
```

core/ex_stage.sv
```diff
@@ -289,6 +291,7 @@ module ex_stage
   logic [CVA6Cfg.TRANS_ID_BITS-1:0] mult_trans_id;
   logic mult_valid;
 
+  bp_resolve_t resolved_branch;
```
numero-744: `resolved_branch` should be declared just after `branch_result` IMO

core/issue_read_operands.sv
```diff
@@ -134,7 +134,9 @@ module issue_read_operands
     // Information dedicated to RVFI - RVFI
     output logic [CVA6Cfg.NrIssuePorts-1:0][CVA6Cfg.XLEN-1:0] rvfi_rs2_o,
     // Original instruction bits for AES
-    output logic [5:0] orig_instr_aes_bits
+    output logic [5:0] orig_instr_aes_bits,
+    // Signals speculative loads to LSU for non-idempotent load handling - EX_STAGE
+    output logic speculative_load_o
```
numero-744: `speculative_load_o` should be inserted just after `lsu_valid_o` IMO

core/issue_read_operands.sv
```diff
@@ -1098,6 +1100,11 @@ module issue_read_operands
         branch_predict_o      <= issue_instr_i[0].bp;
         if (CVA6Cfg.RVZCMT) is_zcmt_o <= issue_instr_i[0].is_zcmt;
         else is_zcmt_o <= '0;
+        if (CVA6Cfg.SuperscalarEn && CVA6Cfg.SpeculativeSb) begin
+          if (issue_instr_valid_i[1] && issue_instr_i[1].fu == LOAD && issue_ack_o[1]) begin
+            speculative_load_o <= 1'b1;
+          end
+        end
```
numero-744: Two questions here
- Isn't the load speculative only when the first instruction is a branch? (Or is implementing this additional condition in hardware useless?)
- This condition could probably be moved along with the `lsu_data` mux in the execute stage to make things more local. It would avoid adding the `speculative_load_o` signal which goes up into issue stage, then cva6, then execute stage.

core/load_store_unit.sv
```diff
@@ -74,8 +75,11 @@ module load_store_unit
     // Commit queue is ready to accept another commit request - TO_BE_COMPLETED
     output logic commit_ready_o,
     // Commit transaction ID - TO_BE_COMPLETED
-    input logic [CVA6Cfg.TRANS_ID_BITS-1:0] commit_tran_id_i,
-
+    input logic [CVA6Cfg.NrCommitPorts-1:0][CVA6Cfg.TRANS_ID_BITS-1:0] commit_tran_id_i,
+    // Signals speculative loads for non-idempotent load handling - ISSUE_STAGE
+    input logic speculative_load_i,
+    // Result from branch unit - EX_STAGE
+    input bp_resolve_t resolved_branch_i,
```
numero-744, line 82: Does this new input add a critical path inside the execute stage?

core/ex_stage.sv
```diff
@@ -558,6 +563,8 @@ module ex_stage
       .commit_i              (lsu_commit_i),
       .commit_ready_o        (lsu_commit_ready_o),
       .commit_tran_id_i,
+      .speculative_load_i,
```
numero-744: `speculative_load_i` should be inserted just after `lsu_valid_i` IMO

It is connected to issue stage, not commit stage, right?

core/load_store_unit.sv
```diff
@@ -74,8 +75,11 @@ module load_store_unit
     // Commit queue is ready to accept another commit request - TO_BE_COMPLETED
     output logic commit_ready_o,
     // Commit transaction ID - TO_BE_COMPLETED
-    input logic [CVA6Cfg.TRANS_ID_BITS-1:0] commit_tran_id_i,
-
+    input logic [CVA6Cfg.NrCommitPorts-1:0][CVA6Cfg.TRANS_ID_BITS-1:0] commit_tran_id_i,
+    // Signals speculative loads for non-idempotent load handling - ISSUE_STAGE
+    input logic speculative_load_i,
```
numero-744: `speculative_load_i` should be inserted just after `lsu_valid_i` IMO

It is connected to issue stage, not commit stage, right?

core/load_unit.sv
```diff
@@ -74,7 +75,9 @@ module load_unit
     // Store buffer is empty - STORE_UNIT
     input logic store_buffer_empty_i,
     // Transaction ID of the committing instruction - COMMIT_STAGE
```
numero-744, line 76:
```suggestion
    // Transaction ID of the committing instructions - COMMIT_STAGE
```

core/load_unit.sv
```diff
@@ -223,15 +227,17 @@ module load_unit
...
   assign paddr_ni = config_pkg::is_inside_nonidempotent_regions(
       CVA6Cfg, {{52 - CVA6Cfg.PPNW{1'b0}}, dtlb_ppn_i, 12'd0}
   );
-  assign not_commit_time = commit_tran_id_i != lsu_ctrl_i.trans_id;
+  for (genvar i = 0; i < CVA6Cfg.NrCommitPorts; i++) begin : gen_not_commit_time
+    assign not_commit_time[i] = (commit_tran_id_i[i] != lsu_ctrl_i.trans_id);
+  end
   assign inflight_stores = (!dcache_wbuffer_not_ni_i || !store_buffer_empty_i);
-  assign stall_ni = (inflight_stores || not_commit_time) && (paddr_ni && CVA6Cfg.NonIdemPotenceEn);
+  assign stall_ni = (inflight_stores || &not_commit_time) && (paddr_ni && CVA6Cfg.NonIdemPotenceEn);
```
numero-744: Previously, the `not_commit_time` guaranteed that there was no previous instructions as it could be a branch miss or a exception.
Here `&not_commit_time = 1'b0` if the load is the second committing instruction, so there might be an uncommitted instruction before it.

I do not yet understand why is this change needed (e.g.: correctness or performance?) and why it is correct.

numero-744 (commented):
Thanks for the contribution!

Unfortunately, I cannot review LSU's FSM as I am not familiar enough with it to understand the internal changes. I don't know enough about the TLB either.
Still, I'm wondering: in the execute stage, the branch unit answers ~~immediately~~ combinatorially, so do we need to go to a new state in LSU's FSM?

Is there a maximum frequency regression due to the ALU->BRANCH->LSU path, or is the logic of the FSM shallow enough?
How does the performance of this PR compare with the performance of the fast but wrong implementation where we only remove the stall condition from the issue stage?

Below are minor comments about the code.

Cc @Gchauvon

core/load_unit.sv
LorenzoR01: No this change is not needed for speculative loads and now I also realized that this is wrong, so I'm removing this part.

numero-744: Okay, I'm curious to know whether removing this part of the diff changes performance

LorenzoR01: I removed this part, performance are still the same

core/commit_stage.sv
numero-744: This comment is not accurate anymore

core/ex_stage.sv
```diff
@@ -512,14 +514,19 @@ module ex_stage
...
+    lsu_data = lsu_valid_i[0] ? fu_data_i[0] : '0;
     lsu_tinst = tinst_i[0];
+    speculative_load = 1'b0;
 
     if (CVA6Cfg.SuperscalarEn) begin
       if (lsu_valid_i[1]) begin
         lsu_data  = fu_data_i[1];
         lsu_tinst = tinst_i[1];
+        if (CVA6Cfg.SpeculativeSb) begin
+          speculative_load = branch_valid_i[0] ? 1'b1 : 1'b0;
```
numero-744:
```suggestion
          speculative_load = branch_valid_i[0];
```

Let's avoid the "if true then true else false" tautology

Gchauvon:
Great ! Well it's ok for me. 
Cc @JeanRochCoulon

JeanRochCoulon:
Some Thales CI tests are failed. So, as the URL is broken, I will try to give you access.

Issue #3190, opened 2026-01-20
[BUG] Spectre-type Vulnerability and Its Different Variants were Found in CVA6


zxz2004626:
### Is there an existing CVA6 bug for this?

- [x] I have searched the existing bug issues

## Summary

We creatively used `ecall` to delay the resolution of the branch, creating a branch misprediction transient window. In this window, we can access kernel-space data and encode it in different microarchitecture components at U-mode without exception. 

## Bug Description
Hi! Our processor side-channel vulnerability detection work has found Spectre vulnerability and its different variants on CVA6, which allows an attacker to leak kernel-space data from user space without exception. 

CVA6 employs thie Scoreboard for out-of-order issuing. The Scoreboard algorithm won't issue instructions that have RAW conflicts. Therefore, when a B-type instruction is issued, its operands are already ready. So, B-type instructions can be resolved very quickly. As a result, the traditional Spectre vulnerabilities cannot perform the secret access and encoding operation within the CVA6's 1-2 cycles branch midprediction window.

However, we creatively introduce `ecall` to delay the resolution of the branch. `ecall` is a instruction that will implicitly write register `a0`. But in the implementation of CVA6 Scoreboard, `ecall` won't be in conflict with the subsequent instructions that take `a0` as an operand. This give us a chance to issue B-type instruction with operand `a0`. CVA6 will read the old value of `a0` and make a wrong descision on the B-type instrucion. Then, we accessed and encoded the secret on the wrong branch. This takes advantage of the feature of the CVA6 Scoreboard: in-flight intructions in the scoreboard won’t trigger exceptions until they commit, which we have illustrated in detail in previous issue: [An exploitable Meltdown vulnerability was found in CVA6](https://github.com/openhwgroup/cva6/issues/3163).

We provide an end-to-end exploit code and vulnerability explanation below.

```
2f40:	fdiv.d  fa5, fa5, fa4
2f44:	ecall                               # overwrite a0 to a non-zero value
2f48:	bnez    a0, 2f84 <springboard>      # the old value of a0 is zero
2f4c:	lb      t0, 0(a1)                   # a1 contains secret address 0xffffffffffe04000
2f50:	slli    t0, t0, 0x5                 # use secret as index
2f54:	add     t0, t1, t0                  # t1 contains probe array address 0x10000
2f58:	ld      t1, 0(t0)                   # encode secret to probe array
... 
2f84    <springboard>:                      # the true control flow will jump to here
...
```

[...]

Issue #3163, opened 2025-12-14
[BUG] An exploitable Meltdown vulnerability was found in CVA6


zxz2004626:
### Is there an existing CVA6 bug for this?

- [x] I have searched the existing bug issues

### Bug Description

Hi! Our processor side-channel vulnerability detection work has found an efficient and fault-tolerant Meltdown vulnerability on CVA6, which allows an attacker to leak kernel-space data at arbitrary addresses from user space. This vulnerability takes advantage of the feature of the CVA6 scoreboard: in-flight intructions in the scoreboard won’t trigger exceptions until they commit. Thus, we can access privilege data in a transient execution window and encode the privileged data in the microarchitecture.

We provide an end-to-end exploit code and vulnerability explanation below.

```asm
2f40:    fdiv.d	fa5,fa5,fa4
2f44:    lb t1, 0(a0)                # a0 contains secret address 0xffffffffffe06f30, secret data is 0x30
2f48:    slli t1, t1, 5              # encode secret to a probe array (index = secret << 5, t1 = 0x30 << 5 = 0x600)
2f4c:    add t0, t1, t0              # encode secret to a probe array (base = 0x10000, t0 = base + t1 = 0x10600)
2f50:    ld t1, 0(t0)                # encode the secret to cache
```
The code runs under U-mode. The `lb` at `0x2f44` tries to access `0xffffffffffe06f30`, which is a kernel-space address. After the secret data is loaded, it will be encoded into the cache through the user-space array with the base address of `0x10000`. Due to the long latency of `fdiv` and the function [unit ready status](https://github.com/openhwgroup/cva6/blob/master/core/issue_read_operands.sv#L413), subsequent instructions will be issued during the `fdiv` transition execution window, as the waveform shows.

**G1** represents the commit signals, **G2** represents scoreboard entries and **G3** represents the LSU, Dcache and memory bus behaviors. The first `lsu_valid_i` is triggered by `lb`, and the access address is successfully translated. The secret data is read from the Dcache and forwarded to the scoreboard. Thus, the `slli` and `add` can be issued in sequence. Finally, the secret is encoded in `t0`, which is the load address of` ld` at `0x2f50`. This is a cache-miss address. Thus, Dcache will fill the cacheline by initiating a memory access, as observed in `io_axi4_0_ar_valid`. All of the above processes take place within the `fdiv` transient execution window.

[...]

PR #3191, opened 2026-01-21
Fix the Meltdown and Spectre vulnerabilities that break the isolation between user-space and kernel-space

zxz2004626:
This PR is to fix the https://github.com/openhwgroup/cva6/issues/3163 and https://github.com/openhwgroup/cva6/issues/3190. 

This fix has prohibited the result of the instruction with exception from being forwarding to subsequent instructions in the scoreboard. This ensures that the kernel-space data cannot be transmitted to the user-space code, thereby preventing the secret from being encoded into the microarchitecture.

By invalidating the corresponding [`fwd_res_valid`](https://github.com/openhwgroup/cva6/blob/master/core/issue_read_operands.sv#L541) entry for the scoreboard entries with exceptions, it is possible to defend against attacks of Meltdown, Spectre and their variants.

### DIFF
--- a/Bender.yml
+++ b/Bender.yml
@@ -72,6 +72,10 @@ sources:
         files:
           - core/include/cv32a65x_config_pkg.sv
 
+      - target: cv64a60ax_cvfpu_uvm
+        files:
+          - core/include/cv64a60ax_config_pkg_cvfpu-uvm.sv
+
       # General config infrastructure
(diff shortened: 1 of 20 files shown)


JeanRochCoulon:
Thanks @zxz2004626 for this PR and your OpenHW presentation. See you soon !

PR #3193, opened 2026-01-26
Removed halt on set_pc_commit when accelerator not enabled

youssefekr:
`set_pc_commit_i` and `halt_i` are mutually exclusive conditions, so there is no need to check `halt_i`. The only situation where it may be necessary is if accelerator is enabled (I have not studied this case, maybe it is not necessary either), so I have made the halt check conditional to the presence of an accelerator.


JeanRochCoulon:
@youssefekr   I appreciate your PR. But after reviewing your PR with colleagues, we understand your PR is based on your understanding of the RTL, and not verification tests. As you know, we try to avoid regression to keep the RTL quality. Merging this PR without associated tests would be going in the right direction.

youssefekr:
Yes, closing it now, with issue https://github.com/openhwgroup/cva6/issues/3207 pointing to this PR. Thanks.

Issue #3170, opened 2025-12-29
[BUG] HFENCE.VVMA/HFENCE.GVMA Ignore rd Field Constraint when decode


oChunCai:
### Is there an existing CVA6 bug for this?

- [x] I have searched the existing bug issues

### Bug Description

**Summary**
The **decode.sv** incorrectly accepts HFENCE.VVMA (line 308), and HFENCE.GVMA (line 317) instructions with non-zero `rd` fields, violating the RISC-V specification requirement that `rd` must be `0` for these instructions.

specification:   
[https://riscv.github.io/riscv-isa-manual/snapshot/privileged/#_risc_v_privileged_instruction_set_listings](url)

<img width="1600" height="173" alt="Image" src="https://github.com/user-attachments/assets/e4d76b46-35b4-42c2-bc41-688c4bea5db2" />

**Root Cause**
Similar to the bug fixed in PR #922 and #876, these instructions unconditionally reset the `illegal_instr` to `0` when privilege level checks pass, **ignoring the previous illegal instruction detection** from the outer `rd != 0` check (line 203).

**code:**
<img width="1670" height="323" alt="Image" src="https://github.com/user-attachments/assets/7f97da50-ced1-4de2-a7ee-1ee30e85fdd9" />

PR #3204, opened 2026-02-06
fix:#3170

oChunCai:
this fix has refered to the methods of #933


JeanRochCoulon:
Thnaks @oChunCai for your CVA6 contribution

PR #3215, opened 2026-02-23
make user_config understand inline comments in config

cainria:
Otherwise comments in config make the script fail

### DIFF
--- a/util/user_config.py
+++ b/util/user_config.py
@@ -135,7 +135,7 @@ def build_param(line, value):
     return param_re.sub(r"\g<1>\g<2>\g<3>" + value + r"\g<5>", line)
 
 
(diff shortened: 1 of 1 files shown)


cainria:
Yes!

util/user_config.py
```diff
@@ -135,7 +135,7 @@ def build_param(line, value):
     return param_re.sub(r"\g<1>\g<2>\g<3>" + value + r"\g<5>", line)
 
 
-cfg_re = re.compile(r"^(\s*)(?P<name>\w+)(\s*:\s*)(?P<value>.*?)(,?\s*)$")
+cfg_re = re.compile(r"^(\s*)(?P<name>\w+)(\s*:\s*)(?P<value>.*?)(,?\s*)(//.*)\?$")
```
ASintzoff: What is the purpose of `\?` at the end of the modified expression?

cainria: Good catch, `?` should be used instead to indicate that the comment is optional

PR #3217, opened 2026-02-23
model & cycle_diff: update minstret regex

cainria:
Spike DASM seems to not use pseudo instructions anymore, so the regexes needed to be adjusted accordingly.


cainria:
> Spike DASM seems to not use pseudo instructions anymore

@MiniVoitz can you confirm please?

PR #3220, opened 2026-02-25
Fix/3201 interrupt masking

### LINKED ISSUE #3201: [BUG] Interrupts wrongly masked by fetch side exceptions

### Bug Description

With Questa Processor, we found a scenario where CVA6 raises a fetch side exception (concretely fetch access fault). However, an interrupt is expected. This seems due to the code structure
`if (~ex_i.valid) begin`
`    // synch exceptions relying in fetched instruction word, so must not evaluate since fetch side exception means no instruction word to decode`
`    // interrupts should be evaluated regardless of fetch side exception`
`end`
`// debug request`
So debug requests indeed always take priority, but regular interrupts only take priority over exceptions from decode like illegal instruction or environment call, but a fetch side exception like a fetch access fault simply blocks evaluation of all interrupt conditions.

Kaveesha-98:
Fixes #3201 

### Description
This PR addresses the bug where interrupts were incorrectly masked by fetch side exceptions in the CVA6 core.


JeanRochCoulon:
Thanks @Kaveesha-98 for your contribution. I am running the CI to check everything is ok. I plan to merge asap it is green.

Kaveesha-98:
Thanks, happy to help 😊

PR #3221, opened 2026-02-27
fix: no error on h-mode SV39x4 translation

NicolasDerumigny:
G-Stage translation (SV39x4/SV32x4) mandates the upper bits to be 0, not to be sign-extended.
This PR was tested working in SV39x4 mode ; we plan to add further tests for the H-Mode in a future PR.

### DIFF
--- a/core/cva6_mmu/cva6_mmu.sv
+++ b/core/cva6_mmu/cva6_mmu.sv
@@ -382,8 +382,11 @@ module cva6_mmu
     // AXI decode error), or when PTW performs walk due to ITLB miss and raises
     // an error.
     if ((enable_translation_i || enable_g_translation_i)) begin
-      // we work with SV39 or SV32, so if VM is enabled, check that all bits [CVA6Cfg.VLEN-1:CVA6Cfg.SV-1] are equal
-      if (icache_areq_i.fetch_req && !((&icache_areq_i.fetch_vaddr[CVA6Cfg.VLEN-1:CVA6Cfg.SV-1]) == 1'b1 || (|icache_areq_i.fetch_vaddr[CVA6Cfg.VLEN-1:CVA6Cfg.SV-1]) == 1'b0)) begin
+      // If we work with SV39 or SV32, so if VM is enabled:
+      // - in VS stage, check that all bits [CVA6Cfg.VLEN-1:CVA6Cfg.SV-1] are equal
+      // - in G stage (x4 mode), [CVA6Cfg.VLEN-1:CVA6Cfg.SV+1] must be zero.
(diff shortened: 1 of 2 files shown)


JeanRochCoulon:
image

NicolasDerumigny:
I see bugs on my tests following the merge of master, some access errors in VS-mode are not forwarded correctly (maybe following speculative loads?). I will look tomorrow how to fix that.

core/.vimrc
cainria, line 1: Did you forget to add this file to your `.git/info/exclude`?

NicolasDerumigny: It is not as I commit / change it on other branches. I am removing it from this PR!

JeanRochCoulon:
Concerning the bug you have seen. Unfortunately it is difficult for @LorenzoR01 or me to detect a regression without tests dedicated to your use case. Look forward to know whether you are able to fix the bug. For the futur, it would be good to discuss how to make the ci cover your use case. This would prevent from having again this kind of regression.

LorenzoR01:
Exceptions from misspredicted speculative loads are discarded on the load unit, maybe this is why some errors are not propagated.

NicolasDerumigny:
I was mistaken. The issue is not related to speculation, but comes from the fact that you test is on SV32 mode, and I test on SV39.
In SV32x4 mode (G-Stage translation only), the address is 34-bit wide, which is not taken into account on 32-bit (`CVA6Cfg.VLEN`) `fetch_vaddr` size.

I have update the condition to be partially valid in both 64 and 32-bit versions, but the 32-bit H-mode is still broken: only 32-bit GPAs are possible instead of 34 on pure G-Stage translations.
We need either:
- to change semantic of the `fetch_vaddr` in case of pure G-Stage translation to be shifted right by two bits (I do not like it)
- to enlarge `fetch_vaddr` to be 34 bits (`max(CVA6Cfg.VLEN, CVA6Cfg.GPLEN))`), but this may also generate further errors

I also fixed #3232 (parenthesis were missing in two lines of `cva6_tlb`, here also tested on Verilator.

NicolasDerumigny:
I also realized that the check is also wrong (as well as the size of `lsu_vaddr_i`) in both SV32 and SV39 for the data interface. I will look at that in the following weeks, probably in a follow-up PR, and after having submitted the tests I currently use.

JeanRochCoulon:
On next Tuesday CVA6 weekly, we will speak about adding tests to CI to test several new features. We are welcomed !

JeanRochCoulon:
Hi @NicolasDerumigny
Sorry for the delay
Can you rebase and solve conflict to allow the merge ?

JeanRochCoulon:
image

NicolasDerumigny:
I am on it, using #3246 to test correctness.

NicolasDerumigny:
#3246 tests has passed on my side. You can merge after unit tests succeed.

JeanRochCoulon:
I am generating the FPGA bitstream...

PR #3227, opened 2026-03-04
Correctly forward GPA to TLB on gvma flush

NicolasDerumigny:
GVMA flushes takes as parameter a shifted address, but the un-shifting was wrongly computed compared to the TLB's implementation. Fixing that, and minor formatting on `cvs6_tlb.sv`.

Tested to boot Linux and an hypervisor, and reproduced in Verilator.


JeanRochCoulon:
Thanks @NicolasDerumigny for the fix and the test which allows to merge calmly.

PR #3242, opened 2026-03-19
Reduce Coremark execution time to alleviate the CI

cainria:
Run one iteration before starting cycle counter, then run 1 iteration


JeanRochCoulon:
What a pity, I merged based on Zbigniew's approval before seeing your PR !    We will do better next time

PR #3247, opened 2026-03-24
issue stage: reuse signals instead of repeating

cainria:
The comb signals were assigned-to but not read.
The always_comb process was repeated in the always_ff process below.
Now the comb signals are used in the always_ff process instead of
repeating the logic.


Gchauvon:
I believe it is better to use those signals in the always_ff process below instead of removing it.
This would be more consistent with the coding style already used in the other CVA6 modules.

cainria:
Updated in accordance with @Gchauvon's advice

JeanRochCoulon:
Merci @cainria

PR #3248, opened 2026-03-25
core-v-verif: update submodule

cfuguet:
Update the core-v-verif submodule to its latest version


cainria:
```
Timeout[50s]: make spike steps=2000000 target=cv64a6_imafdc_sv39_wb variant=rv64gc_zba_zbb_zbs_zbc priv=msu elf=/home/runner/work/cva6/cva6/verif/sim/out_2026-03-25/directed_tests/beq.o tool_path=/home/runner/work/cva6/cva6/tools/spike/bin log=/home/runner/work/cva6/cva6/verif/sim/out_2026-03-25/spike_sim/rv64ui-v-beq.cv64a6_imafdc_sv39_wb.log spike_params='' &> /home/runner/work/cva6/cva6/verif/sim/out_2026-03-25/spike_sim/rv64ui-v-beq.cv64a6_imafdc_sv39_wb.log.iss
```

The 50s timeout seems really short, should we extend it?

(Btw this PR is a duplicate of https://github.com/openhwgroup/cva6/pull/3230 but i don't know why #3230 was closed.)

cfuguet:
> The 50s timeout seems really short, should we extend it?

We might... but the design has not changed. Only the verification environment (i.e. spike version). I'm not sure why this should make the tests to take longer than before.

cfuguet:
@ArnauBigas and @JeanRochCoulon, I've finally reproduced the issue of the CVA6 CI's locally. I was not able to reproduce locally the error before because the SPIKE_TANDEM environment variable was not set.

The tests are failing because the Spike (in TANDEM), runs into a C++ exception (exception of the Spike C++ code, not of the simulated system).

I will have a look of it

cfuguet:
Well, I'm analyzing the traces. Currently I have the following behavior:
- When I execute the same test on Spike alone, or on Verilator alone, the instruction traces are identical. The test finishes correctly.
- When I execute the test on Verilator with the Spike tandem, the instruction trace is different than the ones above. And the test fails.

In this test interrupts are not used. Therefore, I do not expect that the execution in TANDEM mode changes anything. Do you have an idea why TANDEM execution may differ to standalone executions ? @MikeOpenHWGroup, @JeanRochCoulon 

Thanks

MikeOpenHWGroup:
This looks like a job for either @ASintzoff or @zchamski .

JeanRochCoulon:
Verilator/testharness does not support Tandem mode et Gihub Actions CI does not enable Tandem. So I think the root issue of GitHub Action failure is not Tandem.

cfuguet:
Hi @JeanRochCoulon,

May be I'm getting something wrong, but I looked into the Github's CI, and it is normally doing Tandem. And I was able to run in tandem mode locally with the verilator harness testbench.

You can see that the TANDEM mode is being activated here:
https://github.com/openhwgroup/cva6/blob/dbb954eefc54a98505f086c96f27d106b28cdb28/.github/workflows/ci.yml#L62

As I said, tests are passing when I execute separately in Spike and in RTL. But fails when on TANDEM mode.

JeanRochCoulon:
Ok, Tandem is supported by Verilator/testharness (implemented by Mario) and enabled in GitHub Actions CI. Thanks for the clarification.

I can share these comments:
- The Github Actions CI is failed on the bump of core-v-verif (without your modification) https://github.com/openhwgroup/cva6/pull/3271.  This PR bumps 155 commits, that is huge. This means that your modification should not be the root cause of the Spike issue. Bad for core-v-verif. @MikeOpenHWGroup, this topic need to be discussed !   A first solution could be to create a cva6 dedicated to cva6... Not very good solution. Or to solve the core-v-verif bump.
- A discrepancy between Solo and Tandem modes is commonly due to Spike configuration discrepancy.

cfuguet:
Ok, thank you @JeanRochCoulon for confirming the issue.

JeanRochCoulon:
@cfuguet I would encourage you to create a dedicated cvv branch to add only your commit. When Plan-V will finish their merge between master and master_candidate, we will update cvv branch by cherry-picking your commit. I think your get the superpower to do it. Does it make sense for you ?

cfuguet:
Yes, we can do like that. Thank you @JeanRochCoulon for the suggestion. I'll do this tomorrow

JeanRochCoulon:
cva6/master could be a good branch name to remain aligned with previous terminology

JeanRochCoulon:
The Thales CI is still failed ;-(

cfuguet:
Yes, for a reason I do not yet understand, in the Thales's CI, Spike traps (illegal instruction) with the fence.i instruction, while the RTL does not. In this case the core's behaviour is the correct one, as the fence.i instruction is supported (and the ISA string includes the zifencei extension)...

So it looks like there is again a problem when passing the supported ISA extensions to Spike in the Thales's CI (Github CI's is passing).

cfuguet:
Hi @JeanRochCoulon, could you please retrigger the Thales's CI ?

cfuguet:
Thanks.

Now it is better but there is still this test failing : "Hello world with veri-testharness-pk on multiple RTL targets.". However there is no log. Could you help me to see what is the issue ?

To fix the issues related to Spike, I reverted some modifications I've made on the format used to pass RISC-V extensions from the cva6.py script to tools (including Spike). It looks like passing zifencei to have fencei does not work. Spike only integrates this instruction when passing the 'g' alias extension...

JeanRochCoulon:
Merci @valentinThomazic for the CI fix !

valentinThomazic:
@cfuguet I did not trigger the report generation when manually triggering the pipeline which is why it is empty.
All the tests are actually passing now!

cfuguet:
Thanks !

Issue #371, opened 2020-02-03
64 bit addresses must have bits 63-39 cleared


zarubaf:
If this is not the case a page fault should be generated. Right now we generate an instruction access fault.

https://github.com/pulp-platform/ariane/blob/ad70ce1f30dad539e5a365ffe71a02aaf20b397e/src/load_store_unit.sv#L339

MikeOpenHWGroup:
Hey @JeanRochCoulon, this would be a good first issue for a new member of the CVA6 team!

cathales:
Hello,

Is this issue fixed?

- https://github.com/openhwgroup/cva6/blob/master/core/load_store_unit.sv#L186C10-L186C19
- https://github.com/openhwgroup/cva6/blob/master/core/load_store_unit.sv#L610

JeanRochCoulon:
Thsi does not apply for CV32A65X. I remove the dedicated label.

chasl59:
Hi @JeanRochCoulon,
When you have time, please review the PR for a new test for this issue:
https://github.com/openhwgroup/cva6/pull/3255  
Thanks,
Charles

chasl59:
This issue has been resolved by PR [3255](https://github.com/openhwgroup/cva6/pull/3255)   
This issue may be closed.   I do not have permission to close it.

rknastenka:
@chasl59 , you can edit your PR description and write "closes #371"
since the PR got merged, the issue will close automatically if you used "fix, close" keywords

cainria:
This issue can be closed as it has been fixed

PR #3255, opened 2026-03-30
Add a new sv39 test for open issue 371: 64 bit addresses must have bi…

chasl59:
…ts 63-39 cleared

The RISC-V Privileged Architecture spec, ver 20250508, 12.4.1, requires "Instruction fetch addresses and load and store effective addresses, which are 64 bits, must have bits 63–39 all equal to bit 38, or else a page-fault exception will occur."  

This test confirms that a page fault occurs for each of these cases and confirms that the problem reported in issue 371 (i.e. an access fault instead of a page fault) has been fixed.


JeanRochCoulon:
Sorry for the delay. Thanks for your add-on.

chasl59:
Hi @JeanRochCoulon,
Thanks for processing this PR.

What is the process to close the original issue?:   
https://github.com/openhwgroup/cva6/issues/371  

I don't have permission to close that issue, and the PR has not closed it automatically.
Thanks,
Charles

PR #3265, opened 2026-04-07
Cv32a65x-act4

arshiarifat-10x:
This PR integrates the RISC-V Architectural Compatibility Test (ACT) v4.0 framework for the CV32A65x configuration of the CVA6 core, enabling architecture compliance test generation and execution within the CVA6 simulation environment.


.gitmodules
```diff
@@ -52,3 +52,6 @@
 [submodule "corev_apu/fpga/src/gpio"]
 	path = corev_apu/fpga/src/gpio
 	url = https://github.com/pulp-platform/gpio.git
+[submodule "external/act4"]
+	path = external/act4
+	url = https://github.com/arshiarifat-10x/riscv-arch-test_act4.git
```
cainria: We should probably not point our submodules to personal repositories except in rare temporary cases.

cainria (changes requested):
Hi @arshiarifat-10x 

Please add a description to this pull request so that we can understand the rationale of these changes and discuss them.

JeanRochCoulon:
Hello @arshiarifat-10x 
Thanks for your interest in cva6.
It would be useful to integrate your tests, or a part of them (depend on the duration), to the Github Actions CI. The scope is to prevent regression. Of course, to run on Github Actions, they should be executed with Verilator (possible?)

arshiarifat-10x:
@JeanRochCoulon 
I have integrated all the tests into the GitHub Actions CI using Verilator.
The current full regression takes approximately ~25 minutes to complete in CI. This is within a workable range, but if needed, we can later review the suite and discuss trimming specific tests to optimize runtime while preserving regression coverage.

JeanRochCoulon:
As specified in CONTRIBUTING.md, can you give me the right to rebase your branch.

arshiarifat-10x:
@JeanRochCoulon 
Yes, I’ve enabled maintainer access for rebasing.

.github/workflows/ci.yml
```diff
@@ -16,7 +16,12 @@ jobs:
     - uses: actions/checkout@v6
       with:
         submodules: recursive
-
+
```
cainria:
```suggestion

```

Remove trailing spaces (but keep the empty line, I'm unsure how GitHub will display my suggestion).

.github/workflows/ci.yml
```diff
@@ -16,7 +16,12 @@ jobs:
     - uses: actions/checkout@v6
       with:
         submodules: recursive
-
+    
+    - name: Fix submodule sync
+      run: |
+           git submodule sync --recursive
+           git submodule update --init --recursive
+
```
cainria:
```suggestion

```

Same

.gitmodules
```diff
@@ -52,3 +52,7 @@
 [submodule "corev_apu/fpga/src/gpio"]
 	path = corev_apu/fpga/src/gpio
 	url = https://github.com/pulp-platform/gpio.git
+[submodule "external/act4"]
+	path = external/act4
+	url = https://github.com/arshiarifat-10x/riscv-arch-test.git
```
cainria: ~~We are still pointing to your own repository, is it intended?~~

EDIT: do we want to merge this or are we waiting for the submodule patches to be merged upstream?

verif/regress/wrapper-cv32a65x-act.sh
```diff
@@ -0,0 +1,63 @@
...
+# Build the Verilator model
+echo "Building Verilator model for ${TARGET_RTL}"
+make verilate target=${TARGET_RTL} -j$(nproc)
+
+cd "${CVA6_REPO_DIR}/verif/sim"
+
+# Run generation and certification via Makefile
+# This triggers the 'gen' then 'certify' targets defined in verif/sim/Makefile
+echo "Starting ACT Regression (Generation + RTL Simulation)"
+make gen-certify target=${TARGET_RTL}
```
cainria:
```suggestion
make gen-certify target="$TARGET_RTL"
```

It shouldn't be an issue here but let's be future-proof

verif/regress/wrapper-cv32a65x-act.sh
```diff
@@ -0,0 +1,63 @@
...
+  exit 1
+fi
+
+echo "$SPIKE_INSTALL_DIR"
+
+set -e
+
+# ================== CONFIGURATION ==================
+export CVA6_REPO_DIR="$(pwd)"
+export ACT4_PKG="$(pwd)/external/act4"
```
cainria:
```suggestion
export ACT4_PKG="$CVA6_REPO_DIR/external/act4"
```

We might re-use the variable above here

verif/regress/wrapper-cv32a65x-act.sh
```diff
@@ -0,0 +1,63 @@
...
+export TARGET_RTL="cv32a65x"
+export RISCV="${RISCV}"
```
cainria: We did not define or modify `RISCV` in this file so it is not a shell variable, I guess it is already an environment variable so we should not need to export it.

```suggestion
```

verif/regress/wrapper-cv32a65x-act.sh
```diff
@@ -0,0 +1,63 @@
...
+export CV_SW_PREFIX="riscv64-unknown-elf-"
+export max_cycles=2000000
+
+echo "Environment Setup Complete"
+
+# Build the Verilator model
+echo "Building Verilator model for ${TARGET_RTL}"
+make verilate target=${TARGET_RTL} -j$(nproc)
```
cainria:
```suggestion
make verilate target="$TARGET_RTL" -j$(nproc)
```
It shouldn't be an issue here but let's be future-proof

verif/sim/Makefile
```diff
@@ -72,6 +72,81 @@ else
...
+# ==============================================================================
+
+# Updated paths for CVA6 directory structure
+certification_PROGRAM_PATH    = $(ACT4_PKG)/work/cv32a65x/elfs
+certification_PROGRAM_RELPATH = ../../vendor_lib/cva6/external/act4/work/cv32a65x/elfs
+VERI_LOG_DIR = $(CVA6_REPO_DIR)/verif/sim/simulation_results
+
+.PHONY: gen
+gen: $(ACT4_PKG)
+	@command -v uv >/dev/null 2>&1 || { echo "Error: 'uv' not found. Install from https://docs.astral.sh/uv/getting-started/installation/"; exit 1; }
```
cainria, line 87: Do we really need or want to be dependent on `uv`?
I wouldn't be against it, but I think we should make it clear.

verif/sim/Makefile
```diff
@@ -72,6 +72,81 @@ else
...
+	@echo "* Cleaning ACT4.0 directory..."
+	@$(MAKE) -C $(ACT4_PKG) clean
```
cainria:
```suggestion
	$(MAKE) -C $(ACT4_PKG) clean
```

I'm not a fan of suppressing the echo of the command, IMO having the command printed here is is already clear (with the "clean" word in it) and it is easier to debug from the commands printed.

You can keep the `@echo` if you want, but please do not use `@` otherwise.
This is the style I can find in other recipes of this Makefile.

```suggestion
	@echo "* Cleaning ACT4.0 directory..."
	$(MAKE) -C $(ACT4_PKG) clean
```

verif/sim/Makefile
```diff
@@ -72,6 +72,81 @@ else
...
+	@echo "* Generating ACT4.0 tests for CVA6 (cv32a65x)..."
+	@# Test generation now relies on the YAML config for extensions and parameters
+	@$(MAKE) -C $(ACT4_PKG) \
```
cainria:
```suggestion
	# Test generation now relies on the YAML config for extensions and parameters
	$(MAKE) -C $(ACT4_PKG) \
```

verif/sim/Makefile
```diff
@@ -72,6 +72,81 @@ else
...
+		--jobs $(shell nproc)
+	@echo "* Tests generated in: $(certification_PROGRAM_PATH)"
+
+.PHONY: certify
+certify: TEST = certification
+certify: 
+	@echo "$(BANNER)"
+	@echo "* Running ALL compliance ELFs with CVA6 Verilator Model"
+	@echo "$(BANNER)"
+	@set -e; \
```
cainria, line 103: I think this recipe should be moved to a new script file. What do you think about it?

cainria (commented):
Shouldn't we add a README to document how to use this new testing framework?

verif/sim/Makefile
arshiarifat-10x: The ACT 4.0 framework officially recommends uv as the project manager to handle Python versions and dependencies transparently.
The riscv-arch-test Makefile notes that while it is possible to run without uv, doing so often leads to "dependency issues" that can cause tests to fail or generate incorrectly.

verif/sim/Makefile
arshiarifat-10x: Keeping the logic in the Makefile allows developers to see the entire execution flow( from address discovery to signature validation) in a single place without needing to track down external helper scripts.
As we integrate this into GitHub Actions, having a self-contained Makefile target ensures that the CI runner only needs to execute make certify. This avoids the overhead of managing absolute/relative pathing issues across different environments.

.gitmodules
arshiarifat-10x: The changes have now been merged upstream, and I’ve updated the submodule to point to the official repository instead of my fork.

verif/sim/Makefile
arshiarifat-10x: But if you suggest i can move this recipe to new script file. However, I am open to either approach. Would you prefer:
1- A clean separate script (e.g., verif/sim/run_certification.sh) with the Makefile target acting as a simple wrapper?
2- Keeping it in the Makefile to match the existing simulation target patterns in this repository?
Please let me know which direction aligns better with the project's long-term coding standards, and I will 
update the PR accordingly.

cainria: We have a lot of scripts in `verif/regress/`, you might be interested in looking at these. Maybe `verif/regress/` is a suitable place to add `certify.sh`?

For CVA6 designers, these scripts are our entry points to run most simulations: they call the relevant tools (`cva6.py` and `make`) for us.
So I think we do not need a `Makefile` wrapper as `make` it is not the user-facing entry to run simulations.

arshiarifat-10x: The current implementation is already aligned with the verif/regress/ flow, the wrapper is placed there intentionally and acts as a standard entry point, similar to existing scripts. It handles environment setup and then invokes the existing Makefile targets (gen and certify) under the hood.

From a user perspective, this keeps the interface consistent with other regression flows:

**bash verif/regress/wrapper-cv32a65x-act.sh**

So users are not expected to interact with the Makefile directly.

I’ll add a short README section to make the flow and entry point clearer.

ASintzoff:
Currently, test suites are cloned on demand when the corresponding tests are executed. For instance, there is already the script `verif/regress/install-riscv-arch-test.sh` cloning `riscv-arch-test` into `verif/tests/riscv-arch-test`.

Could you consider to reuse this script?

MikeOpenHWGroup:
The ACT4 certification test suite does not work that way @ASintzoff.  ACT4 is not a library of tests, it is a test-generator.  The process of generating these tests is rather long and complex so for other CORE-V cores we are employing the following strategy:
1. Use the ACT4 tools to generate tests.  This would be done manually for on-going development and/or integration of ACT4/UDB/Sail updates.  It is anticipated that this would not happen often.
2. Generate the test once and then commit them to a regression directory.  These would be run often.

MikeOpenHWGroup:
Hi @ASintzoff, we should discuss how to move forward on this.

JeanRochCoulon:
Hello @arshiarifat-10x   Why core-v-verif is updated ?  This PR should not modify the core-v-verif hash

JeanRochCoulon:
@AlexChenIC The Verible repo cannot be fetched. Do you have an idea howto fix it ?

JeanRochCoulon:
Hello @arshiarifat-10x    we are ready to merge.   But the CI is red due to an installation issue.

verif/sim/Makefile
cainria: Hi, FYI GitHub Actions fail because `uv` is not installed in the GitHub runners.

cainria:
I add it here for visibility: as mentioned in https://github.com/openhwgroup/cva6/pull/3265#discussion_r3550333942 this PR adds a requirement to `uv` which is not fulfilled by the GitHub runner, hence the job failures.

JeanRochCoulon:
@AlexChenIC What about adding the job in your Github action CI ?   This will reduce engineering work to directly push it into the final targeted CI. It would be more efficient. Do you share this view ?

AlexChenIC:
> @AlexChenIC What about adding the job in your Github action CI ? This will reduce engineering work to directly push it into the final targeted CI. It would be more efficient. Do you share this view ?

Hi @JeanRochCoulon, yes, I share this view. I think that ACT4 would be a good fit for the Tier CI as a dedicated Tier 2 job for CV32A65X. I should be able to take on this integration. I’ll start by evaluating the workflow. Based on my initial assessment, the integration looks feasible

cainria:
Hi! Any news about this PR ?

arshiarifat-10x:
@cainria README was pending i added that. I hope the CI issue is resolved.

arshiarifat-10x:
Since @AlexChenIC  mentioned taking on the Tier-2 ACT4 CI integration, should I wait for that, or should I add the uv installation support on my side? @JeanRochCoulon

PR #3266, opened 2026-04-07
Refresh CONTRIBUTING.md

jquevremont:
Be more assertive on parameters (and no `directives)
Transition of OpenHW Foundation.
Same email address in GitHub and Eclipse for ECA checks.


CONTRIBUTING.md
```diff
@@ -20,26 +20,25 @@ Therefore here are guidelines to help the CVA6 team accept new contributions:
...
 - Specific recommendations:
     * For instruction set extensions, talk to the team to assess the relevance of including it into the core or as a coprocessor on the CV-X-IF interface.
         -  If the extension is custom (not a RISC-V specified extension), a coprocessor on CV-X-IF is definitely its place.
     * Your contribution shall be optional and fully disabled by default.
         - so that projects already using CVA6 are not impacted (no functionality change, no extra silicon...).
-    * To configure your contribution, System Verilog top-level parameters are preferred.
-        - Synchronize with CVA6 team if this is not possible and you need other means (`directives, ariane_pkg parameters, templating...)
-        - Synchronize with CVA6 team (again) because this rule is likely to change over time as we are considering templating.
+    * To configure your contribution, System Verilog top-level parameters are strongly advised.
+        - `directives are not permitted.
```
MikeOpenHWGroup: Can we say, "_Conditional compiler directives such as `ifdef, etc. are strongly discouraged._"?

The term "conditional compiler directives" is used in the SystemVerilog LRM (IEEE-1800). I will approve "not permitted" if you _really_ want it, but there is always an exception to every rule, so I suggest "strong discouraged".

MikeOpenHWGroup (changes requested):
This is great @jquevremont, thanks for this PR.  I have suggested one minor edit.

CONTRIBUTING.md
jquevremont: Thanks for the feedback. I have copied your sentence as is.

PR #3254, opened 2026-03-30
MMU: Add Svpbmt scaffolding and wire pbmt through PTW

StackedArchitect:
## Summary

Adds configuration scaffolding for Svpbmt (page-based memory types), mandatory in RVA23 and currently absent from CVA6. All changes are inert when `SvpbmtEn=0` (the default for every existing config).
Part of #3253

## Changes

| File | Change |
|------|--------|
| `include/config_pkg.sv` | Add `bit SvpbmtEn` after `SvnapotEn` in both `cva6_user_cfg_t` and `cva6_cfg_t` |
| `include/build_config_pkg.sv` | Propagate `SvpbmtEn` from user config into built config (mirrors `SvnapotEn` pattern) |
| `core/cva6_mmu/cva6_mmu.sv` | Split `pte_cva6_t.reserved[8:0]` → `pbmt[1:0]` + `reserved[6:0]`; add `logic [1:0] pbmt` to `tlb_update_cva6_t` |
| `core/cva6_mmu/cva6_ptw.sv` | Wire `pte.pbmt` into `shared_tlb_update_o.pbmt` (gated on `SvpbmtEn`); clean up invalid-PTE comment |

## Why this structure

1. `SvpbmtEn` follows the `SvnapotEn` pattern exactly. Optional, parameter-gated, disabled by default (parachute rule satisfied). All future functional code will be gated on `CVA6Cfg.SvpbmtEn`.
2. `pte_cva6_t.reserved[8:0]` currently buries PTE bits [62:61] (the MT field) inside the reserved check at `cva6_ptw.sv:430`, causing a page fault on any Svpbmt-typed page. After this split, `pte.reserved` is 7 bits so `|pte.reserved` naturally excludes the MT field, no logic change required.
3. `tlb_update_cva6_t` is a `localparam type` defined only at `cva6_mmu.sv` and passed as `parameter type` to all three submodules, one field addition propagates everywhere automatically, the same mechanism used for `is_napot_64k`.
4. The PTW `tlb_update` block now assigns `shared_tlb_update_o.pbmt = CVA6Cfg.SvpbmtEn ? pte.pbmt : 2'b00`, following the same pattern as `is_napot_64k`. When `SvpbmtEn=0`, `pbmt` is tied to zero with no logic overhead.

## Testing
With `SvpbmtEn=0` (all existing configs): no functional change. CI should pass unmodified. Subsequent commits will wire `pbmt` through TLB storage, LSU port, and cache backends.

[...]

PR #3000, opened 2025-06-06
Making clear that AI-generated PRs are not allowed

jquevremont:
This follows PRs from @ibooga generated by Claude Code that spam the reviewers. 
We need to find a technical solution to instruct language modes or AI coding assistants not to submit PRs.
In the meantime, the rule is written in CONTRIBUTING.md.

PR #3269, opened 2026-04-10
Implemented Svpbmt PTW validation and CSR control bits

StackedArchitect:
## Implement Svpbmt PTW validation and menvcfg/henvcfg.PBMTE

This PR builds on the scaffolding from #3254 (which split the PTE 'reserved' field into a 2-bit 'pbmt' and a 7-bit 'reserved') and makes the Svpbmt extension actually functional.

The core problem it solves: prior to this, _pte.pbmt_ was being extracted from the PTE but never validated and never gated by any CSR. That means a guest could set PBMT bits on a page and the hardware would silently accept them even if the supervisor hadn't enabled the extension, which contradicts section 12.3 of the RISC-V Privileged Spec.

The fix has two parts. First, I added _menvcfg.PBMTE_ and _henvcfg.PBMTE_ register bits to _csr_regfile.sv_, with _henvcfg.PBMTE_ gated on _menvcfg.PBMTE_ at write time (a guest can't enable PBMT if M-mode hasn't allowed it). The _pbmte_ signal is then wired down through cva6.sv to ex_stage to load_store_unit to cva6_mmu to cva6_ptw, following the same path as existing CSR-to-MMU signals.

Second, in the PTW I added validation on leaf PTEs: encoding '2'b11' is explicitly reserved by the spec and always raises a page fault, and any non-zero PBMT value raises a fault if 'pbmte_i' is low (i.e., the extension is not enabled). Non-leaf PTEs must always have 'pbmt == 0'; the spec does not allow memory type hints on intermediate entries. On a clean walk, 'pbmt' is forwarded into the TLB update struct so it's available for future use at the access stage.

The 'SvpbmtEn' config bit gates all of this at synthesis time, so designs that don't need the extension pay no cost.


cainria:
:warning: **Please only use ASCII characters in the code (including comments)**

AI;DR

StackedArchitect:
> ⚠️ **Please only use ASCII characters in the code (including comments)**
> 
> AI;DR

Thanks for catching that! I had two em dash characters (U+2014 ) that slipped into comments in cva6_mmu.sv line 111-112 inside the pte_cva6_t struct. I've replaced them with ASCII double hyphens (--) and force-pushed the fix to both branches. I've also audited all 9 changed files and confirmed there are no remaining non-ASCII characters in any of my additions.

StackedArchitect:
Hi @cainria ,
I saw PR #3000 and wanted to address this directly since my PR is open.
I did use AI assistance during development, mainly to navigate the CVA6 codebase faster and understand where signals were being passed. But the implementation decisions were mine, splitting the reserved field, gating henvcfg.PBMTE on menvcfg.PBMTE, and the PTW validation logic for rejecting 2'b11 encoding and non-zero PBMT when !pbmte_i, those came from reading the Svpbmt spec (section 12.3) myself.
I also clearly made mistakes AI wouldn't catch, the em dashes in comments being the obvious one.
I'm happy to walk through any part of the code and explain my reasoning. I understand if the PR description doesn't meet the new expectations and can rewrite it. Thank You.

JeanRochCoulon:
Thanks @StackedArchitect.
In the mean time you get the green light to submit the contribution, I will close this PR. It will clarify the PR list to be reviewed.
Of course, you are welcome to come back when it will be the good moment.

PR #3270, opened 2026-04-12
[verif/sva] Add SVA assertions for CVA6 ALU module

angadsingh2:
## Summary

Add SystemVerilog assertions to verify key functional properties
of the CVA6 ALU module that were not previously covered.

## Assertions Added

- `zero_flag_correct`: Verifies zero flag is HIGH only when result is zero
- `less_flag_binary`: Verifies less comparison flag is always binary
- `branch_res_binary`: Verifies branch comparison result is always binary  
- `sltu_result_one_bit`: Verifies SLTS/SLTU result is constrained to 0 or 1

## Testing

- Lint check passed using Verilator
- Assertions verified against CVA6 ALU RTL signals

## References

- ALU RTL: core/alu.sv
- New file: verif/tb/core/sva/cva6_alu_sva.sv

### DIFF
--- a/verif/tb/core/sva/cva6_alu_sva.sv
+++ b/verif/tb/core/sva/cva6_alu_sva.sv
@@ -1,0 +1,49 @@
+// Copyright 2024 OpenHW Group
+// Solderpad Hardware License, Version 2.1
+// SPDX-License-Identifier: Apache-2.0 WITH SHL-2.1
+//
+// SVA assertions for CVA6 ALU module
+// Contributor: Angad Singh <s.angad2.2004@gmail.com>
+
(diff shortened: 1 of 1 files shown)


cainria:
Signing the ECA is required to merge PRs.

Also a few questions:

- Do we need these assertions compared to what is already performed during verification? (Aka is this PR needed?)
- It seems that the created file is unused, is it intended?
- You comment the rules as "Rule 1" etc. Does the order matter?
- I don't understand, what do you mean in the Reference section of the PR description?

jquevremont:
ECA not signed. PR closed.

Issue #3286, opened 2026-04-28
[TASK] Rewriting flush logic for cva6_shared_tlb.sv


ahmadtkhan:
### Is there an existing CVA6 task for this?

- [x] I have searched the existing task issues
- [x] I am a human engaging in an interpersonal interaction. During this interaction, my words are my own and are not generated. If relevant, I provide links to my sources.

### Task Description

The L1 TLB or cva6_tlb performs an entry-by-entry matching using VA, ASID, and VMID while, the L2 TLB or cva6_shared_tlb clears the [entire shared entries using any of the flush strobes](https://github.com/openhwgroup/cva6/blob/master/core/cva6_mmu/cva6_shared_tlb.sv#L420)

Although this implementation is simple, functionally correct and efficient in terms of timing and area, it results in a high loss of unrelated address spaces and pages, creating a design which will have more misses. 

If the shared TLB is redesigned like a fully associative cache, the timing and area will be impacted significantly. Hence I am proposing and working on a multi-stage TLB flush walker that iterates through cache sets incrementally rather than a single cycle improving shared-TLB hit rate and reducing PTW activity . When a flush is requested in the idle state, the operands and the flush type is flushed. Another state reads one set from the PTE arrays and the final state evaluates each way using the same conditions as L1 TLB, clearing only the matching entries. 

Let me know if this looks like a valid concern and any other comments about this issue. 

### Required Changes

I think changes will only be needed in cva6_shared_tlb module. Most likely a one-file patch improving flushing logic

### Current Status

WIP

### Risks

Impact to timing and area. Accurate functionality needs to be verified after implementation

### Description of Done

Functionally accurate design with improvements in shared-TLB hit rate and drop in PTW activity

PR #3312, opened 2026-05-09
Improved TLB flushing logic for Supervisor mode

ahmadtkhan:
Improved TLB flushing logic for Supervisor mode decreasing the number of Shared TLB accesses, misses, and PTWs. 

This PR addresses #3286 by changing shared TLB flushing for S-stage translation from coarse invalidation to selective invalidation as S-stage flush invalidated the entire shared TLB in the original design. This patch adds selective flush support for the shared TLB using the flushed address and ASID, using similar flush logic of ITLB/DTLB. The implementation uses a small multi-cycle FSM to walk the shared-TLB sets and clear only matching valid entries instead of dropping the whole structure. This reduces over-invalidation in the shared TLB and preserves unrelated translations.

This PR is limited to S-stage flushing. VS-stage and G-stage cases should be added in a few days after they are validated and tested.


JeanRochCoulon:
To fix Verible, you can run the command which is at the bottom of CONTRIBUTING.md file

JeanRochCoulon:
Hello, can I ask you to fix Verible ?  (just execute the Veribla command on your branch)

ahmadtkhan:
Hey, sorry for the delay but I completed the PR by adding logic for flush_vvma_i and  flush_gvma_i and ran Verible as well.

fatimasaleem:
@JeanRochCoulon can we merge this?

sjo99-kr:
Hi, @fatimasaleem , @JeanRochCoulon , @ahmadtkhan 

I do not mean to be offensive or to bother you, but I am also looking into the shared TLB module while working on the Svadu, Svpbmt, and Svinval extensions. While reading this implementation, I was wondering whether the new multi-cycle flush behavior could introduce two corner-case issues.

I think there are two separate corner cases:

1) A second or later flush pulse may be missed if it arrives while the shared TLB multi-cycle flush FSM is still busy.
2) The flush operands (vaddr, asid, gpaddr, vmid) may not be captured correctly in ex_stage for the accepted flush request.

---

In the current CVA6 flow, `sfence.vma` (including `hfence.vvma`, `hfence.gvma`) appears to assume that the TLB flush request is effectively handled as a one-cycle pulse. If there are no pending stores, the `sfence.vma` instruction can commit, and the flush operands such as `vaddr_to_be_flushed_i`, `asid_to_be_flushed_i`, `vmid_to_be_flushed_i`, and `gpaddr_to_be_flushed_i` are provided for that flush request.

With the new shared TLB implementation, however, the flush operation can take **multiple cycles**. In that case, if two `sfence.vma` or related invalidation instructions are committed close (or sequentially) to each other, the shared TLB may still be processing the first flush when the second flush request is asserted.  
(The commit condition of `sfence.vma` is just no pending stores.)
If the shared TLB only accepts a new flush request in the idle state, **the second flush pulse may be missed or dropped**.

Another concern is that **the flush operand signals from `ex_stage` are not guaranteed to stay stable for the whole multi-cycle flush**.  After the first fence commits and `flush_i` clears the fence state, these registers may be overwritten by later rs1/rs2 values or by a subsequent fence. Therefore, the shared TLB should latch the flush operands when it accepts the request, or `ex_stage` should hold them until the flush is acknowledged.

So I was wondering whether we need some kind of `flush_ready / flush_busy` handshake back to the `controller` or `commit_stage`, so that subsequent `sfence.vma`, `hfence.*`, `sinval.vma`, or `hinval.*` instructions are not committed while the shared TLB flush FSM is still busy. Another possible solution might be to add a pending-flush mechanism inside the shared TLB.

I understand that this is a microarchitectural issue rather than a direct architectural requirement from the RISC-V specification. However, if the shared TLB flush changes from a single-cycle operation to a multi-cycle FSM, I think some handshake or pending mechanism may be needed to avoid losing back-to-back flush requests.

How you guys think about it?

Thanks :)

ahmadtkhan:
Hey @sjo99-kr , 
Thank you for looking at the code and providing feedback (it is not offensive at all). I kept the implementation minimal to ensure proper functionality and was hesitant to modify files outside the MMU, but your first concern looks valid. Even though I am unsure about the frequency of successive invalidation instructions, since my understanding was that fence instructions are usually isolated, it still makes sense to have a flush_busy signal to ensure flushes are not missed in any case.
I can get started on an implementation in which flush_busy_o is generated in cva6_shared_tlb and propagated upstream outside the MMU, so that fence retirement waits until the flush request is accepted and the operands are naturally held stable until acceptance. I think this would require some logic in cva6_shared_tlb and commit_stage, along with wiring in cva6_mmu, load_store_unit, ex_stage and cva6.
I am not sure if I fully understand your second concern. Once the flush is accepted in IDLE, all the flush_*_q operands are already latched, so the active flush no longer depends on ex_stage. If your point is that the operands for a subsequent flush can be overwritten before they are accepted, then I think that is not really separate from issue 1 and should be resolved by adding a flush_busy signal.
Please let me know if you have any thoughts or concerns about this plan.

sjo99-kr:
Hi @ahmadtkhan 

Yes, I agree that both concerns are rooted in the same request/accept protocol issue, so they can likely be solved by more robust commit-side handling for multi-cycle TLB flushes.

To clarify my second point, I did not mean that the active flush would use wrong operands after it has already been accepted by the shared TLB. As you said, once the request is accepted in IDLE, the `flush_*_q` operands are latched inside the shared TLB, so the active flush should be safe.

My concern was more about a subsequent fence/invalidation instruction. The flush operand signals come from `ex_stage`, and after a fence commits and `flush_i` clears the fence state, those operand registers may be updated with later `rs1_forwarding` / `rs2_forwarding` values. Therefore, if another fence reaches commit while the shared TLB is still busy and cannot accept the new request, not only the second flush pulse which comes from `controller.sv`  (i.e., first concern) but also the corresponding operands which comes from `ex_stage.sv` for that request (i.e., second concern) may never be captured by the shared TLB.

So I agree that both concerns can be addressed by adding a `flush_ready` / `flush_busy` handshake, or alternatively a pending-flush mechanism, so that fence/invalidation instructions only retire when their flush request and operands can be accepted.

Thanks :)

JeanRochCoulon:
Thank you @ahmadtkhan @sjo99-kr for this discussion. 
@sjo99-kr do you approve the PR ?

core/cva6_mmu/cva6_shared_tlb.sv
```diff
@@ -402,9 +468,12 @@ module cva6_shared_tlb #(
   // ------------------
   // Update and Flush
   // ------------------
-
```
JeanRochCoulon, line 405: Can you keep this lign ?

core/cva6_mmu/cva6_mmu.sv
```diff
@@ -100,7 +103,6 @@ module cva6_mmu
     output dcache_req_i_t req_port_o,
 
     // PMP
-
```
JeanRochCoulon, line 103: Can you keep this line?

core/cva6_mmu/cva6_mmu.sv
```diff
@@ -170,6 +172,7 @@ module cva6_mmu
 
   logic shared_tlb_access, shared_tlb_miss;
   logic shared_tlb_hit, itlb_req;
+  //logic shared_tlb_lookup_done;
```
JeanRochCoulon: Can you remove this line

core/cva6_mmu/cva6_shared_tlb.sv
```diff
@@ -18,8 +18,7 @@
 //              shared TLB sv32 developed by Sebastien Jacq (Thales Research & Technology)
 //              to be used with sv32, sv39 and sv39x4.
 
-/* verilator lint_off WIDTH */
-
```
JeanRochCoulon, line 22: Can you cancel this modification ?

core/cva6_mmu/cva6_shared_tlb.sv
```diff
@@ -91,10 +97,38 @@ module cva6_shared_tlb #(
     logic [HYP_EXT*2:0] v_st_enbl;  // v_i,g-stage enabled, s-stage enabled
     logic is_napot_64k;  // Svnapot: Flag indicating a 64KiB NAPOT page
   } shared_tag_t;
-
```
JeanRochCoulon, line 94: Same

core/cva6_mmu/cva6_shared_tlb.sv
```diff
@@ -551,5 +833,3 @@ module cva6_shared_tlb #(
     end
   end
 endmodule
-
```
JeanRochCoulon, line 554: Can you keep it?

core/cva6_mmu/cva6_shared_tlb.sv
```diff
@@ -417,14 +489,226 @@ module cva6_shared_tlb #(
...
+          flush_kind_d = FLUSH_VVMA;
+          flush_vmid_d = lu_vmid_i;
+          flush_gpaddr_d = '0;
+        end else if (flush_gvma_i) begin
+          flush_state_d = FLUSH_READ;
+          flush_idx_d = '0;
+          flush_asid_d = asid_to_be_flushed_i;
+          flush_vaddr_d = vaddr_to_be_flushed_i;
+          flush_kind_d = FLUSH_GVMA;
+          flush_vmid_d = '0;
```
sjo99-kr: `flush_vmid_d = vmid_to_be_flushed_i; `
`flush_gpaddr_d = gpaddr_to_be_flushed_i; `

I guess above code is correct for flush_gvma insturction, how do you think?

sjo99-kr:
Hi, @JeanRochCoulon 
Without above concerns for `gvma` case, it seems architecturally clear. 
And I ran my code based on this PR for "Svinval" ACT with "sfence.vma", it passed :)

core/cva6_mmu/cva6_shared_tlb.sv
ahmadtkhan: you are right G-stage needs VMID and gpaddr. Thank you for pointing that out. fixed it

JeanRochCoulon:
Great discussion, I merge

PR #3318, opened 2026-05-11
Add link to wiki

MikeOpenHWGroup:
I took the liberty of adding a Wiki to this repo to discuss the evils of ifdefs.  This PR adds a link to it.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,7 +11,7 @@ Drop a comment in your pull request so we know about your problem and we do not 
 
 Contributors are encouraged, but not required, to be a [member](https://openhwfoundation.org/membership/become-a-member/) of the OpenHW Foundation, part of the Eclipse Foundation.
 
-## Read this before
+## Read this first
 
 ### Major evolutions
 
@@ -36,6 +36,7 @@ Therefore here are guidelines to help the CVA6 team accept new contributions:
         - so that projects already using CVA6 are not impacted (no functionality change, no extra silicon...).
     * To configure your contribution, System Verilog top-level parameters are strongly advised.
         - Conditional compiler directives such as <code>`ifdef</code>, etc. are strongly discouraged.
+        - Please review our wiki on the topic of [Configuring SystemVerilog RTL Models](https://github.com/openhwgroup/cva6/wiki/Configuring-SystemVerilog-RTL-Models), in particular the [CVA6 use case](https://github.com/openhwgroup/cva6/wiki/Configuring-SystemVerilog-RTL-Models#use-case-the-openhw-foundation-cva6).
         - Synchronize with CVA6 team if you do not find an appropriate solution.
     * Commit to maintain your contribution 2 years after the pull request
         - We know it's not always possible, so refer to the next rule.


cainria:
Really interesting reading! I knew before reading that I would agree with the conclusion, but I have learnt things from the rationale (as usual, it is inspiring for the language I'm building, where hardware configuration will be one of the key points).

I would like to tell a little bit more on what we did in the CVA6, more specifically *how* it differs and *why* we did it this way.

Our “parametrization” was meant for [Heterogeneous Variants](https://github.com/openhwgroup/cva6/wiki/Configuring-SystemVerilog-RTL-Models#heterogeneous-variants). A package is global too, so we cannot build several versions of the same package at once (in the same build unit) and none of the solutions there could be used for our use-case.
- Solution “1. Expand chip_config_pkg to handle multiple variations” does not seem scalable (how many variants of CVA6 will be in the same platform?)
- Solution “2. Create multiple top-level modules and import the specific chip_config_pkg you want.” requires to create wrappers (the multiple top-modules), then to provide all the parameters read from the packages. (So do we need a package in the end?)

**We decided to use a *structure* instead of a package** so that the user can provide a structure to each instance of CVA6 as a generic parameter.

However, in a structure, contrary to a package, we cannot add fields computed from the values of other fields, and we cannot add types (btw in a package we have to repeat this logic in all packages, which is not *DRY*).
So we added a `build_config` function to return another structure with all the values, user-given and computed.
For the types, they are built at the top of CVA6 and transmitted through the hierarchy.

Then we face the issue with types at the interface of CVA6 (RVFI for instance). Here we had to use macros to define structures from a configuration because SystemVerilog lacks generic types, but shush don't tell it to others :grin:

cainria (approved):

MikeOpenHWGroup:
Thanks for the additional content @cainria.  I updated the wiki to incorporate some of your comments.

@jquevremont and @JeanRochCoulon, what are your thoughts?

cainria:
:+1: for the changes.

<details><summary>Potentially off-topic</summary>

Also, I may be off-topic here, but I see you added a section about macro expansion. The example seems not [hygienic](https://en.wikipedia.org/wiki/Hygienic_macro).

```systemverilog
if (`IS_SLLI) process_shift_left_logical_immediate();
// What should "be slli" here?
// Does it implicitly capture the signal storing the current instruction?

// The following seems more explicit:
if (`IS_SLLI(instruction)) process_shift_left_logical_immediate();
```

A function or a tiny module instead of a macro would do the job here, ~~and would benefit from typing~~ ah, no, sorry, it is SystemVerilog, providing an argument with the wrong type is not an error… :grin: 

To go back to the use of macros in the CVA6 code, the difference with this example is that we use *macros* for *types* (because there are no alternatives) but *functions* for *values*.

</details>

JeanRochCoulon:
Very interesting !

jquevremont:
Excellent! Thanks :-)

jquevremont:
@mike, have you searched how to add backticks in code blocks in Markdown?
Here is the hint:  ``<code>`ifdef</code>``. There is one example in CONTRIBUTING.md.

PR #3324, opened 2026-05-18
fix: tie `debug_mode` to 0 if `DebugEn` is false

ricted98:

`DebugEn` set to 0 results in `debug_mode` being at `x`. This is problematic since it is widely used throughout the design and can generate subtle bugs. For example, I noted it since the performance counters did not increase as they rely on `debug_mode` being 0, while it was instead at `x`.

### DIFF
--- a/core/csr_regfile.sv
+++ b/core/csr_regfile.sv
@@ -2744,7 +2744,7 @@ module csr_regfile
 
   // determine if mprv needs to be considered if in debug mode
   assign mprv = (CVA6Cfg.DebugEn && debug_mode_q && !dcsr_q.mprven) ? 1'b0 : mstatus_q.mprv;
-  assign debug_mode_o = debug_mode_q;
+  assign debug_mode_o = CVA6Cfg.DebugEn ? debug_mode_q : 1'b0;
   assign single_step_o = CVA6Cfg.DebugEn ? dcsr_q.step : 1'b0;
   assign mcountinhibit_o = {{29 - MHPMCounterNum{1'b0}}, mcountinhibit_q};
 


cainria (approved):

cainria:
I'm wondering, is it better to do this change and/or removing the conditional in the register initialization? (Synthesis tools usualy perform constant register elimination.)

https://github.com/ricted98/cva6/blob/d4d3a793a94f9bd94fcec6e5bf0e146cfed12b8b/core/csr_regfile.sv#L2763

ricted98:
it would be fine by me. I chose the current implementation simply to match what was done for ```single_step_o```, but I can do it as you suggest.

cainria:
Yes, I approve your current change, yet I think the other option is worth discussing.

Anybody else have thoughts about it?

MikeOpenHWGroup:
Hi @cainria, I am not clear on what the "other option" is.
This PR matches [refactoring example 2](https://github.com/openhwgroup/cva6/wiki/Configuring-SystemVerilog-RTL-Models#refactoring-example-2-fine-grain-configuration) in the Configuring SystemVerilog RTL Models wik.  If there is another way to do this that is better for synthesis, I would like to add it to the wiki.

cainria:
My concern with the current approach is that we *could* leave `x` values propagate from the un-driven signal elsewhere (but we probably don't so it is okay!)

The other option would be to make the register constant = 0 **once** instead of adding a ternary operator **each time it is read**.
I don't know if a solution is better than another for synthesis.

<details><summary>Off-topic on hardware description languages in general</summary>

IMO the only "correct" approach would be to use a language where we can declare that the register *does not exist* in some configurations, so the compiler errors out if we try reading from it while it does not exist.

</details>

Anyway, this PR applies the current policy to one more place so **I am for merging this PR as it is**.

It does not prevent us from discussing the policy and updating the code later.

JeanRochCoulon:
Thanks for the fix !

PR #3285, opened 2026-04-27
ci: add Phase 1 tier workflows and tier dashboard

AlexChenIC:
## Summary

This PR proposes the first Phase-1 GitHub CI rollout for CVA6, based on prior discussions within OpenHW.

The scope of this first rollout is intentionally limited to Verilator-based GitHub CI for the first three selected green configurations:

- `cv32a65x`
- `cv32a60x`
- `cv64a6_imafdc_sv39_hpdcache_wb`

In addition to the workflow definitions, this PR also adds a lightweight tier dashboard flow to collect CI run data and render a static HTML summary page for Tier 1 / Tier 2 status.

For reviewers who want to see the intended dashboard output quickly, a previously generated demo is available here:

- https://alexchenic.github.io/cva6/tier/

## Background

This change follows the earlier internal review of the Phase-1 Tier CI structure.

The intent of that proposal was to introduce two initial GitHub CI tiers for the selected Phase-1 green configurations:

- Tier 1: a fast sanity regression intended for automatic checks of pull-requests
- Tier 2: a broader Verilator regression intended for daily or manual validation

This PR is the first upstreamable implementation of that Phase-1 structure.

### Phase-1 CI workflows

- Add `openhw-cva6-ci-tier1.yml`
- Add `openhw-cva6-ci-tier2.yml`
- Add a shared composite action: `.github/actions/setup-cva6-env/action.yml`

### Tier dashboard

- Add `openhw-cva6-tier-dashboard.yml`
- Add dashboard data collection / parsing / HTML generation scripts under:
  - `.github/scripts/dashboard_tiers/`

### cv32a60x testlist alignment

Adjust the `cv32a60x` testlists so they match the actual configuration capabilities:

- `testlist_riscv-arch-test-cv32a60x.yaml`
- `testlist_riscv-compliance-cv32a60x.yaml`
- `testlist_riscv-tests-cv32a60x-p.yaml`

These updates avoid running tests that do not match `cv32a60x` capabilities, including:
- unsupported atomic/A-extension coverage
- supervisor-mode tests on an M-mode-only configuration
- tests that rely on `TvalEn=1` when `cv32a60x` has `TvalEn=0`

### Tier 1

Tier 1 is intentionally lightweight and is meant to provide automatic PR checks for the approved Phase-1 configurations.

For this first version, it runs one representative smoke or directed testcase per selected configuration so that feedback stays fast and practical for pull-request use.

### Tier 2

[...]

Issue #3174, opened 2026-01-04
[BUG] `Store` instruction triggers `Load Access Fault (mcause=5)` instead of `Store Fault` upon `PMP` write violation


hjw-arch:
### Is there an existing CVA6 bug for this?

- [x] I have searched the existing bug issues

### Description
During fuzzing, we observed that when store instructions (e.g., `sw`, `fsw`) access a PMP write-protected region, CVA6 incorrectly raises a **Load Access Fault (mcause=5)**. According to the RISC-V Spec, this should be a **Store/AMO Access Fault (mcause=7)**. This behavior is consistent across both integer and floating-point store operations.

### Reproduction Details

- Typical Instruction: sw a2, -580(t6) (PC: 0x80003054)
- Context: Target address is PMP Write-Denied.
- Observation: CVA6 traps the access but reports mcause = 0x5. Spike reports mcause = 0x7.

### Environment

- Commit ID: aa4f5a54
- Config: cv64a6_imafdc_sv39
- ISA: rv64g_zba_zbb_zbs_zbc_zbkb_zbkx_zkne_zknd_zknh

### Potential Root Cause
We suspect a potential issue in the interaction logic between the **LSU (Load-Store Unit)** and the **PMP module**. The behavior suggests that when the PMP trap is triggered, the "Store" attribute of the operation might not be correctly synchronized or propagated, causing the logic to default to a "Load" exception type. This is likely a timing or signaling mismatch.

#### Spike
```
core   0: 0x0000000080003048 (0xdacfae23) sw      a2, -580(t6)
core   0: exception trap_store_access_fault, epc 0x0000000080003048
core   0:           tval 0x0000000080100dbc
core   0: >>>>  other_exp
core   0: 0x0000000080001000 (0x34202f73) csrrs   t5, mcause, zero
core   0: 3 0x0000000080001000 (0x34202f73) x30 0x0000000000000007
```

#### CVA6
```
LD_ACCESS_FAULT exception @ 0x0000000080003048 (0xdacfae23)
core   0: 0x0000000080001000 (0x34202f73) csrrs   t5, mcause, zero
3 0x0000000080001000 (0x34202f73) x30 0x0000000000000005
```

### Related
A similar issue was reported in 2022 as #906 and addressed by #908. On the latest CVA6, we observe a similar symptom. This may indicate a different root cause or a regression. Could you please verify?

### Attachments

ELF binary and full execution logs are attached.

[...]

PR #2953, opened 2025-05-05
Proposed CI fix: remove edatools repository from prerequisites

ricted98:
The CI is currently broken due to an expired GPG key for the `edatools` repository. However, it seems like there is no real need for it in the current CI.

Removing it fixes the CI. I could not observe anything breaking due to this change. However, someone more knowledgeable about the CI could double check the proposed change.

PR #3331, opened 2026-05-27
Master candidate pr

AngelaGonzalezMarino:
First step to merge master candidate branch to master:

- Development of missing features in YPB interface

- Smoke tests pass in all configurations (with both AXI and OBI as noc bus)

- Hello World in FPGA is OK (both 32 and 64 bits)

- Linux boot is successful in 32 bits and in 64 bits (genesys 2 and agilex 7)

- Added verification scripts and configurations used

- Update UVM Infrastructure to support different configurations

- Update APU to master (both genesys 2 and agilex 7)

- Merge E-tracer (only for genesys 2)


JeanRochCoulon:
Hello Angela,  in CI, I get this kind of errors, some tests are not found :
image

JeanRochCoulon:
Hello,   Hello word does not work because cvxif is setup two times for Spike invocation in the infrastructure (cva6.py may set twice the cvxif in ISA):
image
If needed, you can contact @zchamski for deep debugging

JeanRochCoulon:
Dashboard:
https://riscv-ci.pages.thales-invia.fr/dashboard/dashboard_cva6.html

AngelaGonzalezMarino:
Hi, I don't know why you see missing tests. The riscv tests are cloned via scripts which I did not modify, and I see that test in my local folder (jal.S in this case).

Re. Spike, it could be due to the issue with core-v-verif versions.

Master was pointing here when we started this work: 60e57248c48b0f8f90cf52af20aecc5c16e9d6d8

Master candidate before this pr points to: https://github.com/openhwgroup/core-v-verif/tree/b7a496781c6906a0cd8996ad426d7c36010840cb

In this PR, I point to same as master, because that is how I was able to run most of the tests we defined for this work (with Questa). But some tests like PMP only run in the other commit (previously pointed by master candidate).

@ASintzoff can you help on this? I see master updated core-v-verif last month. Should we also update?

JeanRochCoulon:
Spike ISA might be set twice in cva6.py

valentinThomazic:
Hi @AngelaGonzalezMarino!
In the CI, the simulation are run in tandem mode with Spike.
Do you have the same issue when running the tests locally with tandem?
You can enable it via exporting `SPIKE_TANDEM=1` env var.

AngelaGonzalezMarino:
Hi @valentinThomazic 
I do not see the problem locally. I exported SPIKE_TANDEM as you say, and set as simulators both questa-uvm and spike.
I can run the tests in questa without issues. Maybe VCS complains about something that Questa does not?

AngelaGonzalezMarino:
Not sure how much this helps:

- In cva6.py our changes only introduce new configs, like this one:
Screenshot from 2026-05-28 12-34-55

- In makefile, the changes apply only to Questa:
Screenshot from 2026-05-28 12-35-06

- In testbench, we add connections to the interfaces that were missing, plus the signals related to e-tracer

Screenshot from 2026-05-28 12-35-57

JeanRochCoulon:
Executing hello world on cv32a60x in Thales ci:

image

AngelaGonzalezMarino:
Hi @JeanRochCoulon, the previous test changing cvxif was for 65x, so I think it is expected that you saw the same in 60x. Now I did the same thing for 60x, can you try again?

About the topic of cleaning the workspace, I am not sure where in the yml file to put it and whether I can use the same environment variables. Since I cannot test it locally, @yanicasa or @ASintzoff can you help?

image (1)

valentinThomazic:
Hi @AngelaGonzalezMarino!
The dashboard does not reply to this PR nor generate a page with all the reports because it is not targeted at the master branch.
However, the reports are still generated and you can see them on the [dashboard main page](https://riscv-ci.pages.thales-invia.fr/dashboard/dashboard_cva6.html).

AngelaGonzalezMarino:
Thanks @valentinThomazic !
I am looking at the 65x test which the CI reports as PASS, but if you go inside the log it says FAILED.

However, I run the same test locally with questa and I get a PASS
` # UVM_INFO @ 32.000 ns : uvma_cvxif_drv.sv(218) uvm_test_top.env.cvxif_agent.driver [CVXIF_DRV] CVXIF Agent is disabled : All CPU Requests are rejected !!
`# UVM_INFO @ 652.000 ns : uvmt_cva6_firmware_test.sv(137) uvm_test_top [TEST] Started RUN
`# *** [rvfi_tracer] INFO: Simulation terminated after       7837 cycles!
`# 
`# UVM_INFO @ 156792.000 ns : uvmt_cva6_firmware_test.sv(143) uvm_test_top [TEST] Test FINISHED
`# ** Note: $finish    : /opt/questasim2023.4/questasim//verilog_src/uvm-1.2/src//base/uvm_root.svh(517)
`#    Time: 156812 ns  Iteration: 76  Instance: /uvmt_cva6_tb
`# 
`# uvmt_cva6_tb.end_of_test: *** Test Summary ***
`# 
`#     PPPPPPP    AAAAAA    SSSSSS    SSSSSS   EEEEEEEE  DDDDDDD     
`#     PP    PP  AA    AA  SS    SS  SS    SS  EE        DD    DD    
`#     PP    PP  AA    AA  SS        SS        EE        DD    DD    
`#     PPPPPPP   AAAAAAAA   SSSSSS    SSSSSS   EEEEE     DD    DD    
`#     PP        AA    AA        SS        SS  EE        DD    DD    
`#     PP        AA    AA  SS    SS  SS    SS  EE        DD    DD    
`#     PP        AA    AA   SSSSSS    SSSSSS   EEEEEEEE  DDDDDDD     
`#     ----------------------------------------------------------
`#                         SIMULATION PASSED                     
`#                  test exit code = 0 (0x00000000)
`#     ----------------------------------------------------------
`# End time: 10:41:16 on Jun 30,2026, Elapsed time: 0:01:15
`
In the CI it seems the failure comes from spike tandem reporter. I am using DV_SIMULATORS=questa-uvm,spike, and the report comparing spike with questa seems ok

spike : /home/angela/Documents/github/cva6_merge/mustafa/cva6/verif/sim/out_2026-06-30/spike_sim/hello_world.cv32a65x.csv
questa-uvm : /home/angela/Documents/github/cva6_merge/mustafa/cva6/verif/sim/out_2026-06-30/questa-uvm_sim/hello_world.cv32a65x.csv
[PASSED]: 829 matched

Do you think I am missing something in the setup?

valentinThomazic:
Yes it seems the simulation actually passes but prints "failed" which is weird.
Do you run your simulation in Tandem? Maybe it is the mismatches that make it print fail (even though 5 mismatches are allowed iirc).
Also, if you are, you do not need to specify spike in `DV_SIMULATORS`

AngelaGonzalezMarino:
@valentinThomazic  How do I set tandem? I thought specifying spike in simulators was enough

valentinThomazic:
Exporting `SPIKE_TANDEM=1` should work!

ASintzoff:
FWIW, on the CVA6 dashboard (https://riscv-ci.pages.thales-invia.fr/dashboard/dashboard_cva6.html) for Smoke test vcs-uvm cv32a65x (on CI for c82cd4e), the tandem report (https://riscv-ci.pages.thales-invia.fr/dashboard/logs/287757/artifacts/logs/hello_world.cv32a65x.log.yaml) and the TB logs (https://riscv-ci.pages.thales-invia.fr/dashboard/logs/287757/artifacts/logs/hello_world.cv32a65x.log.iss) contain four mismatches between Spike and CV32A65X.

Strangely, the mismatches are related to incorrect instructions.
Sometimes, a compressed instruction is considered as a 32-bit instruction by the test bench for the CVA6. For instance, a compressed `ret` pseudoinstruction encoded with `0x8082` is seen as encoded with `0x71798082`. `0x7179` corresponds to the first instruction of the next function.

I'm wondering if there is some problems with RVFI for the 65x configuration.

AngelaGonzalezMarino:
> Exporting `SPIKE_TANDEM=1` should work!

I did this, but I see the same. Test passes and I do not see any traces from spike-tandem reporter. Should I see something even in the event of pass?

One thing I noticed is that removing spike from the DV_SIMULATORS list I do not get the comparison in the overall iss_regr.log. Is this also ok?

valentinThomazic:
I think the arguments to enable spike tandem might not be passed to questa uvm run in `verif/sim/Makefile`.
There are theses lines for vcs and xrun but none for the questa uvm target.
{C00B514D-F94A-4024-8902-B11552BBD16E}
On Thales side, we would not know because we do not use this target. This would mean that you are indeed running simulations without tandem. The fix would be to pass that arg to vsim.

AngelaGonzalezMarino:
Thanks @valentinThomazic this makes a lot of sense. We are going to try to integrate spike-tandem with Questa and see if we can reproduce the problem locally.
Meanwhile I also suggest that you disable spike-tandem (maybe only for one config is enough). Just to see if the rest of the tests are OK or whether there are other issues.

.gitlab-ci.yml
valentinThomazic:
```suggestion
```

valentinThomazic (commented):
The configuration is in the gitlab-ci so you can tweak it as you want!
By removing this line, it will disable it (but on all configs).

AngelaGonzalezMarino:
@yanicasa can you try running the tests now with spike_tandem disabled? Thanks!

JeanRochCoulon:
@AlexChenIC  We get this Github Actions error:
image
Have you an idea howto fix it ?

cainria:
Hi! The GitHub CI fails because of a GPG issue, which was already addressed by https://github.com/openhwgroup/cva6/pull/2953. Could you try cherry-picking this commit to fix the CI?

AngelaGonzalezMarino:
I am not sure what the previous fix solved. Now I see that the github CI is trying to run some tests with verilator. This wont work because we use now the UVM environment (specially for the configs using obi)

cainria:
Here is what is being run: https://github.com/planvtech/cva6/blob/master_candidate_pr/.github/workflows/ci.yml

`veri-testharness` appears several times in the file. Maybe you can replace this with something like `vcs-uvm` ? But I am not sure whether the GitHub runners have VCS or not :thinking:

JeanRochCoulon:
Github Actions does not support VCS

JeanRochCoulon:
In that case, I will relaunch the Thales CI supporting vcs-uvm

AngelaGonzalezMarino:
I can run locally hello world for 65x_axi with veri-testharness and it passes. Can we get more visibility on why it fails in the github actions?

cainria:
https://github.com/openhwgroup/cva6/actions/runs/29017647967/job/86116826275?pr=3331#step:7:342

> FileNotFoundError: [Errno 2] No such file or directory: '../tests/testlist_riscv-tests-cv32a65x_axi-p.yaml'

AngelaGonzalezMarino:
> https://github.com/openhwgroup/cva6/actions/runs/29017647967/job/86116826275?pr=3331#step:7:342
> 
> > FileNotFoundError: [Errno 2] No such file or directory: '../tests/testlist_riscv-tests-cv32a65x_axi-p.yaml'

I have seen this, the issue is that I changed the target to _axi so the test files do not exist in that path.
However, it tries to run hello world anyway and it fails, this is what I do not understand
https://github.com/openhwgroup/cva6/actions/runs/29017647967/job/86116826275?pr=3331#step:7:359

cainria:
The script runs all the tests and does not stop after the first failure, hence the many errors and the hello-world running at the end. https://github.com/planvtech/cva6/blob/master_candidate_pr/verif/regress/cv32a6_tests.sh

For the hello-world error, unfortunately we do not have access to the artifacts from GitHub actions when a job fails.
Maybe you can try to branch from `master_candidate_pr` in your fork and cherry-pick #3372 ?

JeanRochCoulon:
The Thales ci fails: https://riscv-ci.pages.thales-invia.fr/dashboard/dashboard_cva6.html

AngelaGonzalezMarino:
@JeanRochCoulon can you launch the smoke tests for 65x? Not smoke-gen.
In those I disabled spike tandem so I would like to see if we are able to get the same result that I get locally. Thanks!

JeanRochCoulon:
image

AngelaGonzalezMarino:
I also see it failed, but I do not see any output logs 

image

AlexChenIC:
Hi @AngelaGonzalezMarino, I reproduced the GitHub Actions failure at e99e11c in my fork and tested a small CI update with two goals: 
1. Allow `_axi` configurations to reuse the existing ISA test lists. 
2. Preserve logs and artifacts when CI jobs fail.

For `cv32a65x_axi`, the simulation target remains `DV_TARGET=cv32a65x_axi`, while the test-list lookup uses the existing `cv32a65x` lists. This removes the previous missing-test-list error.

I also configured the full matrix to continue after individual job failures and to upload artifacts in all cases. The new run now exposes the next Verilator compilation error, which I am investigating separately.

Reference branch:
https://github.com/AlexChenIC/cva6/tree/jchen/pr3331-ci-e99e11c

Moreover, could you please confirm whether `cv32a65x_axi` is intended to share the ISA test lists with `cv32a65x`?

AngelaGonzalezMarino:
Thanks @AlexChenIC 
Yes, _axi configs use same ISA testlists as non _axi configs.
Keep in mind that non axi configs will fail in verilator because OBI is not supported in testharness testbench.
I would suggest to keep only _axi configs in the github actions for now.

AlexChenIC:
> Yes, _axi configs use same ISA testlists as non _axi configs.
> Keep in mind that non axi configs will fail in verilator because OBI is not supported in testharness testbench.
> I would suggest to keep only _axi configs in the github actions for now.

Thanks, that confirms the RV32 test-list mapping. Just to make sure I understand “keep only _axi configs” correctly: the existing RV64 CI targets do not have `_axi` in their names, but their configuration packages use `PipelineOnly=0` and `NOCType=NOC_TYPE_AXI4_ATOP`

Do you expect the six RV64 jobs to remain, while only the RV32 jobs use `cv32a65x_axi`?

AngelaGonzalezMarino:
Yes, you are absolutely right, I mean configs with pipelineonly=0 and NOCType=NOC_TYPE_AXI4_ATOP

Another thing to consider is that after the pipeline merge only hpdcache is supported, so configs with other cache should not be included in these tests either.

MikeOpenHWGroup:
> after the pipeline merge only hpdcache is supported, so configs with other cache should not be included in these tests either.

It is my understanding that the hpdcache is to permanently replace the dcache from the Ariane days.  So this is not a big restriction.

AlexChenIC:
Thanks @AngelaGonzalezMarino and @MikeOpenHWGroup for the clarification. Before making this CI matrix change, I would like to confirm the exact intended scope for PR #3331.

The RV64 smoke-test matrix currently contains four configurations. Based on my understanding, the two legacy non-HPDCache configurations should be removed:
- `cv64a6_imafdc_sv39`
- `cv64a6_imafdc_sv39_wb`

while the following two HPDCache configurations should remain:
- `cv64a6_imafdc_sv39_hpdcache`
- `cv64a6_imafdc_sv39_hpdcache_wb`

The two existing RV64 HPDCache architecture-test jobs and the two `cv32a65x_axi` jobs would also remain unchanged.

Including the build job, this would reduce the main `ci` workflow run from 9 job executions to 7.

Is this the intended CI matrix for PR #3331? @JeanRochCoulon, could you also confirm that this matches the intended upstream CI coverage?

AngelaGonzalezMarino:
I guess we could have 60x_axi, is this not there right now?

JeanRochCoulon:
For the verification of #3331 we agreed to verify 4 configurations:
- cv32a60x and cv32a65x in OBI mode
- cv64_imafdc_sv39 in AXI mode. HPDCache was not considered in the discussion that is why cv64a6_imafdc_sv39 and cv64a6_imafdc_sv39_wb can be used to verify cv64_imafdc_sv39.
- cv32a6_imac_sv32 in AXI mode.

AngelaGonzalezMarino:
but we cant run the obi configs with verilator, so for the github actions I would use the _axi versions. Then the obi versions are tested in the UVM environment

AlexChenIC:
Thanks @AngelaGonzalezMarino and @JeanRochCoulon for the clarification.

`cv32a60x_axi` already exists and is listed in the GitLab CI, but it is not currently included in the GitHub Actions RV32 matrix.

To keep the initial GitHub Actions update focused, my understanding is that the existing `cv32a65x_axi` target should remain and `cv32a60x_axi` should be added.

This would run both existing RV32 test cases, `dv-riscv-arch-test` and `cv32a6_tests`, for these two AXI target.

Does this match the intended RV32 GitHub Actions scope? If so, I can first prevalidate this exact matrix change in my fork.

AngelaGonzalezMarino:
Looks good to me!

AlexChenIC:
Hi @AngelaGonzalezMarino, following your confirmation, I added `cv32a60x_axi` alongside the existing `cv32a65x_axi` target in my fork's GitHub Actions matrix. This creates four RV32 GitHub Actions jobs, and each job now keeps downloadable artifacts even when it fails.

In the first run, all four RV32 jobs failed during Verilator elaboration with `Pin not found: 'early_valid_o'`:
https://github.com/AlexChenIC/cva6/actions/runs/29322801890

This appears to be an interface version mismatch. At `e99e11c`, PR #3331 pins `core/cvfpu` to `2c794772`. In that CVFPU version, the `fpnew_top` port list does not include `early_valid_o`:
https://github.com/openhwgroup/cvfpu/blob/2c7947726a0859c08cc27cbbd45662424dedb41e/src/fpnew_top.sv#L44-L57
However, the corresponding CVA6 `core/fpu_wrap.sv` connects that port:
https://github.com/openhwgroup/cva6/blob/e99e11c660b96aa1c4a0dda0f6c4eb631d8523ca/core/fpu_wrap.sv#L531-L558

As a change in my fork only, I disabled this unmatched 'early_valid_o' connection. In the following run, three of the four RV32 jobs passed, This indicates that the port mismatch was the blocker.The remaining `cv32a60x_axi` architecture-test job now fails later while compiling an A-extension test, which I will work on it later:
https://github.com/AlexChenIC/cva6/actions/runs/29331747297

Therefore, could you please confirm the intended CVFPU alignment?  Should this PR keep CVFPU `2c794772` and remove the unmatched `early_valid_o` connection, or should it update to a newer CVFPU version and keep that connection?

AngelaGonzalezMarino:
Thanks for the analysis @AlexChenIC !

I guess when we started this work the FPU was in an outdated commit. Since the TRL-5 version from Thales did not use FPU, I would say we should use the FPU version used in master and adjust the ports connection accordingly.

@JeanRochCoulon do you have a different opinion?

MikeOpenHWGroup:
> I would say we should use the FPU version used in master and adjust the ports connection accordingly.

I also think that is what we should do.  The version of the unused FPU in the cv32a6*x is very out-of-date and has multiple known bugs.

MikeOpenHWGroup:
Hi @udayaa-init.  It appears that you have pushed an update related to "_Cherry picked the Aes for 64 boot_" on top of the `planvtech:master_candidate_pr` fork.  Was that your intention?

udayaa-init:
Hi @MikeOpenHWGroup , I cherry-picked from the commit 6d9b76, and the file name was changed to cva6_pipeline.sv, so I had to create the patch. It was needed for the 64-bit Linux boot, as the current build-root uses the _zkn extension.

AngelaGonzalezMarino:
Hi @MikeOpenHWGroup 
Yes it is intended that @udayaa-init pushed here. He has been fixing some remaining issues in this branch. Some needed to port features from current master like the last one.
Overall this should reduce the gap between master and master candidate a little bit :)

AlexChenIC:
Hi @AngelaGonzalezMarino, following the `cv32a60x_axi` architecture-test failure observed during
my fork CI validation, I performed a deeper investigation. I also repeated the same validation against the current
`master_candidate_pr` head, `d8272e7f`, to check whether the previously identified CVFPU issue was still present.

### A. `cv32a60x` A-extension test-list mismatch
Both `cv32a60x` and `cv32a60x_axi` are configured with `RVA=0`, so their generated ISA does not include the A extension:
- [cv32a60x configuration](https://github.com/planvtech/cva6/blob/d8272e7f27844553f1275ce1154edcba5f921c49/core/include/cv32a60x_config_pkg.sv#L39)
- [cv32a60x_axi configuration](https://github.com/planvtech/cva6/blob/d8272e7f27844553f1275ce1154edcba5f921c49/core/include/cv32a60x_axi_config_pkg.sv#L39)

As previously discussed, the `_axi` target reuses the base `cv32a60x` ISA test list. However, that shared architecture test list still contained nine A-extension/AMO tests:
- [A-extension entries in the test list](https://github.com/planvtech/cva6/blob/d8272e7f27844553f1275ce1154edcba5f921c49/verif/tests/testlist_riscv-arch-test-cv32a60x.yaml#L566-L610)

The architecture test flow therefore generated an ISA string without `a`, but then attempted to assemble AMO instructions. This caused the assembler error `extension 'a' required`.

As a candidate correction in my fork, I removed only those nine test entries without changing the processor configuration:
- [Candidate test-list change](https://github.com/AlexChenIC/cva6/commit/2d6e504a1b6711b2dc4dc92f3edf21292288c3fc)

As a result, with the temporary FPU port bypass already present in this branch, all four RV32 AXI GitHub Actions jobs now pass:
- [RV32 AXI prevalidation run](https://github.com/AlexChenIC/cva6/actions/runs/29414263333)

### B. CVFPU issue on the current PR head
Separately, I repeated the validation on a fresh branch based on `d8272e7f`, with only the CI-related updates applied and without the temporary FPU port bypass:
- [Latest-head prevalidation branch](https://github.com/AlexChenIC/cva6/tree/jchen/pr3331-ci-d8272e7f)

The build job passed, but all test-matrix jobs stopped at the same CVFPU interface mismatch: `Pin not found: 'early_valid_o'`
- [Latest-head prevalidation run](https://github.com/AlexChenIC/cva6/actions/runs/29405589621)

This reproduces the existing CVFPU integration blocker on the current `master_candidate_pr` head.

Could you please confirm the following two points?
1. Is `cv32a60x` intended to remain configured with `RVA=0`? If so, is excluding these nine A-extension tests from its shared architecture-test list the intended correction?
2. Following the agreed direction to use the CVFPU revision currently used by CVA6 master and adapt the port connections, could you confirm whether the PlanV team plans to make this update directly on `master_candidate_pr`? Once the update is available, I can rerun the same GitHub Actions prevalidation.

AngelaGonzalezMarino:
Thanks a lot @AlexChenIC !

About point 1 I will let @JeanRochCoulon to give his feedback

About point 2, yes, we will do it soon

JeanRochCoulon:
I do not understand why you get this AMO issue. Verification infrastructure was ok on cv32a60x in OBI mode (AMO tests were disabled for instance). By defining the new cv32a60x_axi, the same verification infrastructure should be reuse to avoid regression (in that case, AMO tests would be already disabled). Reusing the same infrastructure reduces the work.

AlexChenIC:
Hi @JeanRochCoulon, thanks for the feedback. The successful OBI flow you referred to and my fork validation may
have used different test paths.

My understanding is that the Thales CI is triggered through `.gitlab-ci.yml`. In that flow, the `cv32a60x` job calls `smoke-tests-cv32a60x.sh`, which explicitly selects only five basic tests. Therefore, this smoke-test flow does not run any AMO tests.:
https://github.com/planvtech/cva6/blob/24a8077e250b880307c9c25eb471fa9a28be1ef1/verif/regress/smoke-tests-cv32a60x.sh#L51-L67

The failure I observed came from the GitHub Actions `ci.yml` workflow using the `dv-riscv-arch-test` flow. This flow reads the `cv32a60x` architecture-test list. In the current `master_candidate_pr` version, that list still contains nine AMO tests, even though the `cv32a60x` configuration has `RVA=0`:
https://github.com/planvtech/cva6/blob/24a8077e250b880307c9c25eb471fa9a28be1ef1/verif/tests/testlist_riscv-arch-test-cv32a60x.yaml#L566-L610

This test list mismatch has already been addressed upstream in OpenHW CVA6 PR #3285, which removes the unsupported AMO entries from the `cv32a60x` test list. However, the current `master_candidate_pr` head does not appear to include that upstream change and still contains the older AMO entries.

Could you please confirm whether the intended solution is to bring the relevant test list updates from #3285 into `master_candidate_pr`? Also, could you point me to the successful CV32A60X OBI CI job or configuration you mentioned? I can check whether it ran the full architecture test list or excluded the AMO tests through another mechanism.

JeanRochCoulon:
Thank @AlexChenIC for your detailed. It is very clear.

In master, the Thales ci execute https://github.com/openhwgroup/cva6/blob/master/verif/regress/dv-riscv-arch-test.sh with DV_TARGET=cv32a65x each night.

But this job has been removed from cv32a60x TRL5 branch (so it is not present in master_candidate !). Be careful, the target is cv32a65x and not cv32a60x.

The targeted final strategy is to create dedicated testlist by extension, and create related job to execute the testlist.
For example, testlist-arch-test-amo.yaml for the 9 amo tests. This is a new strategy.

Create such a files is a good idea and would allow you to make the arch-test pass.

AlexChenIC:
Hi @AngelaGonzalezMarino and @JeanRochCoulon, I revalidated the current PR head `b9e7eeb` after the CVFPU submodule update. This branch is based directly on the PR head and does not include the previous temporary `fpu_wrap.sv` port bypass.

For this validation, I applied the same three `cv32a60x` test-list alignments already merged in #3285, together with the previously discussed AXI test-list mapping and CI artifact updates. The build job and all four RV32 AXI jobs passed, and the previous `early_valid_o` elaboration error no longer appears: https://github.com/AlexChenIC/cva6/actions/runs/29487373227

This provides evidence that the CVFPU integration issue is resolved for the `cv32a60x_axi` and `cv32a65x_axi` GitHub Actions subset. It does not cover the separate OBI/UVM verification.

@JeanRochCoulon, your latest comment proposes dedicated test lists and jobs per extension. For the immediate #3331 unblock, should we first bring the already-merged #3285 test-list alignment into `master_candidate_pr` and track the extension specific structure as a follow-up, or do you expect that new structure to be implemented as part of #3331?

AngelaGonzalezMarino:
@AlexChenIC That is great news!
I have reverted disabling spike tandem and fixed the issue that caused the mismatch in 65x configs.
I am not sure if you use spike tandem in the github actions, but we saw this problem in the Thales CI.

AlexChenIC:
> @AlexChenIC That is great news! I have reverted disabling spike tandem and fixed the issue that caused the mismatch in 65x configs. I am not sure if you use spike tandem in the github actions, but we saw this problem in the Thales CI.

Hi @AngelaGonzalezMarino. I checked the artifacts from my fork Github action CI validation. Yes, the RV32 GitHub Actions jobs did run with Spike tandem enabled. The Verilator build included `+define+SPIKE_TANDEM`, and the runtime logs show `Running binary in tandem mode`

AlexChenIC:
Hi @AngelaGonzalezMarino, I revalidated the latest PR head `8bcee9ce` after the RVFI fix.  The build and all four RV32 AXI jobs passed with `SPIKE_TANDEM=1`:
https://github.com/AlexChenIC/cva6/actions/runs/29818403667

The fork branch is based on that PR head and includes only the previously discussed ci.yml and rv32 axi test-list candidate changes. The artifacts confirm that the tests entered tandem mode and started Spike. I did not find any tandem-mismatch in the four RV32 jobs. 

The complete workflow is still red because four separate RV64 jobs failed, so this conclusion is limited to the RV32 subset.

Therefore, the initial four Github Action RV32 job matrix was already agreed earlier. To determine whether the RV32 GitHub Actions scope can now be considered complete, could you please confirm how the AXI `cv32a6_imac_sv32` configuration previously listed by Jean-Roch should be covered? Should it also be added to GitHub Actions now, or is it intended to be covered by another CI flow?

Moreover, I would like to confirm the intended acceptance scope for GIthub Action RV64: Which exact RV64 configurations and test suites should be included in the GitHub Actions? In the latest fork run, both `cv64a6_imafdc_sv39_hpdcache` jobs passed, while the HPDCache-WB and legacy-cache jobs failed.

AngelaGonzalezMarino:
Hi @AlexChenIC 

Based on the discussion at cva6 meeting I would say it would be good to cover cv32a6_imac_sv32  in the github actions. The Thales CI will do very basic tests on it only. I would recommend to run the mmu tests at least.

About cache:

- legacy jobs are expected to fail, no other caches than the hpdcache are supported now.

- HPDCache-WB: I never used this configuration so it might very well fail. @cfuguet @JeanRochCoulon Is this configuration needed after the merge? And maybe silly question, is it currently functional in master?

cfuguet:
> Hi @AlexChenIC
> 
> Based on the discussion at cva6 meeting I would say it would be good to cover cv32a6_imac_sv32 in the github actions. The Thales CI will do very basic tests on it only. I would recommend to run the mmu tests at least.
> 
> About cache:
> 
>     * legacy jobs are expected to fail, no other caches than the hpdcache are supported now.
> 
>     * HPDCache-WB: I never used this configuration so it might very well fail. @cfuguet @JeanRochCoulon Is this configuration needed after the merge? And maybe silly question, is it currently functional in master?

Hi @AngelaGonzalezMarino,

Yes, this configuration works in master. All non-regression tests (on the Github CI) pass and it is also able to run Linux on FPGA.

This configuration is used in particular by the PULP team. They always use the WB version of the HPDcache.

AngelaGonzalezMarino:
> > Hi @AlexChenIC
> > Based on the discussion at cva6 meeting I would say it would be good to cover cv32a6_imac_sv32 in the github actions. The Thales CI will do very basic tests on it only. I would recommend to run the mmu tests at least.
> > About cache:
> > ```
> > * legacy jobs are expected to fail, no other caches than the hpdcache are supported now.
> > 
> > * HPDCache-WB: I never used this configuration so it might very well fail. @cfuguet @JeanRochCoulon Is this configuration needed after the merge? And maybe silly question, is it currently functional in master?
> > ```
> 
> Hi @AngelaGonzalezMarino,
> 
> Yes, this configuration works in master. All non-regression tests (on the Github CI) pass and it is also able to run Linux on FPGA.
> 
> This configuration is used in particular by the PULP team. They always use the WB version of the HPDcache.

Ok, we will have a look then

AlexChenIC:
Hi @AngelaGonzalezMarino and @JeanRochCoulon, following our discussion about adding MMU coverage for `cv32a6_imac_sv32`, I added a dedicated GitHub Actions job on my fork branch based on the final PR head `8bcee9ce`. For clarity, this branch also contains the previously discussed CI, artifact, AXI test-list, and matrix candidate updates, but no RTL change.

CI run:
https://github.com/AlexChenIC/cva6/actions/runs/29859293427

In this run, the build and all four previously agreed RV32 AXI jobs passed. The new `cv32a6_imac_sv32` MMU job entered Verilator simulation with Spike tandem enabled, matched Spike for the first 23 MMU/SV32 tests, and then failed on `rv32_vm_pmp_check_level_0_s`: https://github.com/AlexChenIC/cva6/actions/runs/29859293427/job/88735680832

```text
vm_pmp_check_level_0.o *** FAILED *** (tohost = 1) after 1051 cycles
*** [rvfi_tracer] INFO: Simulation terminated after 976 cycles!
```

The detailed logs can be downloaded from the run page under **Artifacts**: `veri-testharness.dv-riscv-mmu-sv32-test.cv32a6_imac_sv32`

The most relevant files are:
- `out_2026-07-21/iss_regr.log`, which shows the first 23 Spike/TestHarness comparisons passing;
- `out_2026-07-21/veri-testharness_sim/rv32_vm_pmp_check_level_0_s.cv32a6_imac_sv32.log.iss`, which contains the failure above.

This is a failure after successful build and elaboration. This symptom appears to be similar to the MMU/PMP behavior already tracked in #3174, although this anomaly alone is not sufficient to determine the root cause at the RTL level.

I also saw that #3331 has now been merged into `openhwgroup:master_candidate`. Could you please clarify the preferred follow-up workflow? Should I continue running fork-only GitHub Actions validation against the official `master_candidate` branch and report findings through the corresponding issues, or is another follow-up branch or CI coordination path planned?

AngelaGonzalezMarino:
Hi @AlexChenIC Thanks for the update!

The failure of mmu tests in pmp checks happens also in master branch. I do not know if the root cause is related to what you mention. If this is the only failure I would say we are OK, since behavior is same as master branch.

The next step is that @yanicasa will push his PR for the new GitLab CI. Then we will be able to have the results of that one too.
I would say afterwards you can merge your changes if you are ready, so that we get all the results together.

On PlanV side we are also preparing an update with some fixes for the WB configuration, enabling SPIKE_TANDEM with Questa and further tests available.

Issue #172, opened 2019-02-02
ndmreset


suppamax:
Hi,

I've noticed that in dm_csrs.sv you modified the generation of ndmreset_o
From
ndmreset_o = dmcontrol_q.ndmreset | (~rst_ni);
to
ndmreset_o = dmcontrol_q.ndmreset;

Since the core's only reset is ndmreset, with the new version it won't get reset when the global reset is triggered.
What's the reason of this choice?

suppamax:
I see, the control has been moved to top level

PR #3335, opened 2026-06-02
Fix AMO operation cast in WT L15 adapter.

AnaBSF:

Bumping CVA6 in OpenPiton, which requires wt_l15_adapter, reveals that AMO ops are not being correctly sent to L1.5, and the simulation hangs. This PR fixes the cast bug that was causing all incoming CVA6 AMO operations to permanently map to AMO_NONE (0).


cainria (approved):
The change looks good to me, however without context or knowledge on this topic, I'm wondering why we define two identical types `l15_pkg::l15_amo_e` and `ariane_pkg::amo_t`.

The cast assumes that these types are similar, so if we modify one, we will have to modify the other to not have bugs. Having only one type would eliminate this issue, and no casts would be needed.

Is there any reason why we have this type defined twice?

cfuguet:
Hi @AnaBSF,

I'll let @Jbalkind confirm, but I think the issues you are encountering are due to a transition to the HPDcache cache subsystem. This will be the only officially supported cache subsystem (both legacy WT and STD/WB cache subsystems are being deprecated progressively).

A recent PR added full-support for OpenPiton using the HPDcache.

AnaBSF:
@cfuguet 

Yes, I am aware, as I have been following that transition. In the meantime, this is a simple fix so that the most recent CVA6 can be used in OpenPiton with the legacy cache. I have also checked OpenPiton's open PR related to this issue ([#172](https://github.com/PrincetonUniversity/openpiton/pull/172)).

JeanRochCoulon:
By the end of august the legacy WT and WB caches will be no more supported and removed from the CVA6 RTL. For this reason, any related modification is useless. I close the PR.

PR #3341, opened 2026-06-08
cache_subsys: cut timing path of AMOs with a buffer

cfuguet:

Recent ASIC synthesis with the CVA6+HPDcache exhibit a critical path concerning the response path for AMOs (Atomic Memory Operations) from the cache to the CVA6.

This PR adds a pipeline buffer into this path to enhance the timing (clock frequency). However, it increases the latency of AMOs by one cycle. This extra cycle is anyway acceptable as AMOs are not part of the the most frequent operations, and the clock frequency speedup largely compensates this.


JeanRochCoulon:
@cfuguet   I took the right to update the gate count to make the synthesis CI pass.

cfuguet:
@JeanRochCoulon of course ! thanks

PR #3352, opened 2026-06-16
3136 - Fix flush_bp_i for M-mode-only configurations

gm-openchip:
* [x] I have searched for similar pull requests
* [x] I am a human engaging in an interpersonal interaction. During this interaction, my words are my own and are not generated. If relevant, I provide links to my sources.

## Why this PR is needed

`flush_bp_i` in `i_frontend` (`core/cva6.sv`) has been hardwired to `1'b0` since the first commit, meaning the branch predictor is never flushed on privilege mode transitions (exceptions and `eret`).

The controller already generates `flush_bp_o` on exceptions and `eret`, and this signal is captured as `flush_ctrl_bp` in `cva6.sv`. However, it is currently not connected to the frontend.

As a result, branch predictor state can persist across privilege boundaries (e.g. M↔S, M↔U, S↔U), potentially causing stale predictions to be used after a privilege transition.

In particular, this can lead to:

* Instruction access faults during kernel boot when stale virtual addresses from lower privilege modes remain in the predictor while executing in M-mode.
* Potential timing side-channel leakage across privilege domains through predictor state reuse.

## Change

This PR connects the existing controller-generated flush signal to the frontend branch predictor flush input:

```systemverilog
.flush_bp_i ((CVA6Cfg.RVU || CVA6Cfg.RVS) ? flush_ctrl_bp : 1'b0),
```

Since `CVA6Cfg.RVU` and `CVA6Cfg.RVS` are compile-time configuration constants, synthesis tools will optimize this expression to `1'b0` for M-mode-only configurations, preserving zero-propagation and eliminating unnecessary logic in those variants.

## Related issues

Related to issue #3136.

## Limitations

This change flushes the branch predictor on every exception and `eret`, which may have a small performance impact on systems with frequent privilege transitions.

The existing TODO in `controller.sv` regarding removal of this flush mechanism once PMA checkers are introduced is not addressed by this PR. This change only ensures that the already-generated flush signal is correctly propagated to the frontend.

Issue #3136, opened 2025-11-06
[BUG] frontend's flush_bp_i tied to 1'b0


gm-openchip:
### Is there an existing CVA6 bug for this?

- [x] I have searched the existing bug issues

### Bug Description

Hi,
I can't find any info on why the `flush_bp_i` signal, input to `i_frontend` in cva6.sv is always tied to `1'b0`.
https://github.com/openhwgroup/cva6/blob/89986dfe867dada9752eee19a4ed4a9d274484a6/core/cva6.sv#L682
It seems to always have been like this since the first commit.

I checked in my emulation run and the missing flush is triggering some bad prediction during kernel boot. In particular, it seems that while in M-MODE (sbi calls) bp still contains "dirty addresses" from previous Linux Kernel operations (S-Mode).

Since in M-Mode MMU translation is disabled, from time to time, a virtual address passes through and causes `FE_INSTR_ACCESS_FAULT`.

This seems not be causing any particular issue, but just curious if it's a bug or a wanted feature.

<img width="1486" height="515" alt="Image" src="https://github.com/user-attachments/assets/23a1439c-0bdb-4148-b910-9a9e8da18ae1" />

PS: `flush_ctrl_bp` (https://github.com/openhwgroup/cva6/blob/89986dfe867dada9752eee19a4ed4a9d274484a6/core/cva6.sv#L1300C31-L1300C44) is correctly generating flushes at when switching privilege level, but is not tied to anything..

cainria:
I guess flushing branch prediction would be useful to avoid timing side channel attacks when we switch to user mode.
But you are right, it has never been connected.

We should probably connect but still make sure that synthesis tools still propagate zeroes when relevant (eg: there are no user modes in this hardware configuration).

Feel free to send a PR!

gm-openchip:
Thanks for confirmation.
I'm going to change the logic on the frontend to this: 
``` Verilog
.flush_bp_i ((CVA6Cfg.RVU || CVA6Cfg.RVS) ? flush_ctrl_bp : 1'b0),
```
which should be enough to propagate zero when HW configuration foresees no other conditions but M-Mode only.

Created PR --> https://github.com/openhwgroup/cva6/pull/3352

cainria:
Can we close this issue as the associated PR has been merged?

gm-openchip:
Merged with https://github.com/openhwgroup/cva6/pull/3353

PR #3353, opened 2026-06-16
fix for 3136 - tie frontend's flush_bp_i to flush_ctrl_bp

gm-openchip:
Reopening https://github.com/openhwgroup/cva6/pull/3352 to fix for Eclipse ECA

## Why this PR is needed

`flush_bp_i` in `i_frontend` (`core/cva6.sv`) has been hardwired to `1'b0` since the first commit, meaning the branch predictor is never flushed on privilege mode transitions (exceptions and `eret`).

The controller already generates `flush_bp_o` on exceptions and `eret`, and this signal is captured as `flush_ctrl_bp` in `cva6.sv`. However, it is currently not connected to the frontend.

As a result, branch predictor state can persist across privilege boundaries (e.g. M↔S, M↔U, S↔U), potentially causing stale predictions to be used after a privilege transition.

In particular, this can lead to:

* Instruction access faults during kernel boot when stale virtual addresses from lower privilege modes remain in the predictor while executing in M-mode.
* Potential timing side-channel leakage across privilege domains through predictor state reuse.

## Change

This PR connects the existing controller-generated flush signal to the frontend branch predictor flush input:

```systemverilog
.flush_bp_i ((CVA6Cfg.RVU || CVA6Cfg.RVS) ? flush_ctrl_bp : 1'b0),
```

Since `CVA6Cfg.RVU` and `CVA6Cfg.RVS` are compile-time configuration constants, synthesis tools will optimize this expression to `1'b0` for M-mode-only configurations, preserving zero-propagation and eliminating unnecessary logic in those variants.

## Related issues

Related to issue #3136.

## Limitations

This change flushes the branch predictor on every exception and `eret`, which may have a small performance impact on systems with frequent privilege transitions.

The existing TODO in `controller.sv` regarding removal of this flush mechanism once PMA checkers are introduced is not addressed by this PR. This change only ensures that the already-generated flush signal is correctly propagated to the frontend.

### DIFF
--- a/core/cva6.sv
+++ b/core/cva6.sv
@@ -690,7 +690,8 @@ module cva6
       .rst_ni,
       .boot_addr_i        (boot_addr_i[CVA6Cfg.VLEN-1:0]),
       .flush_bp_i         ((CVA6Cfg.RVU || CVA6Cfg.RVS) ? flush_ctrl_bp : 1'b0),
-      .flush_i            (flush_ctrl_if),                  // not entirely correct
+      // below line is not entirely correct
+      .flush_i            (flush_ctrl_if),
       .halt_i             (halt_ctrl),
       .halt_frontend_i    (halt_frontend),
       .set_pc_commit_i    (set_pc_ctrl_pcgen),
--- a/core/cva6.sv
+++ b/core/cva6.sv
@@ -689,7 +689,7 @@ module cva6
       .clk_i,
       .rst_ni,
       .boot_addr_i        (boot_addr_i[CVA6Cfg.VLEN-1:0]),
-      .flush_bp_i         (1'b0),
+      .flush_bp_i         ((CVA6Cfg.RVU || CVA6Cfg.RVS) ? flush_ctrl_bp : 1'b0),
       .flush_i            (flush_ctrl_if),                  // not entirely correct
       .halt_i             (halt_ctrl),
       .halt_frontend_i    (halt_frontend),


JeanRochCoulon:
@gm-openchip  thanks for your CVA6 interest.  Could you execute the Verible command present in CONTRIBUTING.md to fix the format issue ?

JeanRochCoulon:
@yanicasa Can you approve this PR ?

gm-openchip:
Hi, Verible has been executed. I didn't include it in the PR cause the comment at [this line](https://github.com/openhwgroup/cva6/blob/3f39c6830d333ecc40b28790f8f9a1de6463124d/core/cva6.sv#L693) was breaking Verible formatting.
Suggestion was to format the comment differently, but that's out of the scope of this PR and I didn't want to mess with things I didn't want to touch.

That's how Verible updates the formatting for core/cva6.sv (even without the modifications introduced with this PR)
image

If I remove the comment, it passes with no issue and format isn't broken.
image

JeanRochCoulon:
We cannot merge if a Verible error is present. That is why I recomand to remove the comment.

gm-openchip:
Done

yanicasa:
Is the flush triggered only by mode switching, or by every exception, including those that occur within the same mode?

gm-openchip:
@yanicasa by looking at the code inside controller.sv, it actually fires on every exception (`ex_valid_i`), every `eret` and every debug entry `set_debug_pc_i`, not just on privilege mode switches.

It does make sense 'cause after an exception privilege level may change (`ecall` in cva6 triggers an `ex_valid`) and without PMA checker the frontend can't understand if a speculative fetch is based on prediction built in the previous priv-level is still secure. So, I think flushing also at every exception is the safer way to proceed.

In context in which only M-mode is foresaw, so no virtual memory though, and no S-Mode or U-Mode or privilege jumps, bp flush is still tied to 0 as flushing during exception-only may result in performance drop, as also stated in the comments.

yanicasa:
The frontend indeed doesn’t have that information. However, could the controller be modified to distinguish between exceptions that change the privilege mode and normal ones? The idea would be to generate a dedicated flush only when the mode actually changes, instead of flushing on every exception, in order to avoid unnecessarily invalidating the branch prediction history. Or maybe I’m mistaken?

gm-openchip:
@yanicasa , I think in the long term the right approach is what is suggested in the comments, so implements PMA checker to validate every speculative fetch in the frontend thus eliminating the necessity of BP.
This is another full topic IMHO.

gm-openchip:
@yanicasa The proposed solution is an intermediate solution; consider that also FE_INSTR_ACCESS_FAULTS "consumes" pipeline cycles, hence has some performance impact. In my tests even running some benchmaks, I wasn't able to spot any significative difference in terms of stalled cycles when comparing the original with the proposed solution.

yanicasa:
Sounds good to me!

JeanRochCoulon:
@gm-openchip Before merging, can I ask you to open an GitHub issue reporting the "work to be done" to complete the flush optimization ?  Maybe just a link to this PR discussion

gm-openchip:
@JeanRochCoulon TASK opened --> https://github.com/openhwgroup/cva6/issues/3358

JeanRochCoulon:
Thanks @gm-openchip @yanicasa I merge !

PR #3372, opened 2026-07-01
Fix the Ci to always dump simulation artifacts

vverdillon:
Add yaml rules to dump CI artifacts in case of failure

The current CI does not dump artifacts when a job fails
so it can be very hard to identify, reproduce and fix a bug that occurs
during CI execution. With this patch, we force dumping
simulation outputs when a CI error occurs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -115,12 +115,14 @@ jobs:
     - name: Run Tests
       run: |
           set -x
+          mkdir -p "verif/sim/out_$(date +%F)"
           export RISCV=$(pwd)/tools/riscv-toolchain/
           source ci/install-prereq.sh
           source verif/sim/setup-env.sh
           DV_SIMULATORS=${{matrix.simulator}},spike DV_TARGET=${{matrix.config}} bash verif/regress/${{matrix.testcase}}.sh
 
     - name: Upload Lint Report to Github
+      if: always()
       uses: actions/upload-artifact@v4
       with:
         name: ${{matrix.simulator}}.${{matrix.testcase}}.${{matrix.config}}
@@ -181,12 +183,14 @@ jobs:
     - name: Run Tests
       run: |
           set -x
+          mkdir -p "verif/sim/out_$(date +%F)"
           export RISCV=$(pwd)/tools/riscv-toolchain/
           source ci/install-prereq.sh
           source verif/sim/setup-env.sh
           DV_SIMULATORS=${{matrix.simulator}},spike DV_TARGET=${{matrix.config}} bash verif/regress/${{matrix.testcase}}.sh
 
     - name: Upload Lint Report to Github
+      if: always()
       uses: actions/upload-artifact@v4
       with:
         name: ${{matrix.simulator}}.${{matrix.testcase}}.${{matrix.config}}


MikeOpenHWGroup:
Hi @pulsar2105, thanks for your contribution.  I am not convinced that this update is a good idea.  The whole purpose of a CI regression is to create a simple (but not too simple), _fast_ regression to provide a "thumbs up" or "thumbs down" gate for accepting pull-requests.

Typically, if one of my pull-requests fails a CI test, I will re-run it locally with waves and increased verbosity.

vverdillon:
Hi, @MikeOpenHWGroup  in this case I don't understand at all why the CVA6 CI dump artifacts when tests success.
Currently, if the CI success to run all tests its dump artifacts, but this gives no additional information because its success. 

I'm currently working on a 128 bits version of the CVA6 at TIMA lab and I encounter some bugs about environment setup that only occurs in the CI environment and are very hard to reproduce, so I am very interested to recover artifacts when CI crash to understand why it crashes. I added this feature in my fork and actively use it and thought it has a good idea to share it in the main project.

cainria:
I agree with @pulsar2105, artifacts are most useful when jobs fail, and the environment is sometimes hard to reproduce accurately.
Contributors often have mismatches between local and CI setup, and artifacts can help debugging the local setup. This is also true for regular contributors (I think there are such issues in the `master_candidate` merge PR).

I don't know how big artifacts are but if we can afford their upload when the thumb is up, so we should when the thumb is down.

.github/workflows/ci.yml
```diff
@@ -115,12 +115,14 @@ jobs:
     - name: Run Tests
       run: |
           set -x
+          mkdir -p "verif/sim/out_$(date +%F)"
```
cainria, line 118: Isn't this folder created during the bash script? Do we need to add this line?

.github/workflows/ci.yml
```diff
@@ -115,12 +115,14 @@ jobs:
...
           export RISCV=$(pwd)/tools/riscv-toolchain/
           source ci/install-prereq.sh
           source verif/sim/setup-env.sh
           DV_SIMULATORS=${{matrix.simulator}},spike DV_TARGET=${{matrix.config}} bash verif/regress/${{matrix.testcase}}.sh
 
     - name: Upload Lint Report to Github
+      if: always()
```
cainria, line 125: You mentioned that you are using this in your own fork so I will not double-check that this is the correct way to proceed ;)

.github/workflows/ci.yml
vverdillon: I figured that since we're forcing the dumping of artifacts, it would be safer if the CI could dump something even if the directory is empty. But if you think it useless I'm good to go delete it.

MikeOpenHWGroup:
> if we can afford their upload when the thumb is up, so we should when the thumb is down.

As I say, this is not my decision.  I would not dumb anything on a CI,beyond pass/fail.  But that's me.

JeanRochCoulon:
For information, the plan is to superseed this CI by a new CI during Summer. I propose to close this PR, and maybe have a talk in the coming months about the support this feature on the new CI.

vverdillon:
Thank you @JeanRochCoulon for clarifying that. I think indeed that if nobody have seen this problem of artifacts dumping since May 2024 that is a sign that the role of the CI need to be rethought. I would be happy to help in this direction in the future and even provide a Dockerfile to run the CVA6 when the time will be right.

PR #3382, opened 2026-07-07
Draft: Add support for the CHERI "RV64Y" extension

PRugg-Cap:

As discussed, at Capabilities Limited based on initial work at Zero Day Labs, we have been integrating CHERI (the RISC-V Y extension): see https://github.com/riscv/riscv-cheri. This adds "capabilities" in hardware to provide memory safety and compartmentalisation that allows software to protect itself against common security vulnerabilities like buffer overflows.

This is a draft PR with the current state of our changes to get feedback on a path to upstreaming. I have extracted out only the CHERI features from our fork https://github.com/Capabilities-Limited/cheri-cva6, along with the minimal changes required for tests to pass etc. Only the last commit "Add RV64Y support" is relevant for review: other things will be addressed in other PRs. It would be very helpful to get feedback on the code style, if there are any changes that can be extracted out and made upstream separately, and if there are any changes we can make to the CHERI implementation to reduce the diff.

This PR does not introduce CHERI-specific tests. We primarily use TestRIG for this testing. We understand this is a requirement for upstreaming, but the aim of this PR is to establish that we can integrate CHERI without regressing the baseline design. Since RVY is backwards compatible, all the existing tests in this repository we have tried pass, both with and without CHERI enabled in.

A few other notes on limitations:
 - As discussed, this will not be merged until the RISC-V Y extension is ratified.
 - The current implementation is based on version 0.9.4 of the RISC-V Y spec. Updating is a work in progress, but changes are relatively minor.
 - Various further updates and optimisations are planned for the CHERI implementation.
 - We need to pull forward the version of verilator used, and there is an associated workaround: please ignore. We can discuss this closer to ratification if it's still an issue.
 - I've made a change to how debug single-step is implemented. I believe this should be an area saving for non-CHERI but still need to evaluate this. To be discussed in a separate PR.
 - I currently have reverted the big-endian support, as it's a bit of a pain to merge CHERI with those changes. I will of course resolve the conflicts properly prior to upstreaming.


core/issue_read_operands.sv
```diff
@@ -1039,18 +1215,18 @@ module issue_read_operands
...
-    pc_n = '0;
+    pc_n = pcc_q;
     is_compressed_instr_n = 1'b0;
     branch_predict_n = {cf_t'(0), {CVA6Cfg.VLEN{1'b0}}};
     if (CVA6Cfg.SuperscalarEn) begin
       if (issue_instr_i[1].fu == CTRL_FLOW) begin
-        pc_n                  = issue_instr_i[1].pc;
+        pc_n = cva6_cheri_pkg::set_cap_reg_address(pcc[1], issue_instr_i[1].pc, pcc_meta);
         is_compressed_instr_n = issue_instr_i[1].is_compressed;
         branch_predict_n      = issue_instr_i[1].bp;
```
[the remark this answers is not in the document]
PRugg-Cap: Hmm, not sure why the verible check is failing when it works on our fork. I'll have a look at that.

JeanRochCoulon:
Great @PRugg-Cap !  Great step !

core/include/ariane_pkg.sv
```diff
@@ -28,6 +29,12 @@
 /// moved out to favour a fully parameterizable core.
 package ariane_pkg;
 
+  localparam XLEN = cva6_config_pkg::CVA6ConfigXlen;
+  localparam CheriPresent = cva6_config_pkg::CVA6ConfigRVZcheripurecap;
+  localparam CLEN = CheriPresent ? 2 * XLEN : XLEN;
+  localparam REGLEN = CheriPresent ? $bits(cva6_cheri_pkg::cap_reg_t) : XLEN;
+  localparam logic [REGLEN-1:0] REG_NULL = CheriPresent ? cva6_cheri_pkg::REG_NULL_CAP : '0;
+  localparam logic [REGLEN-1:0] REG_ROOT = CheriPresent ? cva6_cheri_pkg::REG_ROOT_CAP : '0;
```
cainria, line 37: Have you considered using the config structure instead?

core/include/ariane_pkg.sv
```diff
@@ -713,6 +793,93 @@ package ariane_pkg;
...
   // generate byte enable mask
+  function automatic logic [15:0] be_gen_128(logic [3:0] addr, logic [2:0] size);
+    case (size)
+      // 16-bytes size
+      3'b100: begin
+        return 16'b1111_1111_1111_1111;
+      end
+      // 8-bytes size
+      3'b011: begin
+        case (addr[3:0])
```
cainria, line 804: Would the following shorter code be correct?

```
if (addr[3:0] <= 16) begin
  return 16'h00ff << addr[3:0];
end
```

core/cva6.sv
```diff
@@ -1,4 +1,6 @@
 // Copyright 2017-2019 ETH Zurich and University of Bologna.
+// Copyright 2025 Bruno Sá and Zero-Day Labs.
+// Copyright 2025 Capabilities Limited.
```
cainria, line 3: Is this copyright compatible with the CVA6 project?

.gitmodules
```diff
@@ -39,7 +39,7 @@
 	url = https://github.com/openhwgroup/cvfpu.git
 [submodule "core/cache_subsystem/hpdcache"]
 	path = core/cache_subsystem/hpdcache
-	url = https://github.com/openhwgroup/cv-hpdcache.git
+	url = https://github.com/Capabilities-Limited/cv-hpdcache.git
```
cainria, line 42: We will need to revert this before merge

.gitmodules
```diff
@@ -52,3 +52,6 @@
 [submodule "corev_apu/fpga/src/gpio"]
 	path = corev_apu/fpga/src/gpio
 	url = https://github.com/pulp-platform/gpio.git
+[submodule "vendor/zero-day/axi_tagcontroller"]
+	path = vendor/zero-day/axi_tagcontroller
+	url = https://github.com/Capabilities-Limited/axi_cheri_tagcontroller.git
```
cainria, line 57: Are we willing to add a new submodule?

cainria (commented):
Really partial review with a few questions for both the author and the project maintainers.

.gitmodules
PRugg-Cap: Okay: we have some non-trivial changes to the HPDCache to add capability tag support, but agreed, we should merge those there then this will become a bump of the existing submodule.

core/cva6.sv
PRugg-Cap: They're the same as the ETH and Bologna ones just above. All the changes are of course made under the indicated license.

core/include/ariane_pkg.sv
PRugg-Cap: I believe so. This mirrors the coding style used in the existing `be_gen` function just below. We assumed it was expressed this way for a reason, but I guess we should evaluate and possibly change both to be more concise.

core/include/ariane_pkg.sv
PRugg-Cap: It would be nice. The problem is that I want to introduce the functions `reg_to_x` and `x_to_reg`, which need to take a REGLEN and return an XLEN, and vice versa. That means they can't be taken from the config struct, as they have to be more static. `CVA6Cfg.CLEN` and `CVA6Cfg.REGLEN` are used elsewhere where they can be.

PRugg-Cap: If I'm misunderstanding something about how the types can be used, do let me know though!

.gitmodules
PRugg-Cap: It doesn't seem like this is a question for us, but just to give context: the tag controller is needed to provide tagged memory, splitting tags out from the AXI requests coming out of the caches and storing them in a dedicated region of DRAM. It definitely "feels" like its own component, much like the HPDCache. If the question is whether we should vendor instead, then we'd likely be happy with that approach.

core/include/deprecated_packages/cv64a6_imadfcv_sv39_polara_config_pkg.sv
```diff
@@ -28,6 +28,8 @@ package cva6_config_pkg;
   localparam CVA6ConfigHExtEn = 0;
   localparam CVA6ConfigVExtEn = 1;
   localparam CVA6ConfigRVZiCond = 0;
+  localparam CVA6ConfigRVZcheripurecap = 0;
+  localparam CVA6ConfigRVZcherihybrid = 0;
```
JeanRochCoulon, line 32:
```suggestion
```

core/include/build_config_pkg.sv
```diff
@@ -26,20 +27,24 @@ package build_config_pkg;
...
 
     // MMU
-    int unsigned VpnLen = IS_XLEN64 ? (CVA6Cfg.RVH ? 29 : 27) : 20;
-    int unsigned PtLevels = IS_XLEN64 ? 3 : 2;
+    int unsigned VpnLen = (CVA6Cfg.XLEN == 64) ? (CVA6Cfg.RVH ? 29 : 27) : 20;
+    int unsigned PtLevels = (CVA6Cfg.XLEN == 64) ? 3 : 2;
 
     config_pkg::cva6_cfg_t cfg;
 
     cfg.XLEN = CVA6Cfg.XLEN;
```
JeanRochCoulon, line 35: CLEN means Cheri len, also when Cheri is disable, coding style can be improved.
Maybe this would bring confusion, but defining cfg.XLEN as below would remove a lot of line modifications. @PRugg-Cap or @Gchauvon Do you have a better suggestion ?

```suggestion
    cfg.XLEN = CheriPresent ? (CVA6Cfg.XLEN * 2) : CVA6Cfg.XLEN;
```

PRugg-Cap:
Verible is now failing because of the IS_XLEN{32,64} checking script. I did cherry-pick in the commit to fix it, but I'll hold off on pushing it for now because that would pollute the diff. We'll get it when we rebase anyway.

core/include/cv64a6_imafdczcheri_sv39_config_pkg.sv
```diff
@@ -0,0 +1,183 @@
+// Copyright 2021 Thales DIS design services SAS
```
JeanRochCoulon, line 1: Using cv32a60x config package as example is better, it does not define useless localparams

core/include/deprecated_packages/cv64a6_imadfcv_sv39_polara_config_pkg.sv
PRugg-Cap: Is this suggestion because we shouldn't modify deprecated packages, or because we shouldn't have these localparams? If the latter, I can have a go at removing them, but I think we need them statically in some cases.

core/include/build_config_pkg.sv
PRugg-Cap: Hmm, not quite. CLEN is "capability len" and stands for the architectural length of a capability (excluding tag). However, it's actually quite rarely used. We still need XLEN, which is the size of the x registers, e.g. throughout the frontend, and in the ALU. This matches what is written in the RISC-V architecture document. The parameter we use more broadly is REGLEN, which is bigger than CLEN because we store some extra microarchitectural partially decompressed metadata in the registers.

core/include/cv64a6_imafdczcheri_sv39_config_pkg.sv
PRugg-Cap: Okay: I'll give that a go

core/include/ariane_pkg.sv
cainria: This is indeed a limitation of the language (it lacks generic types).

What I recommend in this case:
- If the function is called only in one module (not the case here), move the function into this module. This is what we did for mult functions.
- Otherwise, create a module instead of a function to allow a generic interface. But SystemVerilog does not allow to instantiate modules with call syntax so it might not be handy…

I would still be in favor of the component approach to not break the configuration effort.

core/include/build_config_pkg.sv
PRugg-Cap: I did a quick grep of the repository, filtering out submodules etc. After our changes the counts of uses of the different parameters are as follows:
CLEN: 198
REGLEN: 123
XLEN: 1038
If we changed XLEN to be CLEN, then we'd save those 198, but the 1038 XLENs would all have to be "ADDRLEN" or "ARITHLEN" or something, so it would actually make the diff a lot bigger.

PRugg-Cap: (Though I was surprised we use CLEN more than REGLEN... The parts of the code I was mostly interacting with were more REGLEN heavy)

PRugg-Cap: Though I do take the point that having "CLEN", e.g. on the path from memory into the pipeline, is confusing when CHERI is disabled out. We can consider another option, e.g. something like "MEMLEN" (better name needed!) which is the maximum width of something loaded in from memory in one cycle.

core/include/ariane_pkg.sv
PRugg-Cap: Hmm, okay, these functions were introduced to reduce diff and avoid lots of:
```
if (CheriPresent) begin
  a_reg = {{REGLEN-XLEN}{1'b0}, a};
end else begin
  a_reg = a;
end
```
If we need a module instantiation, then it's probably better to just go back to that.
There are currently 46 calls to these functions throughout the codebase.
I'll contemplate if there's a tidier way. Any chance you could elaborate on how the current approach breaks the configuration effort, so that I can understand the constraints?

cainria: Or maybe

```systemverilog
a_reg = CVA6Cfg.CheriEn ? {{REGLEN-XLEN}{1'b0}, a} : a;
```

To put it short, we would like to be able to instantiate simultaneously 2 ariane cores with Cheri enabled and different XLEN.
Having values in the package instead of the config struct prevents us from doing that.

core/include/deprecated_packages/cv64a6_imadfcv_sv39_polara_config_pkg.sv
JeanRochCoulon: These config packages are not all deprecated, we need to maintain them. The localparams are not used, that is why it is better to remove them

core/include/ariane_pkg.sv
PRugg-Cap: Yeah, we had that form in some cases as well. The zero replicate does seem to generate warning in some cases though. Okay, thanks for the summary of the goal: I'll contemplate how to proceed.

core/include/build_config_pkg.sv
JeanRochCoulon: Thanks for the explanations. I really prefer MEMLEN. I would be please to get @Gchauvon feedback to grant this modification (Guillaume worked on type strategy among the cva6)

PRugg-Cap:
Apologies: I'm still hoping to find time to do the type refactor, but hopefully we can leave this open.

PR #1766, opened 2024-01-16
Adding configuration-specific CSR doc

jquevremont:
First attempt to add configuration-specific CSR lists and descriptions in the user manual.
First attempt to point to a file in another subdirectory (common between user manual and design document). I think we need to PR to test if it creates the expected result in ReadTheDocs.

PR #3385, opened 2026-07-10
Document which medeleg/mideleg bits are actually writable

### LINKED ISSUE #1207: Which exceptions and interrupts can be deligated?

### Bug Description

The `medeleg` and `mideleg` CSRs indicate which exceptions can be delegated to non-machine mode code.  The [User Manual](https://docs.openhwgroup.org/projects/cva6-user-manual/01_cva6_user/CV32A6_Control_Status_Registers.html#machine-exception-delegation-medeleg) indicates that all 32-bits of these CSRs are R/W which implies that all interrupts and exceptions can be delegated.  It seems unlikely that this is correct.

Please update the User Manual to indicate which exceptions and interrupts can be delegated, and the R/W bits of the `medeleg` and `mideleg` CSRs.

vs4848:
The medeleg/mideleg entries in the CVA6 User Manual described all 32 bits as R/W with no further detail, implying every exception and interrupt could be delegated to a lower privilege level. Only a subset of bits are actually writable per the CSR write-mask logic in core/csr_regfile.sv:
- medeleg: the bits corresponding to the exception causes CVA6 actually raises (with H-extension-only bits gated by RVH)
- mideleg: only the S-mode interrupt-pending bits (SSIP/STIP/SEIP)

Add a bit-level table for both registers, matching the write masks in csr_regfile.sv, and note that all remaining bits are hardwired read-only.

Also restore CV32A6_Control_Status_Registers to the User Manual's toctree in index.rst. This generic CSR reference was dropped from the toctree in #1766 when CSR docs were split by configuration (CV32A60X/CV32A60AX/CV64A6), but none of those per-config docs currently document medeleg/mideleg at all (CV32A60X's generated source has no delegation fields; CV32A60AX's page is still a 'not yet available' stub). Without this, the fix below would not appear anywhere in the published manual, per the issue's generic 'User Manual' wording and the notCV32A65X label (this does not touch the CV32A65X embedded doc tree).

Fixes #1207

- [ ] I have searched for similar pull requests
- [ ] I am a human engaging in an interpersonal interaction. During this interaction, my words are my own and are not generated. If relevant, I provide links to my sources.

### DIFF
--- a/docs/01_cva6_user/CV32A6_Control_Status_Registers.rst
+++ b/docs/01_cva6_user/CV32A6_Control_Status_Registers.rst
@@ -802,17 +802,69 @@ Machine Exception Delegation (``medeleg``)
 :Width (bits): 32
 :Access Type: RW
 :Reset Value: 0x00000000
-:Description: Provides individual read/write bits to indicate that certain exceptions should be processed directly by a lower privilege level.
+:Description: Provides individual read/write bits to indicate that certain exceptions should be processed directly by a lower privilege level. Only the bits listed below are writable; all other bits are hardwired to 0 and read-only.
 
 .. csv-table::
    :widths: auto
    :align: left
    :header: "Bits", "Name", "Display Name", "Access Type", "Reset"
 
-   "[31:0]", "``Synchronous_Exceptions``", "Synchronous Exceptions", "RW", "0b0"
+   "[31:24]", "``reserved_0``", "Reserved", "RO", "0b0"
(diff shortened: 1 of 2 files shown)


JeanRochCoulon:
Hello @vs4848   As told in CONTRIBUTING.md, the Eclipse agreement must be signed to become contributor.
I will close the PR, and will be glad to review your new PR when the agreement will be signed.

PR #3386, opened 2026-07-11
csr_regfile: zero mtval/stval/vstval on interrupt traps 

### LINKED ISSUE #3379: [BUG] csr_regfile: mtval zeroing tests ex_i.cause[GPLEN-1] instead of [XLEN-1]; interrupt traps leave instruction bits in mtval/stval
### Code of Conduct

### Bug Description

Two related defects in the `mtval`/`stval` write path leave the interrupted instruction's encoding in the trap-value CSR after an **interrupt** trap, where the privileged spec requires zero. Found by formal property checking of the unmodified `csr_regfile` (Yosys + yosys-slang + SymbiYosys), config `cv64a6_imafdc_sv39`.

This is not the ecall/ebreak case already tracked in #898 and #448 (decoder side partly addressed by PR #3226) those cover exceptions; this is the interrupt path, which none of them touch. #1177 is a separate MMU fetch-fault lint task. A search for `GPLEN` in the tracker returns nothing, so the bit-select in Bug 1 below is unreported.

# Steps to reproduce

**Formal (what found it).** Bind an immediate assertion to `csr_regfile`: after an interrupt trap taken to M mode, `mtval == 0`. It fails in BMC at depth 12. A cover proves the antecedent (interrupt trap to M, nonzero `ex_i.tval`) is reachable, so the CEX is not vacuous.

Image

Checker and trace: https://github.com/codeadpool/cva6-priv-sva/tree/main/evidence/mstatus/mstatus_f5_bmc

**Dynamic (integrated core).** With `TvalEn` set, take any interrupt (e.g. a machine timer interrupt), then in the handler execute `csrr t0, mtval`. The read-back value is the encoding of whatever instruction was interrupted, not zero.

Source pointers (v5.3.0 `2ef1c1b`; master `b30e7db` ):
- `core/csr_regfile.sv:1919` [`:2117`] - the mtval zeroing term.
- `core/include/ariane_pkg.sv:104-109` - `ZERO_TVAL` (default 0, SPIKE_TANDEM-gated).
- `core/decoder.sv:1852` preload of `ex.tval`; interrupt attach `:1948-1980`.

## Expected behavior

The privileged spec (v1.13) requires the trap-value CSR to be zero for interrupts. Machine-Level ISA, `mtval`:

> For other traps, mtval is set to zero, but a future standard may redefine mtval's setting for other traps.

Supervisor-Level ISA, `stval`, carries the identical sentence for `stval`. Interrupts are "other traps" (they appear in none of the exception-specific lists), and the interrupted instruction's encoding is not exception-specific information for an interrupt. So after an interrupt trap, `mtval` (trap to M) / `stval` (trap to S) must read 0.

## Observed behavior

**Bug 1 - wrong bit-select in the M-mode mtval zeroing term.**
`core/csr_regfile.sv:1919` [`:2117`] gates zeroing on `ex_i.cause[CVA6Cfg.GPLEN-1]`:

```systemverilog
mtval_d = (ariane_pkg::ZERO_TVAL
           && (ex_i.cause inside { ... ENV_CALL_MMODE
           } || ex_i.cause[CVA6Cfg.GPLEN-1])) ? '0 : ex_i.tval;
```

Every sibling site tests the actual interrupt flag `ex_i.cause[CVA6Cfg.XLEN-1]`: vstval `:2059`, stval `:2078`, htinst `:2091`, mtinst `:2137`, and the delegation decision `:2020-2022`. With `GPLEN-1` = 40 (RV64) or 33 (RV32), that bit is 0 for every architecturally valid cause value, so the "or this is an interrupt" term is constant-false: the intended zero-mtval-on-interrupt can never fire, even in builds where zeroing is enabled (`SPIKE_TANDEM`, the tandem-verification config). One-line fix: `GPLEN` -> `XLEN`.

**Bug 2 - interrupt traps write the interrupted instruction's encoding into mtval/stval.**
In default builds `ariane_pkg::ZERO_TVAL = 0` (`ariane_pkg.sv:104-109`, set to 1 only under `SPIKE_TANDEM`), so the zeroing arm above is skipped entirely for both the M and S trap paths. The value that flows through instead is not benign:

1. `core/decoder.sv:1852` preloads `instruction_o.ex.tval` with the fetched instruction bits (when `TvalEn`).
2. The interrupt attach (`decoder.sv:1948-1980`) sets only `ex.valid` and `ex.cause`; `ex.tval` keeps the preloaded encoding. PR #3226 fixed the ecall (`:1880`) and ebreak (`:1889`) cases but did not touch the interrupt path.
3. `commit_stage.sv` forwards the exception struct wholesale, and with `ZERO_TVAL=0` `csr_regfile.sv` writes `mtval_d`/`stval_d = ex_i.tval`.

Net effect: after any interrupt trap, `mtval` (trap to M) or `stval` (trap to S) reads back the encoding of the instruction that happened to be interrupted. This is the interrupt analogue of the ecall/ebreak reports #898 / #448; those are being tracked for exceptions, this path is not.

**Suggested fix (happy to open a PR for either or both):**
- `csr_regfile.sv:1919` [`:2117`]: `GPLEN` -> `XLEN` (restores the zeroing backstop for tandem builds).
- Clear `ex.tval` in the decoder's interrupt attach (same shape as the PR #3226 ecall fix), or zero `tval` on `cause[XLEN-1]` unconditionally in `csr_regfile` rather than under `ZERO_TVAL`.

codeadpool:

# Why this change is needed

issue #3379, filed on 6 July against master `b30e7db`, reported two separate reasons an interrupt could leave `ex_i.tval` in the trap-value CSR:

1. The M-mode `mtval` expression tested `ex_i.cause[GPLEN-1]` instead of the interrupt bit at `XLEN-1`.
2. The interrupt test was inside the `ZERO_TVAL` guard, which is disabled in default builds.

An unrelated change in #3418 later changed `GPLEN-1` to `XLEN-1`. That part of #3379 is therefore already corrected on master and is not part of this diff.

This PR addresses the remaining grouping problem at the `mtval`, `stval`, and `vstval` sites. It changes:

```systemverilog
ZERO_TVAL && (listed_exception || interrupt)
```

to:

```systemverilog
(ZERO_TVAL && listed_exception) || interrupt
```

The `ZERO_TVAL`-gated behavior for listed exceptions is unchanged. Interrupt traps now select zero independently of `ZERO_TVAL`. the `mtval` and `stval` behavior was found and checked through formal property checking of `csr_regfile`.

Fixes #3379.

# Verification scope

The archived formal proof covers `mtval` and `stval` with `RVH=0` on pre-rebase PR head `3250ed48`. The identical `vstval` grouping correction is not covered by that proof. The RVH-only `htinst` and `mtinst` paths are unchanged and outside this PR’s scope.

Formal checker, logs, and traces: https://github.com/codeadpool/cva6-priv-sva/tree/main/evidence/mstatus

### DIFF
--- a/core/csr_regfile.sv
+++ b/core/csr_regfile.sv
@@ -2051,12 +2051,12 @@ module csr_regfile
           // set epc
           vsepc_d = {{CVA6Cfg.XLEN - CVA6Cfg.VLEN{pc_i[CVA6Cfg.VLEN-1]}}, pc_i};
           // set vstval
-          vstval_d        = (ariane_pkg::ZERO_TVAL
+          vstval_d        = ((ariane_pkg::ZERO_TVAL
                              && (ex_i.cause inside {
                              riscv::ILLEGAL_INSTR,
                              riscv::BREAKPOINT,
                              riscv::ENV_CALL_UMODE
-                             } || ex_i.cause[CVA6Cfg.XLEN-1])) ? '0 : ex_i.tval;
+                             })) || ex_i.cause[CVA6Cfg.XLEN-1]) ? '0 : ex_i.tval;
         end else begin
           // update sstatus
           mstatus_d.sie = 1'b0;
@@ -2068,14 +2068,14 @@ module csr_regfile
           // set epc
           sepc_d = {{CVA6Cfg.XLEN - CVA6Cfg.VLEN{pc_i[CVA6Cfg.VLEN-1]}}, pc_i};
           // set mtval or stval
-          stval_d        = (ariane_pkg::ZERO_TVAL
+          stval_d        = ((ariane_pkg::ZERO_TVAL
                                   && (ex_i.cause inside {
                                     riscv::ILLEGAL_INSTR,
                                     riscv::BREAKPOINT,
                                     riscv::ENV_CALL_UMODE,
                                     riscv::ENV_CALL_SMODE,
                                     riscv::ENV_CALL_MMODE
-                                  } || ex_i.cause[CVA6Cfg.XLEN-1])) ? '0 : ex_i.tval;
+                                  })) || ex_i.cause[CVA6Cfg.XLEN-1]) ? '0 : ex_i.tval;
           if (CVA6Cfg.RVH) begin
             htinst_d       = (ariane_pkg::ZERO_TVAL
                               && (ex_i.cause inside {
@@ -2107,14 +2107,14 @@ module csr_regfile
         mepc_d = {{CVA6Cfg.XLEN - CVA6Cfg.VLEN{pc_i[CVA6Cfg.VLEN-1]}}, pc_i};
         // set mtval or stval
         if (CVA6Cfg.TvalEn) begin
-          mtval_d        = (ariane_pkg::ZERO_TVAL
+          mtval_d        = ((ariane_pkg::ZERO_TVAL
                                     && (ex_i.cause inside {
                                       riscv::ILLEGAL_INSTR,
                                       riscv::BREAKPOINT,
                                       riscv::ENV_CALL_UMODE,
                                       riscv::ENV_CALL_SMODE,
                                       riscv::ENV_CALL_MMODE
-                                    } || ex_i.cause[CVA6Cfg.GPLEN-1])) ? '0 : ex_i.tval;
+                                    })) || ex_i.cause[CVA6Cfg.XLEN-1]) ? '0 : ex_i.tval;
         end else begin
           mtval_d = '0;
         end
--- a/core/csr_regfile.sv
+++ b/core/csr_regfile.sv
@@ -2127,12 +2127,12 @@ module csr_regfile
           // set epc
           vsepc_d = {{CVA6Cfg.XLEN - CVA6Cfg.VLEN{pc_i[CVA6Cfg.VLEN-1]}}, pc_i};
           // set vstval
-          vstval_d        = (ariane_pkg::ZERO_TVAL
+          vstval_d        = ((ariane_pkg::ZERO_TVAL
                              && (ex_i.cause inside {
                              riscv::ILLEGAL_INSTR,
                              riscv::BREAKPOINT,
                              riscv::ENV_CALL_UMODE
-                             } || ex_i.cause[CVA6Cfg.XLEN-1])) ? '0 : ex_i.tval;
+                             })) || ex_i.cause[CVA6Cfg.XLEN-1]) ? '0 : ex_i.tval;
         end else begin
           // update sstatus
           mstatus_d.sie = 1'b0;
@@ -2144,14 +2144,14 @@ module csr_regfile
           // set epc
           sepc_d = {{CVA6Cfg.XLEN - CVA6Cfg.VLEN{pc_i[CVA6Cfg.VLEN-1]}}, pc_i};
           // set mtval or stval
-          stval_d        = (ariane_pkg::ZERO_TVAL
+          stval_d        = ((ariane_pkg::ZERO_TVAL
                                   && (ex_i.cause inside {
                                     riscv::ILLEGAL_INSTR,
                                     riscv::BREAKPOINT,
                                     riscv::ENV_CALL_UMODE,
                                     riscv::ENV_CALL_SMODE,
                                     riscv::ENV_CALL_MMODE
-                                  } || ex_i.cause[CVA6Cfg.XLEN-1])) ? '0 : ex_i.tval;
+                                  })) || ex_i.cause[CVA6Cfg.XLEN-1]) ? '0 : ex_i.tval;
           if (CVA6Cfg.RVH) begin
             htinst_d       = (ariane_pkg::ZERO_TVAL
                               && (ex_i.cause inside {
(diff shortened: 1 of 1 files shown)


JeanRochCoulon:
By reviewing this PR, we can wonder whether code could be simplified.
ZERO_TVAL seems to be defined to support Spike Tandem. @AyoubJalali do you confirm we need this trick ?

Extracted from ariane_pkg.sv:
image

@ASintzoff

AyoubJalali:
> By reviewing this PR, we can wonder whether code could be simplified. ZERO_TVAL seems to be defined to support Spike Tandem. @AyoubJalali do you confirm we need this trick ?
> 
> Extracted from ariane_pkg.sv: image
> 
> @ASintzoff

LGTM

codeadpool:
@JeanRochCoulon this pr leaves ZERO_TVAL doing exactly what it did before. the only change is the interrupt check now sits outside the guard. with ZERO_TVAL=0 that reduces to just cause[XLEN-1], so interrupts zero tval in every build. that's the part spec requires, tandem or not. the exception cause list is still gated exactly as before, so if you decide the trick isn't necessary, dropping it only affects that list and interrupts stay correct.
on simplification: mtval/stval/vstval all have the same shape now, with interrupt case pulled out of the tandem gated one. can restructure if you'd prefer.

ran this branch against unmodified csr_regfile: 
| v5.3.0 (2ef1c1b) | this pr (3250ed48) |
|------|--------|
| CEX | Pass |

logs and traces: https://github.com/codeadpool/cva6-priv-sva/tree/main/evidence/mstatus

limitation: htinst/mtinst have their interrupt condition inside the ZERO_TVAL guard, so on a RVH build an interrupt can still leave them non-zero. that's outside the config i verified (RVH=0), so i kept it out of this PR, happy to open a separate issue for it.

codeadpool:
the previous table was bmc only (depth 12), re-ran with an unbounded proof, and this pr changes three sites and i'd only checked mtval.

| property                          | v5.3.0 (2ef1c1b) | this pr (3250ed48)   |
|-----------------------------------|------------------|----------------------|
| interrupt trap to M => mtval == 0 | CEX              | proven (k-induction) |
| interrupt trap to S => stval == 0 | CEX              | proven (k-induction) |

the antecedent covers (M/S interrupt with nonzero tval) still reach on the fixed rtl, and both witness covers are the exact negation of a proven assertion, so they're unreachable by proof rather than by bounded search.

still uncovered: vstval (needs RVH=1, outside the config i verify) plus the htinst/mtinst limitation noted above.

logs and traces: https://github.com/codeadpool/cva6-priv-sva/tree/main/evidence/mstatus

cainria:
Hi! Any news about this PR ?

codeadpool:
Hi im waiting for feedback from the code owner to see if any adjustments are needed on my side. I’ll update the PR accordingly once I hear back

JeanRochCoulon:
Could you resolve conflicts and rebase ?

codeadpool:
rebased onto master. master has since changed ex_i.cause[GPLEN-1] to [XLEN-1] at all three sites, so the bit-select half of this PR is already fixed there. The remaining change is the grouping.

I also removed the //TODO ... i fail to understand why comment on the mtval line: #3379 explains why that bit-select was wrong: GPLEN-1 is bit 40, which is 0 for every valid cause, so the interrupt term was constant false.

JeanRochCoulon:
Hello @codeadpool 
As we can read into privilege specification, the tval is implementation specific.

image


In CVA6, we introduced ZERO_TVAL parameter to make tval csr read-only zero. A PR which changes this assumption is not wanted.

Out of H extension, I would think this PR is not needed. When H extension is enable, do you suggest to constraint the ZERO_TVAL to be disable ?

codeadpool:
thanks, I think `ZERO_TVAL` and `TvalEn` are being conflated.

This PR changes only `Z && (E || I)` to `(Z && E) || I`. With `Z=1`, they are identical; with `Z=0`, only the interrupt case changes. The existing listed-exception and Spike-tandem behavior is unchanged.

`ZERO_TVAL` adjusts selected trap-entry values but does not gate CSR reads or software writes, so it does not make the CSRs read-only zero. For `mtval` and `stval`, `CVA6Cfg.TvalEn` gates their state updates directly.

the implementation choice in the paragraph you shared is scoped to exceptions , which exceptions provide information and which may return zero. interrupts are covered by a separate requirement: “For other traps, `mtval` is set to zero,” with the same rule for `stval` ([machine](https://docs.riscv.org/reference/isa/v20260120/priv/machine.html#norm:mtval_other_traps_zero), [supervisor](https://docs.riscv.org/reference/isa/v20260120/priv/supervisor.html#norm:stval_op_other_traps)).

[machine]: https://docs.riscv.org/reference/isa/v20260120/priv/machine.html#norm:mtval_other_traps_zero
[supervisor]: https://docs.riscv.org/reference/isa/v20260120/priv/supervisor.html#norm:stval_op_other_traps

No `ZERO_TVAL` constraint is needed for H. `vstval` substitutes for `stval` when V=1, so I kept the same correction there. My formal check covers `RVH=0`, however, so I can split that hunk if you want this PR restricted to the verified M/S scope.

JeanRochCoulon:
Ok, thanks for the information. I am ok with the PR. I will launch the CI and merge asap green.

Issue #3313, opened 2026-05-09
[BUG] MRET incorrectly sets V=1 when MPP=M-mode


oChunCai:
### Code of Conduct

- [x] I have searched the existing bug issues.
- [x] I am a human engaging in an interpersonal interaction. During this interaction, my words are my own and are not generated. If relevant, I provide links to my sources.

### Bug Description

When the hypervisor extension (RVH) is enabled, the CVA6 MRET execution path unconditionally assigns the virtualization mode `V` to `mstatus.MPV` without handling the special case where `MPP = M-mode` requires `V` to be forced to 0.

As RISC-V Privileged ISA:
> "When an MRET instruction is executed, the virtualization mode V is set to MPV, unless MPP=3, in which case V remains 0."

### Vulnerable Code and Behavior

`core/csr_regfile.sv` (lines 2323–2329) — missing `MPP == M` special case:

```systemverilog
if (CVA6Cfg.RVH) begin
  // set virtualization mode
  v_d           = mstatus_q.mpv;            // <-- unconditional, MPP not checked
  //set mstatus mpv to false
  mstatus_d.mpv = 1'b0;
  if (mstatus_q.mpp != riscv::PRIV_LVL_M) mstatus_d.mprv = 1'b0;
end
```

124107157-KV:
Hi, I would like to work on this issue.

I reviewed the current `MRET` handling in `core/csr_regfile.sv` and confirmed that, when `CVA6Cfg.RVH` is enabled, `v_d` is restored directly from `mstatus_q.mpv` without checking whether `mstatus_q.mpp` is Machine mode.

I plan to:

* update the `MRET` logic so that `V` is forced to `0` when `MPP == M`;
* preserve the existing behaviour where `V` is restored from `MPV` when returning to a lower privilege mode;
* add a directed regression covering both cases:

  * `MPP=M, MPV=1` → `V=0`
  * `MPP=S, MPV=1` → `V=1`
* run the relevant CSR and hypervisor regression tests.

Could you please assign this issue to me? Also, please let me know if there is a preferred location or existing test framework for adding this regression.

124107157-KV:
Thanks for reporting this issue. I have opened PR #3438 with a fix.

The change forces `V=0` when `MRET` returns to M-mode with `MPP=M`, while preserving the existing `MPV` behavior for lower privilege modes.

[...]

Issue #1985, opened 2024-04-01
Debug-mode entry reason dcsr field cause is writable by software


salaheddinhetalani:
### Is there an existing CVA6 bug for this?

- [X] I have searched the existing bug issues

### Bug Description

Debug mode entry reason field, `cause`, of `dcsr` should not be writable by software. Same applies to the `debugver` field.

### RISC-V Specification
Unprivileged ISA: [20191213](https://drive.google.com/file/d/1s0lZxUZaa7eV_O0_WsZzaurFLLww7ou5/view) | Privileged Architecture: [20211203](https://drive.google.com/file/d/1EMip5dZlnypTk7pt4WWUKmtjUKTOkBqh/view) | External Debug Support: [0.13.02](https://drive.google.com/file/d/1XSPqW3WIfflIMeuGJBxGTMOEyeuRAQ5F/view)
> Upon entry into Debug Mode, v and prv are updated with the privilege level the hart was previously
in, and cause is updated with the reason for Debug Mode entry. Other than these fields and nmip,
the other fields of dcsr are only writable by the external debugger.

> ![image](https://github.com/openhwgroup/cva6/assets/28307467/9dec9996-0b0a-4ba6-bd25-f36bff097875)

> ![image](https://github.com/openhwgroup/cva6/assets/28307467/9651c89f-a4d1-48fb-8267-c4210651bee6)

### Example Scenario

As shown below, the following sequence of instructions happens: 
> CSRRW x0, dcsr, x3

The instruction `CSRRW x0, dcsr, x3` is decoded at `t_id` and executed in debug mode resulting in updating the architecture state at `t_arch_update`. As a result, `cause` field of `dcsr` is updated, but it shouldn't as it's not writable by software.

![Issue_9](https://github.com/openhwgroup/cva6/assets/28307467/ec0176d7-058f-448b-84f4-f925024eddc0)

### Steps to Reproduce

Git Hash: [de2e254c](https://github.com/openhwgroup/cva6/commit/de2e254cd4c2d3e3a6d4def1652918e8f2535512) | TARGET_CFG: [cv32a6_imac_sv0](https://github.com/openhwgroup/cva6/blob/master/core/include/cv32a6_imac_sv0_config_pkg.sv) | VCD: [issue_9.zip](https://github.com/openhwgroup/cva6/files/14826407/issue_9.zip)

### Component
**Component:RTL**

---
Product: Questa OneSpin Solutions
App: Questa Processor App
Tool's version: 2024.1_1

Issue #1984, opened 2024-04-01
Virtualization mode only dcsr field v and other reserved fields are set


salaheddinhetalani:
### Is there an existing CVA6 bug for this?

- [X] I have searched the existing bug issues

### Bug Description

Virtualization mode only field, `v`, of `dcsr` is set to non zero value. Same case, not shown here, for always zero reserved fields, bits `[27:18]` and bit `[14]`.

### RISC-V Specification
Unprivileged ISA: [20191213](https://drive.google.com/file/d/1s0lZxUZaa7eV_O0_WsZzaurFLLww7ou5/view) | Privileged Architecture: [20211203](https://drive.google.com/file/d/1EMip5dZlnypTk7pt4WWUKmtjUKTOkBqh/view) | External Debug Support: [0.13.02](https://drive.google.com/file/d/1XSPqW3WIfflIMeuGJBxGTMOEyeuRAQ5F/view)
> Upon entry into Debug Mode, v and prv are updated with the privilege level the hart was previously
in, and cause is updated with the reason for Debug Mode entry. Other than these fields and nmip,
the other fields of dcsr are only writable by the external debugger.

> ![image](https://github.com/openhwgroup/cva6/assets/28307467/9dec9996-0b0a-4ba6-bd25-f36bff097875)

> ![image](https://github.com/openhwgroup/cva6/assets/28307467/9c03be99-4b92-44b2-ae9b-f28a3868700f)

### Example Scenario

As shown below, the following sequence of instructions happens: 
> CSRRS x1, dcsr, x3

The instruction `CSRRS x1, dcsr, x3` is decoded at `t_id` and executed in debug mode resulting in updating the architecture state at `t_arch_update`. As a result, `v` field is set, but it should be hardwired to 0 on harts that do not support virtualization mode.

![Issue_8](https://github.com/openhwgroup/cva6/assets/28307467/3c708cc9-0e19-42ba-b54f-04d92c11f130)

### Steps to Reproduce

Git Hash: [de2e254c](https://github.com/openhwgroup/cva6/commit/de2e254cd4c2d3e3a6d4def1652918e8f2535512) | TARGET_CFG: [cv32a6_imac_sv0](https://github.com/openhwgroup/cva6/blob/master/core/include/cv32a6_imac_sv0_config_pkg.sv) | VCD: [issue_8.zip](https://github.com/openhwgroup/cva6/files/14826405/issue_8.zip)

### Component
**Component:RTL**

---
[...]

PR #3387, opened 2026-07-12
csr_regfile: WARL legalize the dcsr write path (prv, v, cause)

### LINKED ISSUE #3383: [BUG] csr_regfile: dcsr.prv is not WARL-legalized on write; dret can set priv_lvl to an unimplemented encoding (2'b10)
### Code of Conduct

### Bug Description

The DCSR write path stores dcsr.prv without legalizing it, the sibling fields right beside it are legalized, prv is not so software can write any 2-bit value. dret then copies dcsr.prv straight into priv_lvl. On a build without the hypervisor extension (RVH=0, the default cv64a6_imafdc_sv39), this lets a debugger place the hart at priv_lvl = 2'b10, an encoding the configuration does not implement. Found by formal property checking of the unmodified csr_regfile (Yosys + yosys-slang + SymbiYosys).

A note on 2'b10: riscv_pkg labels it PRIV_LVL_HS (core/include/riscv_pkg.sv), used as the access level for hypervisor CSRs. It is not a distinct operating privilege level in RISC-V (HS is S-mode with V=0, priv = 2'b01), and with RVH=0 the hypervisor extension is absent, so no implemented mode uses 2'b10.

Adjacent issues #1984 (dcsr.v and reserved bits writable) and #1985 (dcsr.cause writable) report the same missing legalization for other dcsr
fields; neither covers prv or the dret -> priv_lvl path.

# Steps to reproduce

Formal (what found it). Bind an immediate assertion to csr_regfile:
priv_lvl is always M, S, or U. It fails in BMC. The counterexample enters debug mode, writes dcsr with prv = 2'b10, executes dret, and priv_lvl_q reads back 2'b10 the next cycle. A cover proves the write is reachable, so the CEX is not vacuous.

Image

Checker and trace: https://github.com/codeadpool/cva6-priv-sva/tree/main/evidence/probe/probe_priv_dret_bmc

Dynamic (integrated core). In a debug session on an RVH=0 build: enter debug mode, csrw dcsr with bits [1:0] = 2'b10, dret, then observe the privilege level. priv_lvl is 2'b10, which is none of M, S, or U.

Source pointers (upstream master on 2026-07-06):

core/csr_regfile.sv:1162-1173 - the CSR_DCSR write case (prv raw at :1164).
core/csr_regfile.sv:2369-2373 - the dret block (priv_lvl_d at :2373).
core/include/riscv_pkg.sv:846-880 - dcsr_t (prv at :879); priv_lvl_t
encodings at :31-36.

## Expected behavior

The RISC-V Debug spec (v1.0, dcsr) defines prv as a WARL field over the privilege modes the hart supports, so an unsupported value must not be retained. Consequently priv_lvl must only hold an implemented mode on this RVH=0 build, M (2'b11), S (2'b01), or U (2'b00).

## Observed behavior

The DCSR write stores prv raw (core/csr_regfile.sv:1162-1173):

riscv::CSR_DCSR: begin
  if (CVA6Cfg.DebugEn) begin
    dcsr_d           = csr_wdata[31:0];   // whole word, incl prv
    dcsr_d.xdebugver = 4'h4;              // these are legalized
    dcsr_d.nmip      = 1'b0;
    dcsr_d.stopcount = 1'b0;
    dcsr_d.stoptime  = 1'b0;
    // prv is never re-touched -> keeps csr_wdata[1:0]
  end ...

dret loads it into the live privilege register (:2369-2373):

if (dret) begin
  eret_o     = 1'b1;
  priv_lvl_d = riscv::priv_lvl_t'(dcsr_q.prv);   // no clamp, no legalize

mret/sret legalize their target privilege (mpp/spp via WARL); dret does not. So after dret, priv_lvl_q can hold 2'b10, which on an RVH=0 hart
is not an implemented privilege mode. The build's privilege logic compares only against M/S/U - trap delegation (csr_regfile.sv:2027, priv == PRIV_LVL_M ? M : S) and the PMP M-mode bypass (core/pmp/src/pmp.sv:57,67, priv_lvl_i == PRIV_LVL_M) so the hart runs at a privilege the configuration does not define: not M (it loses the M-mode PMP bypass and traps route as if below M), and equal to neither S nor U.

Suggested fix (happy to open a PR). Legalize prv on the DCSR write,
mirroring the fields already legalized above:

if (!(dcsr_d.prv == riscv::PRIV_LVL_M
      || (CVA6Cfg.RVS && dcsr_d.prv == riscv::PRIV_LVL_S)
      || (CVA6Cfg.RVU && dcsr_d.prv == riscv::PRIV_LVL_U)))
  dcsr_d.prv = riscv::PRIV_LVL_M;   // or the highest supported mode

Equivalently, clamp at the dret load (:2373). The write-path fix is cleaner and matches the existing legalization style, and is the same class of fix as #1984 / #1985 for the other dcsr fields.

codeadpool:

DCSR write path stored several fields raw, this legalizes them on write.  prv is clamped to supported privilege modes, v is forced to 0 without hypervisor extension and at RVH=1, when the mode being stored or restored is M: on the write when the written prv is M, and on dret when the restored privilege is M (M-mode is never virtualized), and cause is preserved. Without prv legalization a dret could set priv_lvl to an unimplemented encoding (2'b10/ PRIV_LVL_HS on an RVH=0 build), losing M-mode's PMP bypass. Found by formal property checking of csr_regfile. The property that priv_lvl is always one of M, S, U fails on current RTL and passes after the fix, the illegal dcsr.prv state that produced the cex becomes unreachable. 

Fixes #3383, #1984, #1985 

Verified formally on cv64a6_imafdc_sv39 (RVH=0) for prv/cause/reserved, and on cv64a6_imafdch_sv39 (RVH=1) for the dcsr.v clamps. the legalization is guarded by the same RVH/RVS/RVU params the rest of the CSR file uses. The mret path to the same M-with-V=1 state is a separate register, tracked in #3313.


core/csr_regfile.sv
```diff
@@ -1168,6 +1168,16 @@ module csr_regfile
             dcsr_d.nmip      = 1'b0;
             dcsr_d.stopcount = 1'b0;
             dcsr_d.stoptime  = 1'b0;
+            // dcsr.prv is WARL over supported privilege modes
+            if ((!CVA6Cfg.RVH && dcsr_d.prv == riscv::PRIV_LVL_HS) ||
```
JeanRochCoulon: When !CVA6Cfg.RVH, we cannot have dcsr_d.prv == riscv::PRIV_LVL_HS...

codeadpool: dcsr.prv can hold 2'b10 here, write is raw (dcsr_d = csr_wdata[31:0], dcsr_t.prv is a full 2-bit priv_lvl_t), so a debugger can set it and dret copies it into priv_lvl unlegalized. The formal check behind this PR witnesses this on RVH=0 (fails before the patch, passes after).
 
Agreed, CVA6Cfg.RVH guard is too narrow though. 2'b10 is never a valid operating priv_lvl in any config: HS mode is priv_lvl=S with V=0 not priv_lvl=2'b10, so the guard shouldn't be conditional. I'd mirrored mstatus.mpp handling at csr_regfile.sv:1382, it has the same !RVH form. i've generalized it to legalize prv against the supported set unconditionally and clamp anything else to M, so it also covers RVH=1 builds:

```systemverilog
   if (dcsr_d.prv != riscv::PRIV_LVL_M &&
       !(CVA6Cfg.RVS && dcsr_d.prv == riscv::PRIV_LVL_S) &&
       !(CVA6Cfg.RVU && dcsr_d.prv == riscv::PRIV_LVL_U)) begin
     dcsr_d.prv = riscv::PRIV_LVL_M;
   end 
```
   

Pushed. My formal setup only exercises RVH=0, but this generalization is correct by construction for the hypervisor configs too.

codeadpool:
Pushed 8f74af4, which zeroes DCSR reserved fields zero 1 (bit 14) and zero2 (bits 27:18). they were assigned raw from csr_wdata, so software could set them. this completes the reserved-field half of #1984 that the prv/v/cause change didn't cover.

| property                              | v5.3.0 (2ef1c1b) | this pr (8f74af4a) |
 |---|---|---|
| priv_lvl in {M,S,U} after dret (#3383) | CEX              | pass               |
| DCSR reserved fields read 0 (#1984)   | CEX              | pass               |

In both cases the scenario covers (debug entry, dret in debug) still reach and only the defect witness covers become unreachable so defects are unreachable, not unexercised. 
Logs and traces: https://github.com/codeadpool/cva6-priv-sva/tree/main/evidence/probe

codeadpool:
I ran formal on this PR at head 58e79cff. Each property produces a CEX on the unpatched v5.3.0 baseline and passes here as an unbounded proof.

At cv64a6_imafdc_sv39 (RVH=0):

- dcsr.prv: proven WARL-legal, with priv_lvl after dret constrained to {M, S, U}.
- DCSR reserved fields (#1984): zero1 (bit 14) and zero2 (bits 27:18) proven to read zero.
- dcsr.cause (#1985): proven preserved across an in-debug CSR write, so a debugger cannot overwrite the hardware-set cause.

At cv64a6_imafdch_sv39 (RVH=1):

- dcsr.v write legalization: with v cleared when the written, legalized prv is M, proven that an in-debug DCSR write cannot store {prv=M, v=1}.
- DRET legalization: with v cleared when DRET restores M-mode, proven that DRET never resumes M-mode with V=1, regardless of how DCSR acquired its value.

Each property fails without its corresponding clamp and proves with it.

Checkers and per property logs: https://github.com/codeadpool/cva6-priv-sva/tree/main/evidence/probe

One related path remains: M,V=1 is also reachable through mret when mstatus.mpv=1 and mstatus.mpp=M. that is open issue #3313, whose fix belongs in the mstatus/MRET path. This PR handles the DCSR/DRET paths; #3313 handles MRET

cainria:
Ping @JeanRochCoulon

codeadpool:
still active from my side, sorry for the quite stretch i was out. 
the formal above is against head 8f74af4a and still holds. Happy to rebase or re-run against a newer base whenever its useful. 
closing this would leave open: the dcsr.prv half fixes #3383, and the reserved-field half completes #1984. Both are still unfixed on master.
 
 @JeanRochCoulon no rush, just flagging that it's not abandoned.

core/csr_regfile.sv
```diff
@@ -1229,6 +1229,20 @@ module csr_regfile
             dcsr_d.nmip      = 1'b0;
             dcsr_d.stopcount = 1'b0;
             dcsr_d.stoptime  = 1'b0;
+            // dcsr.prv is WARL over supported privilege modes
+            // legalize any unsupported value to M in every configuration
+            if (dcsr_d.prv != riscv::PRIV_LVL_M &&
+                !(CVA6Cfg.RVS && dcsr_d.prv == riscv::PRIV_LVL_S)  &&
+                !(CVA6Cfg.RVU && dcsr_d.prv == riscv::PRIV_LVL_U)) begin
+              dcsr_d.prv = riscv::PRIV_LVL_M;
```
JeanRochCoulon: Hi @codeadpool 
As dcsr.prv is WARL, the CSR value must be kept when the written value is not legal.
For instance: if the previous value is S, when writing an illegal value into this CSR, the value must remain S. In your case, the CSR takes the value M, which is not correct to me.

codeadpool: changed to preserve the previous dcsr.prv instead of remapping it to M. the dcsr.v clamp still runs after it, so it sees the retained value. 

adding a check that an unsupported write leaves prv unchanged: the legality assertion hold either way and can't distinguish them, will rerun and update the evidence.

Issue #3368, opened 2026-06-28
[BUG] LSU non-idempotent check ignores page offset and allows wrong-path speculative load to leave D-cache footprint


YzhDDDing:
### Code of Conduct

- [x] I have searched the existing bug issues.
- [x] I am a human engaging in an interpersonal interaction. During this interaction, my words are my own and are not generated. If relevant, I provide links to my sources.

### Bug Description

In a CVA6 configuration with `SpeculativeSb` and non-idempotent region checking enabled, the load unit can incorrectly allow a wrong-path speculative load whose full physical address is inside a sub-page non-idempotent region.

The likely root cause is in `core/load_unit.sv`. The `paddr_ni` predicate is computed from the translated page base instead of the full physical address:

```systemverilog
assign paddr_ni = config_pkg::is_inside_nonidempotent_regions(
    CVA6Cfg, {{52 - CVA6Cfg.PPNW{1'b0}}, dtlb_ppn_i, 12'd0}
);
```

That drops the 12-bit page offset. The speculative-load gate then uses this value:

```systemverilog
if (!CVA6Cfg.SpeculativeSb || !lsu_ctrl_i.is_speculative_load || (dtlb_hit_i && !paddr_ni)) begin
  ...
  req_port_o.data_req = 1'b1;
end
```

For a non-idempotent range such as `[0x80002080, 0x80002090)`, a load to `0x80002080` is checked as `0x80002000`. If the page base is not itself inside the non-idempotent range, `paddr_ni` is false and the LSU can issue the speculative load before the branch is resolved.

This violates the intent of the speculative non-idempotent guard: wrong-path loads to non-idempotent addresses should not be sent to the memory system.

I also observed a D-cache side effect when the target address is intentionally kept inside the cacheable DRAM region. In `core/cache_subsystem/wt_dcache_ctrl.sv`, the cacheable/non-cacheable decision is based on `is_inside_cacheable_regions`, not on the LSU's non-idempotent predicate, so this wrong-path request can fill the D-cache if the address is cacheable.

### Step to Reproduce

Target used locally:

```text
target=cv64a6_imafdc_sv39
local checkout based on commit eb6ff337b5477aa352b714ebcce184d64dada2df
```

The attached PoC configures:

```text
Non-idempotent range: [0x80002080, 0x80002090)
Cacheable DRAM range: [0x80000000, 0xc0000000)
```

[...]

PR #3388, opened 2026-07-13
Fix load unit non-idempotence check missing page offset

LorenzoR01:
Non-idempotence checks for speculative loads are based on `paddr_ni` which does not include the page offset, this can allow wrong-path speculative loads whose full physical address is inside a sub-page non-idempotent region.
To fix the issue this PR adds the page offset from `vaddr[11:0]` to the `paddr_ni` computation.

* **Fixes** https://github.com/openhwgroup/cva6/issues/3368

### DIFF
--- a/core/load_unit.sv
+++ b/core/load_unit.sv
@@ -233,7 +233,7 @@ module load_unit
   logic inflight_stores;
   logic stall_ni;
   assign paddr_ni = config_pkg::is_inside_nonidempotent_regions(
-      CVA6Cfg, {{52 - CVA6Cfg.PPNW{1'b0}}, dtlb_ppn_i, 12'd0}
+      CVA6Cfg, {{52 - CVA6Cfg.PPNW{1'b0}}, dtlb_ppn_i, lsu_ctrl_i.vaddr[11:0]}
   );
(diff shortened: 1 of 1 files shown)


JeanRochCoulon:
I have a question: if we assume that idempotency regions are of minimum 4kBytes granularity, this offset PR would be not needed ?  We must check what the RISC-V specification says

LorenzoR01:
> I have a question: if we assume that idempotency regions are of minimum 4kBytes granularity, this offset PR would be not needed ? We must check what the RISC-V specification says

[The RISC-V specification for Idempotency PMAs](https://docs.riscv.org/reference/isa/v20260120/priv/machine.html#3-1-6-7-idempotency-pmas) says: "The size of these naturally aligned power-of-2 regions is implementation-defined, but, for systems with page-based virtual memory, must not exceed the smallest supported page size." 
So it looks like we can't have non-idempotent regions smaller than 4KBytes and thus this PR is not necessary.

cainria:
> must not exceed the smallest supported page size

IIUC the implementation-defined region size *can* be smaller than the page size, but it cannot cross the page boundary

JeanRochCoulon:
Thanks for this double check. I think this PR can be closed.

PR #3403, opened 2026-07-22
update GitHub's actions

cainria:
In #3233 I used a too restrictive command to update actions so it missed the deprecated actions for the Verible workflow, which now fails.

This time I used the following command:

```
ls .github/**/*.yml | get name | each {
  sed -i s=actions/cache@v.=actions/cache@v6= $in
  sed -i s=actions/checkout@v.=actions/checkout@v7= $in
}
```

### DIFF
--- a/.github/actions/setup-cva6-env/action.yml
+++ b/.github/actions/setup-cva6-env/action.yml
@@ -38,7 +38,7 @@ runs:
     # Step 2: Cache RISC-V Toolchain
     - name: Cache RISC-V Toolchain
       id: cache-toolchain
-      uses: actions/cache@v4
+      uses: actions/cache@v6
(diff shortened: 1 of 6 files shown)


cainria:
Verible runs `on: pull_request_target` so **it uses the version of the workflow in `master`, not the one of the PR**, hence the failure here.

JeanRochCoulon:
Hello @cainria   I see that the GitHub CI is red. Thi is not the expected result. Do I understand your fix is not fixing the issue ?

cainria:
Hi @JeanRochCoulon In my last comment I meant that the workflow used is the one currently in `master`, not the one from my PR. It should pass for everyone once merged.

PR #2035, opened 2024-04-12
Fix reserved value in MPP

Moschn:
Fix CSR behavior of MPP field in MSTATUS and SSTATUS. If the hypervisor extension is not enabled, MPP value 2'b10 must be reserved.

Fixes #1988

PR #3414, opened 2026-07-24
csr_regfile: legalize mstatus.MPP against reserved 2'b10 when RVH=1

### LINKED ISSUE #3411: [BUG] csr_regfile: mstatus.MPP accepts reserved 2'b10 when RVH=1
### Code of Conduct

### Bug Description

Unless otherwise stated, RTL line references below refer to upstream master `a3dc2c5e`.

`mstatus.MPP` rejects the reserved value `2'b10` only when the hypervisor extension is disabled. The check at `core/csr_regfile.sv:1552` is gated by `!CVA6Cfg.RVH`:

```systemverilog
if ((!CVA6Cfg.RVH & mstatus_d.mpp == riscv::PRIV_LVL_HS) |
    (!CVA6Cfg.RVS & mstatus_d.mpp == riscv::PRIV_LVL_S) |
    (!CVA6Cfg.RVU & mstatus_d.mpp == riscv::PRIV_LVL_U)) begin
  mstatus_d.mpp = mstatus_q.mpp;
end
```

The RISC-V Privileged architecture specifies that the xPP fields are WARL and can hold only privilege mode x and implemented lower privilege modes ([machine.adoc:433-434](https://github.com/riscv/riscv-isa-manual/blob/971bc71be1429c97a6d3cc0c66a2bb2e6d5472ff/src/priv/machine.adoc#L433-L434)). 2'b10 is not an architectural current-privilege encoding and is therefore not a legal value for mstatus.MPP.

The hypervisor extension does not add a fourth nominal privilege encoding. HS-mode uses S privilege (2'b01) with V=0 ([hypervisor.adoc:55-59](https://github.com/riscv/riscv-isa-manual/blob/971bc71be1429c97a6d3cc0c66a2bb2e6d5472ff/src/priv/hypervisor.adoc#L55-L59), [hypervisor.adoc:98-99](https://github.com/riscv/riscv-isa-manual/blob/971bc71be1429c97a6d3cc0c66a2bb2e6d5472ff/src/priv/hypervisor.adoc#L98-L99)).

CVA6 calls `2'b10` `PRIV_LVL_HS` because it uses that value internally for CSR access checks. That internal use does not make `2'b10` legal in `mstatus.MPP`.

Therefore, enabling H should not disable the `MPP=2'b10` check.

This is the `RVH=1` case missed by #1988 and #2274. PR #2035 added the `2'b10` check only for `RVH=0`, and PR #2285 kept that condition while adding checks for unsupported S and U modes.

This is also separate from #3294. That issue concerns `mret` and `sret` failing to clear `MPRV` when `RVH=0`. This report concerns an `RVH=1` CSR write storing an illegal `MPP` value before any return instruction.

Writing `mstatus` requires M-mode, so this is **not a privilege escalation**. It is a WARL correctness bug that can place the hart in an undefined privilege encoding.

# Steps to reproduce

## Dynamic

On an `RVH=1` build such as `cv64a6_imafdch_sv39`, execute this in M-mode:

```asm
csrr t0, mstatus
li   t1, ~(3 << 11)
and  t0, t0, t1
li   t1, (2 << 11)      # MPP = 2'b10
or   t0, t0, t1
csrw mstatus, t0
csrr t2, mstatus        # t2[12:11] reads back as 2'b10
```

On an `RVH=0` build, the same write retains the previous `MPP` value because the guard fires.

## Formal

An assertion checks that, outside reset, `mstatus.MPP` is always one of `{M,S,U}`.

on the pinned CVA6 v5.3.0 submodule (`2ef1c1b`), the `RVH=1` check produces a three-step CEX. During the write interval:

```text
csr_we               = 1
csr_op_i              = 8'h23 (CSR_SET)
csr_addr_i            = 12'h300 (mstatus)
csr_wdata[12:11]      = 2'b10
debug_mode_q          = 0
priv_lvl_q            = 2'b11 (M)
```

At the next edge, `mstatus_q.mpp` becomes `2'b10`, while `priv_lvl_q` remains M. The failure is therefore a direct, non-debug write to `mstatus`; it does not use the #3383 `dcsr.prv`/`dret` path.

The faulty guard is unchanged on current master `a3dc2c5e`.

The unpatched `RVH=0` check can also fail through the separate #3383 path. to remove that unrelated path, the same property was run on PR #3387 head `58e79cff`, which legalizes `dcsr.prv`:

* `RVH=0`: proved unbounded with PDR.
* `RVH=1`: counterexample at step 3.

The two stock configurations also differ in `ZKN` and `SvnapotEn`, but `csr_regfile` references neither field. `RVS` and `RVU` are the same in both configs.

Image

**Counterexample:** During the 10-20 ns interval, a `CSR_SET` operation to `mstatus` writes `MPP=2'b10` while `debug_mode_q=0` and `priv_lvl_q=M`. At the 20 ns edge, `mstatus_q.mpp` (mst_mpp) becomes `2'b10` and `mpp_legal` becomes false. the clocked assertion reports the failure at formal step 3.

Formal checker, logs, and traces:

* https://github.com/codeadpool/cva6-priv-sva
* https://github.com/codeadpool/cva6-priv-sva/tree/main/evidence/probe/probe_mpp_legal_rvh_bmc

# Expected behavior

`mstatus.MPP` must read back as an implemented privilege mode. A write of `2'b10` must not be retained, regardless of `CVA6Cfg.RVH`.

Under CVA6's current legalization policy, the write should retain the previous legal `MPP` value.

# Observed behavior

With `RVH=1`, `MPP` retains `2'b10`.

The value then reaches other privilege logic:

* `mret` copies `mstatus_q.mpp` directly into `priv_lvl` (`csr_regfile.sv:2315`).
* With `MPRV=1`, ordinary loads and stores use `mstatus_q.mpp` as their effective privilege (`csr_regfile.sv:2280-2281`) without requiring `mret`.
* The MMU's first-stage U-bit checks compare the privilege against S and U.  `2'b10` matches neither branch.

This report only claims bypass of that mode-dependent U-bit check. It does not claim that independent PTE or PMP checks are bypassed. Where PMP is enabled, `2'b10` is treated as non-M and does not receive the M-mode bypass.

# Suggested fix

Reject `2'b10` independently of `RVH`:

```systemverilog
// MPP=2'b10 is reserved even when the H extension is implemented.
if ((mstatus_d.mpp == riscv::PRIV_LVL_HS) |
    (!CVA6Cfg.RVS & mstatus_d.mpp == riscv::PRIV_LVL_S) |
    (!CVA6Cfg.RVU & mstatus_d.mpp == riscv::PRIV_LVL_U)) begin
  mstatus_d.mpp = mstatus_q.mpp;
end
```

The S and U checks remain conditional because those values are legal when their modes are implemented. Only the `2'b10` check must be unconditional.

The similar guard in the `CSR_SSTATUS` path does not need a functional change: `SMODE_STATUS_WRITE_MASK` does not include MPP bits `[12:11]`, so an `sstatus` write cannot change `MPP`.

Clamping only the `mret` path would not fix the `MPRV` path. The value should be legalized when `mstatus` is written. 
Happy to open a PR for the fix

codeadpool:
## Why this PR is needed

`mstatus.MPP` is WARL over the implemented privilege modes, the current write path rejects the reserved encoding `2'b10` (`PRIV_LVL_HS`) only when `CVA6Cfg.RVH=0`:

```systemverilog
if ((!CVA6Cfg.RVH & mstatus_d.mpp == riscv::PRIV_LVL_HS) |
    (!CVA6Cfg.RVS & mstatus_d.mpp == riscv::PRIV_LVL_S) |
    (!CVA6Cfg.RVU & mstatus_d.mpp == riscv::PRIV_LVL_U)) begin
  mstatus_d.mpp = mstatus_q.mpp;
end
```

`2'b10` is not a legal privilege encoding for `MPP`, even when the hypervisor extension is implemented. HS-mode uses S-mode privilege (`2'b01`) with `V=0`; the H extension does not add a fourth privilege encoding.

With `RVH=1`, the first check is disabled, so an M-mode write of `MPP=2'b10` retains that reserved value.

## What changed

This PR makes the `MPP=2'b10` check unconditional by removing the `!CVA6Cfg.RVH` condition. the existing checks for unsupported S-mode and U-mode are unchanged.

An attempted write of `2'b10` now retains the previous `MPP` value under both `RVH=0` and `RVH=1`.

## Verification

Formal checking was run with yosys, yosys-slang, and SymbiYosys on `cv64a6_imafdch_sv39` (`RVH=1`). The checked property requires `mstatus.MPP` to remain in `{M, S, U}`.

PR #3387 was applied to both runs to remove the independent `dcsr.prv`/`dret` path:

* Before this change: counterexample at step 3.
* After this change: unbounded proof with PDR.

Checker and logs:  https://github.com/codeadpool/cva6-priv-sva/tree/main/evidence/probe/probe_mpp_legal_rvh_fixed_prove/

## Related issues

Fixes #3411.

This completes the `RVH=1` case missed by #1988/PR #2035 and #2274/PR #2285.

The trap-to-M path can separately copy an already-invalid `priv_lvl_q` into `MPP`. That independent `dcsr.prv`/`dret` path is covered by #3383 and PR #3387. This PR fixes the direct `mstatus` CSR-write path.

### DIFF
--- a/core/csr_regfile.sv
+++ b/core/csr_regfile.sv
@@ -1549,7 +1549,7 @@ module csr_regfile
             mstatus_d.tw   = riscv::Off;
             mstatus_d.mprv = riscv::Off;
           end
-          if ((!CVA6Cfg.RVH & mstatus_d.mpp == riscv::PRIV_LVL_HS) |
+          if ((mstatus_d.mpp == riscv::PRIV_LVL_HS) |
               (!CVA6Cfg.RVS & mstatus_d.mpp == riscv::PRIV_LVL_S) |
               (!CVA6Cfg.RVU & mstatus_d.mpp == riscv::PRIV_LVL_U)) begin
             mstatus_d.mpp = mstatus_q.mpp;
--- a/core/csr_regfile.sv
+++ b/core/csr_regfile.sv
@@ -1619,7 +1619,7 @@ module csr_regfile
             mstatus_d.mpv = 1'b0;
             mstatus_d.gva = 1'b0;
           end
-          if ((!CVA6Cfg.RVH & mstatus_d.mpp == riscv::PRIV_LVL_HS) |
+          if ((mstatus_d.mpp == riscv::PRIV_LVL_HS) |
               (!CVA6Cfg.RVS & mstatus_d.mpp == riscv::PRIV_LVL_S) |
               (!CVA6Cfg.RVU & mstatus_d.mpp == riscv::PRIV_LVL_U)) begin
             mstatus_d.mpp = mstatus_q.mpp;


Maanvi212006 (changes requested):
Reviewed and the core fix looks correct — legalizing mpp against the reserved HS (2'b10) encoding regardless of RVH matches spec behavior.
One note: the diff includes unrelated CI/workflow bumps and an hpdcache submodule bump (from merge-from-master commits, not yet in upstream master). Could you rebase onto origin/master so the diff is just the one-line fix? Makes it easier to review/merge cleanly.

cainria:
@Maanvi212006 I don't know which diff you are viewing but these are merges from upstream master and the current diff only contains the one-line change.

Maanvi212006:
> @Maanvi212006 I don't know which diff you are viewing but these are merges from upstream master and the current diff only contains the one-line change.

@cainria Ah, my mistake! I was looking at the wrong commit history. Thanks for clearing that up.

cainria:
Cc @JeanRochCoulon this PR is ready to rebase and merge

PR #3415, opened 2026-07-27
GitHub CI: remove pull_request_target from Verible

cainria:
Due to new security measures, GitHub discourages from using checkout of PR head with pull_request_target.

Because of this change, comments will not appear in the PR anymore. Still, the formatting check will be performed like any other GitHub worflow.

### DIFF
--- a/.github/workflows/verible.yml
+++ b/.github/workflows/verible.yml
@@ -4,19 +4,13 @@
 
 name: Verible
 on:
-  pull_request_target:
+  pull_request:
 
 jobs:
   format:
     runs-on: ubuntu-latest
-    permissions:
-      checks: write
-      contents: read
-      pull-requests: write
     steps:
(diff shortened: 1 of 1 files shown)


cainria:
Ok so in this PR checks we have 2 Verible / format workflows
- pull_request_target (removed by this PR, but still shown as failed as it is executed from `master`) and
- pull_request (added by this PR and passing)

As mentioned in the PR description, **there will not be suggestions anymore as PR comments**.
I'll check how formatting issues are rendered in workflow logs before marking this PR as ready for review.

cainria:
Here is how it is rendered: https://github.com/openhwgroup/cva6/actions/runs/30268804311/job/89985933207?pr=3415#step:3:462

This is not as fancy as the suggestions we had before, but at least this PR fixes the workflow and its security vulnerability.

PR #3421, opened 2026-07-30
Master candidate dcls

valentinThomazic:
Proposed base of work for common DCLS implementation.

Elaboration top becomes **cva6_top** which instantiate either a single CVA6 or a CVA6 DCLS with its logic control.

The \`dcls_logic\` module shifts the main core inputs by a configurable delay to drive the shadow core. it also performs the core outputs comparison.

The \`dcls_common\` instantiates common modules depending on whether they are configured to be shared or not. Both cores either instantiate the module or send their control signals and receive data from the dcls common modules at the top.

It compares the received control inputs from both cores and outputs and send the result to the cores, shifted by n cycles for the shadow core.

The comparators output synchronous alarms.

Current options are:

* Configurable delay between the 2 cores
* Common Regfile (without integrity or error correction)
* Common BHT

Target **cv32a60x** is set with:

* 2 Cycle of delay
* Common Regfile & Common BHT

Others configurations are left unchanged and function as before

### DIFF
--- a/core/dcls_delay_ff.sv
+++ b/core/dcls_delay_ff.sv
@@ -1,0 +1,33 @@
+// Copyright 2026 Thales France
+//
+// Licensed under the Solderpad Hardware Licence, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// SPDX-License-Identifier: Apache-2.0 WITH SHL-2.0
+// You may obtain a copy of the License at https://solderpad.org/licenses/
+//
+// Original Author: Valentin Thomazic (valentin.thomazic@thalesgroup.com)
+
+module dcls_delay_ff #(
+    parameter type data_t = logic,
+    parameter int unsigned DCLS_DELAY = 1
+) (
+    input  logic  clk_i,
+    input  logic  rst_ni,
+    input  data_t data_i,
+    output data_t data_o
+);
+  data_t [DCLS_DELAY-1 : 0] data_n;
+
+  assign data_o = data_n[DCLS_DELAY-1];
+
+  always_ff @(posedge clk_i or negedge rst_ni) begin
+    if (~rst_ni) begin
+      data_n <= '{default: 0};
+    end else begin
+      data_n[0] <= data_i;
+      for (int i = 1; i < DCLS_DELAY; i++) begin
+        data_n[i] <= data_n[i-1];
+      end
+    end
+  end
+endmodule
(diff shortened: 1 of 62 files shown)


cainria:
:warning: Verilator was not run in this PR :warning: 

It was a default-branch workflow, but due to new GitHub security rules, we switched to a per-branch workflow #3415

We may want to cherry-pick #3415 in `master_candidate` before or after merging this PR, so that Verible checks the code in that branch too! (#3403 might be required first to avoid git conflicts)
EDIT: #3436 is needed too to freeze Verible version since a recent release changed the expected formatting.

core/dcls_delay_ff.sv
cainria, line 1: This looks like a duplicate of `shift_reg` from pulp-platform, which is used in CVA6 load unit.

core/dcls_logic.sv
```diff
@@ -0,0 +1,81 @@
...
+  );
+  dcls_comparator #(
+      .data_t(core_outputs_t)
+  ) i_dcls_comparator_dup (
+      .clk_i,
+      .rst_ni  (rst_shadow_no),
+      .main_i  (main_outputs_delayed),
+      .shadow_i(shadow_outputs_i),
+      .alarm_o (alarm_o[1])
+  );
```
cainria: Could we use a `for` loop here? Or would it prevent us from telling the synthesis tool to not merge these duplicates instances?

cainria (commented):
Nice contribution! I like the way you handled the common register file.

core/dcls_delay_ff.sv
valentinThomazic: You are right, I will use it instead.
Using this module also gives the option of having no delay between the cores.

core/dcls_logic.sv
valentinThomazic: I tested it and it still synthesizes both comparators, at least with our synthesis flow.
I will use a generate loop then

JeanRochCoulon:
@AngelaGonzalezMarino This DCLS is ready to be merged. Could I ask you to run the Linux boot to test the regression ?  If all ok we could think about merging it into master_candidate.

AngelaGonzalezMarino:
> @AngelaGonzalezMarino This DCLS is ready to be merged. Could I ask you to run the Linux boot to test the regression ? If all ok we could think about merging it into master_candidate.

hi @JeanRochCoulon I have tested the following configurations on genesys 2:

- cv32a6_ima_sv32_fpga
- cv64a6_imafdc_sv39_hpdcache_pmp_mmu_axi
- cv64a6_imafdc_sv39_hpdcache_wb

The 3 boot OK and it is possible to play tetris!

I would like to do the same check on agilex just to be sure (however the platform is right now travelling to one of my colleague's, so he can do it once he gets it)

Another thing we should check is CI result. Do you have a pipeline where we can see result of simulations?

PR #3423, opened 2026-07-31
Integration of Error Correction Mechanisms of the HPDcache with the CVA6

cfuguet:

This PR introduces the following mechanisms in the CVA6:
- Add different configuration bits in the existing custom DCACHE CSR to control and monitor the different error correction mechanisms in the Dcache (HPDcache).
- Add two additional configuration parameters to the CVA6: DcacheEccEnable, DcacheEccScrubberEnable
- Trigger an exception in the core in case of a load access on the cache that responds with an error. An error in this case may indicate an uncorrected error due to a transient error in the cache, or due to an error propagated from the NoC through the cache (e.g., ECC error in DRAM, access to an illegal physical address).
- Add C test with simple sanity check for the scrubbing mechanism in the HPDcache
- Add pre-generated ECC encoders/decoders in the Flist file for common configurations of the cache

These features support both 32-bit and 64-bit configurations of the CVA6.

I did not modify any existing configuration in the core/include directory to enable ECC (but any of them can benefit from error correction in the Dcache by setting the DcacheEccEnable and DcacheEccScrubberEnable parameters to 1).

**If accepted, I suggest that this PR is rebased and merged (not squashed). Each commit is functional separately and introduces the different functionalities mentioned above.**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -336,6 +336,8 @@ incdir := $(CVA6_REPO_DIR)/vendor/pulp-platform/common_cells/include/ $(CVA6_REP
           $(CVA6_REPO_DIR)/verif/core-v-verif/lib/uvm_agents/uvma_core_cntrl/ \
           $(CVA6_REPO_DIR)/verif/tb/core/ \
           $(CVA6_REPO_DIR)/core/include/ \
+          $(HPDCACHE_DIR)/rtl/include \
+          $(HPDCACHE_DIR)/rtl/src/utils/ecc \
           $(CVA6_REPO_DIR)/corev_apu/instr_tracing/ITI/include \
           $(SPIKE_INSTALL_DIR)/include/disasm/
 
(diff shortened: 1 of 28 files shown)


cainria (approved):

JeanRochCoulon:
image

cfuguet:
I've just rebased on master and fixed the issues reported by Spyglass

JeanRochCoulon:
During FPGA build:
image

cfuguet:
Weird... The error is on the generation of an implementation checkpoint of the Xilinx's Memory DDR controller (MIG). This PR does not touch the FPGA flow.

JeanRochCoulon:
BTW do you implement other memories, in this case FPGA flow could need to use other component.
Good news, ASIC synthesis is PASS

JeanRochCoulon:
THe FPGA issue does not come from your PR, master gets the same issue !  I am investigating

JeanRochCoulon:
I think I found the FPGA issue:
image

cfuguet:
Good catch @JeanRochCoulon ! I've made the required modifications to make the header files directories to be included

PR #3438, opened 2026-08-04
Fix MRET virtualization mode for M-mode

### LINKED ISSUE #3313: [BUG] MRET incorrectly sets V=1 when MPP=M-mode

124107157-KV:

## Why this PR is needed

When the Hypervisor extension is enabled, the current `MRET` path
restores virtualization mode `V` directly from `mstatus.MPV`.

This is incorrect when `mstatus.MPP` contains M-mode. In this case,
`MRET` must keep `V` cleared, regardless of the value stored in `MPV`.
The current implementation can therefore return to M-mode with an
incorrect virtualization state.

## Changes

This PR updates the RVH-enabled `MRET` handling in
`core/csr_regfile.sv` to:

* force `V` to zero when `MPP` is M-mode;
* continue restoring `V` from `MPV` for lower privilege modes;
* preserve the existing `MPV` and `MPRV` update behavior.

It also adds the directed `mret-mpv-rv64` regression test and registers
it in `verif/tests/testlist_issues.yaml`.

## Regression test

The test sets `MPP=M` and `MPV=1`, executes `MRET`, and then triggers an
M-mode `ECALL`.

Because `V` cannot be read directly, the test checks the value saved
into `mstatus.MPV` during trap entry. A nonzero saved value indicates
that `MRET` incorrectly retained `V=1`.

The test also verifies that:

* `MPV` can be written before executing `MRET`;
* the trap originated from M-mode;
* `mcause` reports an M-mode environment call.

## Verification

Test target:

```text
cv64a6_imafdch_sv39
```

The test passes on the Spike reference model:

```text
Spike return code: 0
```

The original CVA6 RTL reproduces the issue:

```text
mret-mpv-rv64.o *** FAILED *** (tohost = 8)
```

After the fix, the CVA6 and Spike executions match:

```text
[PASSED]: 128 matched
1 PASSED, 0 FAILED
Final regression return code: 0
```

`core/csr_regfile.sv` also passes the Verible formatting check.

## Related issue

Fixes #3313.

## Limitations

The directed regression covers the reported `MPP=M, MPV=1` case.

A lower-privilege control test was investigated but encountered an
instruction-access fault before reaching its intended check, so it is
not included in this regression. The RTL expression continues to
restore `V` from `MPV` whenever `MPP` is not M-mode.

This change does not modify the CSR layout, configuration interface, or
behavior outside the RVH-enabled `MRET` path.

### DIFF
--- a/core/csr_regfile.sv
+++ b/core/csr_regfile.sv
@@ -2344,8 +2344,8 @@ module csr_regfile
       // set mpie to 1
       mstatus_d.mpie = 1'b1;
       if (CVA6Cfg.RVH) begin
-        // set virtualization mode
-        v_d           = mstatus_q.mpv;
+        // MPV is ignored when returning to M-mode.
+        v_d = (mstatus_q.mpp == riscv::PRIV_LVL_M) ? 1'b0 : mstatus_q.mpv;
         //set mstatus mpv to false
         mstatus_d.mpv = 1'b0;
         if (mstatus_q.mpp != riscv::PRIV_LVL_M) mstatus_d.mprv = 1'b0;
--- a/verif/tests/custom/issues/mret-mpv-rv64.S
+++ b/verif/tests/custom/issues/mret-mpv-rv64.S
@@ -1,0 +1,116 @@
+# Copyright 2026 Keerthivasan
+#
+# Licensed under the Solderpad Hardware Licence, Version 2.0.
+# SPDX-License-Identifier: Apache-2.0 WITH SHL-2.0
+#
+# Regression test for CVA6 issue #3313.
+#
+# When MRET returns to M-mode, virtualization mode V must remain zero,
+# regardless of the value previously stored in mstatus.MPV.
+#
+# V is internal architectural state and cannot be read directly.
+# A subsequent trap to M-mode saves V into mstatus.MPV, allowing the
+# post-MRET V state to be checked.
+
+  .section .text
+  .align 2
+
+  .equ MSTATUS_MPP_MASK, 0x0000000000001800
+  .equ MSTATUS_MPP_M,    0x0000000000001800
+  .equ MSTATUS_MPV,      0x0000008000000000
+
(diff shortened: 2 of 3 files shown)


124107157-KV:
I investigated the failed external CI run for PR #3438. All four failures occur during Questa elaboration in `core/cva6_mmu/cva6_mmu.sv:400` due to the reversed part-select `[31:34]` on `fetch_vaddr[31:0]`. This code is unchanged between `master` and my branch, and the failing `cv32a6_imac_sv32` target has RVH disabled. Therefore, the failure is unrelated to issue #3313 or the MRET fix. The directed `mret-mpv-rv64` regression passes with the patch.

cainria:
Indeed, here are all the runs: https://riscv-ci.pages.thales-invia.fr/dashboard/dashboard_cva6.html

The latest `master` run fails on these tests. Maintainers usually are aware of this and will not consider these tests for the review.

JeanRochCoulon:
Thanks @124107157-KV (and @cainria) for this fix. I merge !

Issue #3439, opened 2026-08-04
[BUG] frontend/instr_scan incorrectly decodes reserved C.JR and C.EBREAK as valid RVC jump instructions


Misbahud-Din:
### Code of Conduct

- [x] I have searched the existing bug issues.
- [x] I am a human engaging in an interpersonal interaction. During this interaction, my words are my own and are not generated. If relevant, I provide links to my sources.

### CVA6 commit affected

030246b12441bb580a05717f7bc9bb9ba9c27798

### Bug Description

In `frontend/instr_scan`, the `rvc_jr_o`, `rvc_jalr_o`, and `rvc_call_o` signals are asserted for reserved or incorrect compressed instructions.

### Relevant Specification Excerpt
> C.JR (jump register) performs an unconditional control transfer to the address in register rs1. C.JR expands to `jalr x0, 0(rs1)`. C.JR is valid only when rs1!=x0; the code point with rs1=x0 is reserved.

> C.JALR (jump and link register) performs the same operation as C.JR, but additionally writes the address of the instruction following the jump (pc+2) to the link register, x1. C.JALR expands to `jalr x1, 0(rs1)`. C.JALR is valid only when rs1!=x0; the code point with rs1=x0 corresponds to the C.EBREAK instruction.

### Buggy Code

```sv
  assign is_jal_r     = (instr_i[15:13] == riscv::OpcodeC2JalrMvAdd)
                        & (instr_i[6:2] == 5'b00000)
                        & (instr_i[1:0] == riscv::OpcodeC2);
  assign rvc_jr_o = is_jal_r & ~instr_i[12];
  // always links to register 1 e.g.: it is a jump
  assign rvc_jalr_o = is_jal_r & instr_i[12];
  assign rvc_call_o = rvc_jalr_o | rv32_rvc_jal;
```

### Reproduction

- `instr_i = 16'h8002` (`c.jr` with `rs1 = x0`) is reserved, but `rvc_jr_o` is high.

<img width="710" height="291" alt="Image" src="https://github.com/user-attachments/assets/ec6f7a78-7a65-4d1e-82af-af0097ea78a2" />

- `instr_i = 16'h9002` (`c.ebreak`) is incorrectly classified as `c.jalr`, so `rvc_jalr_o` is high.

<img width="710" height="291" alt="Image" src="https://github.com/user-attachments/assets/42356222-9947-498e-8919-48b6992e8f4b" />

### Expected Behavior

[...]

PR #3444, opened 2026-08-05
Fixed classifications of c.jr and c.jalr instruction in instr_scan

Misbahud-Din:

## Description
Fixes incorrect assertion of `rvc_jr_o` and `rvc_jalr_o` in `frontend/instr_scan` for reserved/misclassified compressed instructions.

The `is_jal_r` decode logic only checked `funct3`, `instr_i[6:2]`, and the opcode field, but never checked `rs1` (`instr_i[11:7]`). As a result:
- `c.jr` with `rs1 = x0` (reserved encoding) incorrectly asserted `rvc_jr_o`.
- `c.ebreak` (`funct3 = 100`, bit12=1, rs1=0, rs2=0) was misdecoded as `c.jalr`, incorrectly asserting `rvc_jalr_o`.

Since `rvc_call_o` is derived from `rvc_jalr_o`, it was also incorrectly asserted for `c.ebreak`.

## Fix
Added `instr_i[11:7] != 5'b00000` to the `is_jal_r` condition, per the RISC-V C extension spec:
- C.JR is valid only when `rs1 ≠ x0`.
- C.JALR is valid only when `rs1 ≠ x0`; `rs1 = x0` is C.EBREAK, not C.JALR.

## Related Issue
#3439

## Verification
- Re-ran Jasper formal properties against both repro vectors from the linked issue:
  - `16'h8002` (`c.jr`, `rs1=x0`) → `rvc_jr_o` now low.
  - `16'h9002` (`c.ebreak`) → `rvc_jalr_o` now low.

### DIFF
--- a/core/frontend/instr_scan.sv
+++ b/core/frontend/instr_scan.sv
@@ -92,6 +92,7 @@ module instr_scan #(
   // always links to register 0
   logic is_jal_r;
   assign is_jal_r     = (instr_i[15:13] == riscv::OpcodeC2JalrMvAdd)
+                        & (instr_i[11:7] != 5'b00000)
                         & (instr_i[6:2] == 5'b00000)
                         & (instr_i[1:0] == riscv::OpcodeC2);
   assign rvc_jr_o = is_jal_r & ~instr_i[12];


JeanRochCoulon:
Thanks for your interest. Please sign-up the Eclipse ECA agreement. Change the Draft annotation when the ECA will be fixed.

JeanRochCoulon:
This bug is not easy to find because it corrupts the BTB but does not create functional bug. Thanks for this funding and the fix. I will merge a soon as the CI is green.
Feel free to submit other bugs/fixes ;-)

PR #3451, opened 2026-08-08
Fix Zcmt JVT index calculation

### LINKED ISSUE #3442: [BUG] JVT address calculation truncates the 8-bit table index
### Code of Conduct

### CVA6 commit affected

88e810c

## Description

CVA6 calculates the jump-vector-table address using only the lower six bits of the instruction index:

```systemverilog
table_address = {jvt_i.base, 6'b000000}
              + {24'h0, instr_i[7:2], 2'b00};
```

However, Zcmt defines a 256-entry table and encodes the 8-bit `index` in instruction bits `[9:2]`. For RV32, the specification requires:

```text
table_address = jvt.base + (index << 2)
```

Therefore, the current implementation truncates the index from 8 bits to 6 bits. `cm.jalt` indices 64–255 alias entries 0–63.

## Example

For `cm.jalt 64`:

- Expected address: `jvt.base + (64 << 2) = jvt.base + 256`
- CVA6 address: `jvt.base + (0 << 2) = jvt.base`

The instruction is still decoded as `cm.jalt` using the complete index:

```systemverilog
if (instr_i[9:2] < 32) begin
  // cm.jt
end else if (instr_i[9:2] >= 32) begin
  // cm.jalt
end
```

Consequently, CVA6 fetches the wrong JVT entry and jumps to an incorrect target.

## Spike reference implementation

Spike extracts all eight index bits:

```cpp
// riscv/decode.h
uint64_t rvc_index() { return x(2, 8); }
```

It then uses the complete index in the address calculation:

```cpp
// riscv/insns/cm_jalt.h
reg_t index = insn.rvc_index();

if (xlen == 32)
  target = MMU.fetch_jump_table<int32_t>(base + (index << 2));
else
  target = MMU.fetch_jump_table<int64_t>(base + (index << 3));
```

This matches the specification and confirms that bits `[9:2]` must participate in the JVT address calculation.

## Suggested fix

Use the complete index field:

```systemverilog
table_address = {jvt_i.base, 6'b000000}
              + {22'h0, instr_i[9:2], 2'b00};
```

124107157-KV:

## Why this PR is needed

This PR fixes the Zcmt JVT address calculation reported in #3442.

The Zcmt table-jump index is encoded in instruction bits `[9:2]`, providing an 8-bit index for 256 JVT entries. For RV32, the JVT entry address is calculated as:

```text
table_address = jvt.base + (index << 2)
```

The current implementation uses only `instr_i[7:2]`:

```systemverilog
table_address = {jvt_i.base, 6'b000000}
              + {24'h0, instr_i[7:2], 2'b00};
```

This retains only six index bits. Therefore, indices 64–255 lose their upper two bits and alias entries 0–63.

This PR changes the address calculation to use the complete 8-bit index:

```systemverilog
table_address = {jvt_i.base, 6'b000000}
              + {22'h0, instr_i[9:2], 2'b00};
```

This is also consistent with the existing Zcmt instruction classification, which already uses `instr_i[9:2]`.

For a JVT base of `0x80100000`:

| Index | Expected offset | Correct address | Previous address |
| ----: | --------------: | --------------: | ---------------: |
|     0 |         `0x000` |    `0x80100000` |     `0x80100000` |
|    63 |         `0x0fc` |    `0x801000fc` |     `0x801000fc` |
|    64 |         `0x100` |    `0x80100100` |     `0x80100000` |
|   127 |         `0x1fc` |    `0x801001fc` |     `0x801000fc` |
|   128 |         `0x200` |    `0x80100200` |     `0x80100000` |
|   255 |         `0x3fc` |    `0x801003fc` |     `0x801000fc` |

Index 64 is the first value that exposes the truncation bug. Its correct offset is `0x100`, while the previous six-bit calculation wraps the index to zero and produces offset `0x000`.

## Changes

This PR:

* updates `core/zcmt_decoder.sv` to use `instr_i[9:2]` for the JVT address calculation;
* adds a focused `zcmt_decoder` Verilator regression;
* checks JVT indices `0`, `31`, `32`, `63`, `64`, `127`, `128`, and `255`;
* adds a standalone Makefile for the focused decoder test;
* invokes the regression from `verif/regress/issue-tests.sh`;
* propagates a failing regression result before the remaining issue-test flow continues.

The values below 64 verify that the existing behavior before the truncation boundary remains unchanged.

Index 64 verifies the first affected boundary, while 127, 128, and 255 exercise the upper portion of the complete 8-bit index range.

### Before the fix

The regression passes through index 63 and fails at index 64:

```text
index=0 encoding=a002 actual=80100000 expected=80100000
index=31 encoding=a07e actual=8010007c expected=8010007c
index=32 encoding=a082 actual=80100080 expected=80100080
index=63 encoding=a0fe actual=801000fc expected=801000fc
index=64 encoding=a102 actual=80100000 expected=80100100

Assertion failed:
index 64: wrong table address: actual=80100000 expected=80100100
```

The pre-fix regression exits with a non-zero status.

### After the fix

The same regression passes across the tested 8-bit index range:

```text
index=0 encoding=a002 actual=80100000 expected=80100000
index=31 encoding=a07e actual=8010007c expected=8010007c
index=32 encoding=a082 actual=80100080 expected=80100080
index=63 encoding=a0fe actual=801000fc expected=801000fc
index=64 encoding=a102 actual=80100100 expected=80100100
index=127 encoding=a1fe actual=801001fc expected=801001fc
index=128 encoding=a202 actual=80100200 expected=80100200
index=255 encoding=a3fe actual=801003fc expected=801003fc

PASS: complete eight-bit Zcmt JVT index is used
```

## Validation

The final branch was validated with:

* Verilator 5.008;
* focused `tb_zcmt_decoder` regression;
* fail-before/pass-after verification using the same permanent regression;
* `verible-verilog-format --verify core/zcmt_decoder.sv`;
* `verible-verilog-format --verify corev_apu/tb/tb_zcmt_decoder/hdl/tb.sv`;
* `git diff upstream/master...HEAD --check`;
* rebase verification against the current `upstream/master`.

The working tree is clean after validation.

Full repository CI will run on the pull request.

## Scope and limitations

This PR intentionally addresses only the JVT index truncation reported in #3442.

It changes the JVT table-entry address calculation and adds focused regression coverage for that calculation.

It does not modify unrelated Zcmt control-flow behavior, cache-response handling, or the `TABLE_JUMP` state machine.

No known limitation is introduced by the RTL change itself.

Fixes #3442


corev_apu/tb/tb_zcmt_decoder/hdl/tb.sv
```diff
@@ -0,0 +1,179 @@
...
+
+  function automatic config_pkg::cva6_cfg_t make_cfg();
+    config_pkg::cva6_cfg_t cfg;
+
+    cfg = config_pkg::cva6_cfg_empty;
+
+    cfg.XLEN = 32;
+    cfg.VLEN = 32;
+    cfg.IS_XLEN32 = 1'b1;
+    cfg.IS_XLEN64 = 1'b0;
```
cainria: Should we use `build_config` on a user config instead?

cainria (commented):
Thanks for the fix!

Should we add a new testbench or is there a way to trigger this issue with assembly code?

124107157-KV:
> Thanks for the fix!
> 
> Should we add a new testbench or is there a way to trigger this issue with assembly code?

Thanks! Yes, it should be possible to trigger this with an assembly test. I found the existing Zcmt assembly tests under `verif/tests/custom/zcmt/`, including `cm_jalt_long.S`.

I’ll check whether I can adapt one of those to use index 64, which is the first value affected by this bug, and make the test self-checking so that an incorrect jump is detected. If that works cleanly, I can use the assembly regression instead of adding a separate decoder testbench.

corev_apu/tb/tb_zcmt_decoder/hdl/tb.sv
124107157-KV: Yes, I agree. If the standalone decoder testbench is retained, I’ll switch it to use `build_config_pkg::build_config()` with a user config instead of manually filling the required `cva6_cfg_t` fields. That should also keep the test consistent with the existing CVA6 testbench style.

124107157-KV: Update after the review feedback: I pushed commit `939e2b5b3` replacing the standalone decoder testbench with an assembly regression using `cm.jalt` index 64, as suggested.

The regression is now registered as `zcmt-jvt-index-rv32` and runs with `RVZCMT=1`.

Verified with the same test:
* original `[7:2]` calculation: FAILED
* fixed `[9:2]` calculation: PASSED

Since the standalone testbench was removed, the earlier `build_config` comment no longer applies.

Thanks for the review and suggestion.

JeanRochCoulon:
Yes, this fixes the JVT table bug. Good !

PR #3453, opened 2026-08-10
Add false 4 KiB alias regression

Srinidhi-Magesh:
- [x] I am a human engaging in an interpersonal interaction. During this
interaction, my words are my own and are not generated. If relevant, I provide
links to my sources.

## Why is this PR needed?

This adds deterministic regression coverage for a conservative CVA6 LSU false
4 KiB store-load alias corner case.

The store buffer compares address bits `[11:3]` when checking dependencies.
Accesses on different 4 KiB pages with the same page offset can therefore appear
dependent. This test verifies architectural correctness and forward progress in
that situation.

This is verification coverage only. It does not identify an RTL bug or claim a
performance regression.

## What changed?

Added the directed `false_4k_alias` test and registered it in
`testlist_custom.yaml`.

The test:

1. Creates two distinct 4 KiB-aligned pages.
2. Initializes `page_b + 0x18` with a sentinel.
3. Issues an older store to `page_a + 0x18`.
4. Immediately loads from `page_b + 0x18`.
5. Checks that the load returns the sentinel from `page_b`.

The addresses intentionally share address bits `[11:3]`.

## Verification

Target: `cv64a6_imafdc_sv39`

Simulators:

- `veri-testharness`
- Spike

Result:

```text
[PASSED]: 1231 matched
7 PASSED, 0 FAILED


JeanRochCoulon:
Thanks for improving verification !

Issue #3456, opened 2026-08-11
[BUG] HLV.WU unexpectedly reads through an execute-only G-stage page after a prior HLVX.WU


jf-cc727:
### Code of Conduct

- [x] I have searched the existing bug issues.
- [x] I am a human engaging in an interpersonal interaction. During this interaction, my words are my own and are not generated. If relevant, I provide links to my sources.

### CVA6 commit affected

c4c412a9f8c3b9f21ca7f7b6c58b612ec2afe724

### Bug Description

# Introduction 

I found an RVH permission issue on CVA6.

On the `cv64a6_imafdch_sv39` configuration, a plain `HLV.WU` can unexpectedly read from an execute-only G-stage page if the same guest address was previously accessed by a legal `HLVX.WU`.

The minimal access sequence is:

```asm

[rvh-hlvx-to-hlv-execute-only-disclosure.zip](https://github.com/user-attachments/files/30931550/rvh-hlvx-to-hlv-execute-only-disclosure.zip)

hlvx.wu x6, (x5)
hlv.wu  x6, (x5)
```

In my PoC, I install a single Sv39x4 G-stage leaf with:

- `V = 1`
- `X = 1`
- `U = 1`
- `A = 1`
- `R = 0`

and I keep `mstatus.MXR = 0` throughout the run.

Under these conditions:

- `HLVX.WU` is expected to succeed, because it may use execute permission during translation
- the following plain `HLV.WU` should still be treated as a normal read
- since the same G-stage leaf is not readable, that second access should raise a load guest-page fault

However, CVA6 instead allows the second `HLV.WU` to read through and return the guest word.

I also ran a control test with the same X-only G-stage mapping but without the preceding `HLVX.WU`, and in that case CVA6 traps correctly. So this does not look like a generic "`HLV.WU` can read an X-only G-stage page" issue. The unexpected behavior appears only after a prior legal `HLVX.WU` to the same guest address.

# Steps to reproduce

1. Use CVA6 with RVH enabled on the `cv64a6_imafdch_sv39` configuration.

2. Disable first-stage translation for the test and keep `mstatus.MXR = 0`.

3. Install exactly one Sv39x4 G-stage leaf for `GVA = 0x40000000` with:

   - `V = 1`
   - `X = 1`
   - `U = 1`
   - `A = 1`
   - `R = 0`

4. Perform one legal `HLVX.WU` to that guest virtual address.

[...]

PR #3468, opened 2026-08-12
fix(mmu): re-check cached G-stage DTLB-hit permissions after HLVX

jf-cc727:
- [√] I have searched for similar pull requests
- [√ ] I am a human engaging in an interpersonal interaction. During this interaction, my words are my own and are not generated. If relevant, I provide links to my sources.

This PR fixes a guest-stage permission issue on the cached G-stage / DTLB-hit path after a prior legal `HLVX` access.

It latches `hlvx_inst_i` together with the saved LSU request and re-checks guest-stage DTLB-hit permissions against `U`, `A`, and the current load `R/X` rules. This prevents a prior legal `HLVX` access from causing a later plain `HLV` load to inherit execute-only permission on a cached translation.

Related issue: #3456 

Why this PR is still needed:

[#3435](https://github.com/openhwgroup/cva6/pull/3435) fixes a related but different PTW/page-walk path, where an implicit G-stage access happens while translating a VS-stage PTE address. However, the original CV-1 reproducer still hits an additional cached G-stage / DTLB-hit path in `cva6_mmu.sv`, which does not go back through that PTW logic.

Validation:

- the original CV-1 reproducer still reproduces on top of #3435
- the same reproducer passes with this patch applied (`tohost = 0`)

Current limitation:

This is a targeted fix for the cached guest-stage hit path exercised by the reproducer. It does not refactor the related PTW logic.

### DIFF
--- a/core/cva6_mmu/cva6_mmu.sv
+++ b/core/cva6_mmu/cva6_mmu.sv
@@ -506,6 +506,9 @@ module cva6_mmu
   logic [CVA6Cfg.GPLEN-1:0] lsu_gpaddr_n, lsu_gpaddr_q;
   logic [31:0] lsu_tinst_n, lsu_tinst_q;
   logic hs_ld_st_inst_n, hs_ld_st_inst_q;
+  logic hlvx_inst_n, hlvx_inst_q;
+  logic g_load_exec_as_read_ok;
+  logic g_load_access_ok;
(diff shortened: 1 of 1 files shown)


cainria:
Thanks for the contribution,

Do not forget to sign the required ECA

jf-cc727:
Thank you for your review! I have signed the ECA.

Issue #3459, opened 2026-08-11
[BUG] VS-mode SFENCE.VMA with rd!=0 retires instead of raising illegal instruction


jf-cc727:
### Code of Conduct

- [x] I have searched the existing bug issues.
- [x] I am a human engaging in an interpersonal interaction. During this interaction, my words are my own and are not generated. If relevant, I provide links to my sources.

### CVA6 commit affected

c4c412a9f8c3b9f21ca7f7b6c58b612ec2afe724

### Bug Description

# Introduction

I found a decoder-correctness issue on the CVA6 RVH / VS-mode path for `SFENCE.VMA`.

The supplied PoC enters VS-mode and executes one invalid raw `SFENCE.VMA` encoding:

```asm
bad_sfence_vs:
  .word 0x120000f3
```

This is the `rd = x1` variant of `sfence.vma x0, x0`.

For `SFENCE.VMA`, the `rd` field is not a destination register operand and must be `x0`. So this encoding should raise an illegal-instruction trap. Entering VS-mode does not make `rd != 0` acceptable.

However, on the current CVA6 RVH configuration, this invalid VS-mode instruction retires and falls through instead of trapping. The same PoC on the vendor Spike reference raises `trap_illegal_instruction` at the target instruction.

So this looks like a decoder bug on the RVH / VS-mode `SFENCE.VMA` path: an encoding that should remain illegal is being accepted and retired.

# Steps to reproduce

1. Use CVA6 with RVH enabled on the `cv64a6_imafdch_sv39` configuration.

2. Install an M-mode trap handler in `mtvec`.

3. Clear interrupt/delegation state and set up PMP so the test can reach VS-mode cleanly.

4. Check that the H extension is present.

5. Drop from M-mode to HS-mode with `mret`.

6. In HS-mode:
   - set `hstatus.SPV = 1`
   - set `sepc = vs_entry`
   - enter VS-mode with `sret`

7. Make the very first VS-mode instruction the invalid raw encoding:

```asm
vs_entry:
bad_sfence_vs:
  .word 0x120000f3
  EXIT 77
```

8. Treat only the following as success:
   - `mcause = 2`
   - `mepc = bad_sfence_vs`

If the invalid instruction retires and falls through, the PoC exits `77`.

The essential PoC logic is:

```asm
vs_entry:
bad_sfence_vs:
  .word 0x120000f3
  EXIT 77
```

So the oracle is direct:

[...]

PR #3470, opened 2026-08-12
fix(decoder): preserve SFENCE.VMA rd legality in RVH VS-mode path

jf-cc727:
- [√] I have searched for similar pull requests
- [√] I am a human engaging in an interpersonal interaction. During this interaction, my words are my own and are not generated. If relevant, I provide links to my sources.

This PR fixes a VS-mode `SFENCE.VMA` legality bug under RVH.

When `CVA6Cfg.RVH && v_i`, the decoder correctly clears the temporary virtual-illegal condition for the legal VS-mode case. However, that path can also overwrite an earlier structural-illegal result, such as `rd != x0`, and incorrectly make the instruction legal.

This change keeps `SFENCE.VMA` illegal when `rd != x0` while still preserving the intended RVH / VS-mode behavior for the legal case.

Validated with the test reproducer (`tohost = 0`).

Related issue: #3459

### DIFF
--- a/core/decoder.sv
+++ b/core/decoder.sv
@@ -301,6 +301,7 @@ module decoder
                     // only if S mode is supported
                     // otherwise decode an illegal instruction
                     if (CVA6Cfg.RVH && v_i) begin
+                      illegal_instr = (instr.itype.rd == '0) ? illegal_instr : 1'b1;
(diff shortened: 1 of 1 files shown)


cainria:
It seems you are not using the same email for all your contributions, hence the ECA check failure.
The one used in the two PRs where the ECA check fails is `caijf@loca*host DOT localdomain`

JeanRochCoulon:
Thanks for your fix and your interest for CVA6. I Merge !

PR #3486, opened 2026-08-22
Block TLB updates after PTW faults

flaviens:
PTW faults could still install a stale shared TLB entry.

Block updates after invalid Svnapot leaves. Apply the same guard to out-of-range guest PPNs.

### DIFF
--- a/core/cva6_mmu/cva6_ptw.sv
+++ b/core/cva6_mmu/cva6_ptw.sv
@@ -516,6 +516,8 @@ module cva6_ptw
                 end
               end
 
+              if (CVA6Cfg.SvnapotEn && pte.n && !is_napot_64k) shared_tlb_update_valid = 1'b0;
+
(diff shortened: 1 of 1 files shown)


JeanRochCoulon:
@124107157-KV Can you approve this PR ?

124107157-KV:
Sure, I will look into this. Thanks.

124107157-KV:
Reviewed the PTW flow and the changes look correct to me.

The added guards correctly deassert `shared_tlb_update_valid` after the paths that can assert it, preventing invalid Svnapot leaves and out-of-range VS-stage guest PPNs from updating the TLB when the PTW detects a fault. This is also consistent with the existing handling of misaligned-page and PMP fault cases.

The changes are minimal and the current CI is green.

Non-blocking: it would be useful to add targeted regression coverage for these two fault paths in the future, since this PR does not include a dedicated test-case for them.

Issue #3177, opened 2026-01-06
[BUG] `M-mode PMP`: lower-priority locked entry overrides higher-priority unlocked entry


hjw-arch:
### Is there an existing CVA6 bug for this?

- [x] I have searched the existing bug issues

### Description
We have observed a discrepancy between the CVA6 RTL execution and the Spike simulation regarding PMP priority handling in `M-Mode`.

According to the RISC-V Privileged Specification:
> PMP entries are statically prioritized. The lowest-numbered PMP entry that matches any byte of an access determines whether that access succeeds or fails. The matching PMP entry must match all bytes of an access, or the access fails, irrespective of the L, R, W, and X bits. For example, if a PMP entry is configured to match the four-byte range 0xC–0xF, then an 8-byte access to the range 0x8–0xF will fail, assuming that PMP entry is the highest-priority entry that matches those addresses.
> 
> If a PMP entry matches all bytes of an access, then the L, R, W, and X bits determine whether the access succeeds or fails. If the L bit is clear and the privilege mode of the access is M, the access succeeds. Otherwise, if the L bit is set or the privilege mode of the access is S or U, then the access succeeds only if the R, W, or X bit corresponding to the access type is set.

### The Issue
In our test case, we configure:

- PMP Entry 2: Covers the entire address range (-1), **Unlocked**, **RWX permissions** (L=0, R=1, W=1, X=1, cfg=0x1f).
- PMP Entry 7: Covers the entire address range (-1), **Locked**, **No permissions** (L=1, R=0, W=0, X=0, cfg=0x98).

#### Expected Behavior (Spike)
The memory access (ld) should match PMP Entry 2 first. Since it is Unlocked and we are in M-Mode, the access should be allowed. PMP Entry 7 **should** be ignored due to lower priority.

#### Actual Behavior (CVA6)
CVA6 triggers an `LD_ACCESS_FAULT`. It appears to ignore the match on `PMP Entry 2` and enforces the restrictions of `PMP Entry 7` (Locked, No Access), effectively prioritizing the Locked attribute over the entry index order.

### Steps to Reproduce

1. Initialize environment and enter `M-Mode`.
2. Set pmpaddr2 and pmpaddr7 to -1 (covering full address space).
3. Write pmpcfg0 with 0x98000000001f0000:
  - `PMP[2]`: 0x1f (NAPOT, RWX, Unlocked).
  - `PMP[7]`: 0x98 (TOR, No Access, Locked).

[...]

PR #3490, opened 2026-08-22
Fix PMP TOR matching at eight-byte grain

### LINKED ISSUE #3342: [BUG] CVA6 accepts 8-byte S-mode ld/sd across a TOR PMP boundary
### Code of Conduct

### Bug Description

CVA6 appears to allow S-mode naturally aligned 8-byte scalar load/store operations to access bytes that should be outside the matched PMP region. The issue seems to be tied to PMP TOR boundary handling and CVA6's documented PMP granularity. CVA6 documents that it supports only 8-byte PMP granularity (`G=1`) and that `pmpaddr` bit 0 is forced by the address mode: `0` for `TOR/OFF`, `1` for `NAPOT`.

However, the tests below write a TOR top address whose `pmpaddr[0]` is `1` internally by setting:

```text
pmpaddr0 = (cross_addr + 4) >> 2
```

For `G=1`, this low bit should not create a 4-byte TOR boundary. CVA6 still appears to use the raw internal `pmpaddr` value for PMP matching. Combined with a PMP matcher that checks only the requested start address rather than the full access size or byte mask, this lets an S-mode aligned 8-byte `ld`/`sd` cross into bytes that should be denied.

### Bug Description

The tests run in bare S-mode. They configure two TOR PMP entries:

- `pmpaddr0 = (cross_addr + 4) >> 2`
- `pmpcfg0[0] = TOR | R | W | X`
- `pmpaddr1 = (cross_addr + 8) >> 2`
- `pmpcfg0[1] = TOR` with no R/W/X permissions

Then the tests drop from M-mode to S-mode and execute one naturally aligned 8-byte scalar operation at `cross_addr`:

- `ld` should not be able to leak bytes from the denied upper half.
- `sd` should not be able to modify bytes in the denied upper half.

Each test writes a result code to `tohost`.

### Steps to reproduce

Run from the CVA6 repository:

```sh
cd ./cores/cva6



./work-ver/Variane_testharness tor_partial_match_aligned_ld_leak.elf \
  +debug_disable=1 +ntb_random_seed=1 +tohost_addr=80001000 \
  +elf_file=tor_partial_match_aligned_ld_leak.elf \
  +core_name=cv64a6_imafdc_sv39

./work-ver/Variane_testharness tor_partial_match_aligned_sd_side_effect.elf \
  +debug_disable=1 +ntb_random_seed=1 +tohost_addr=80001000 \
  +elf_file=tor_partial_match_aligned_sd_side_effect.elf \
  +core_name=cv64a6_imafdc_sv39
```

Core behavior of the tests:

```asm
# PMP setup shared by both tests
la t0, cross_addr
addi t0, t0, 4
srli t0, t0, 2
csrw pmpaddr0, t0

la t0, denied_window_end
srli t0, t0, 2
csrw pmpaddr1, t0

li t0, (PMP_TOR << 8) | PMP_RWX_TOR
csrw pmpcfg0, t0

# Load case: should not leak 0x89abcdef from the denied upper half
la t0, cross_addr
ld t1, 0(t0)

# Store case: should not write 0xa5a55a5a to the denied upper half
la t0, cross_addr
li t1, 0xa5a55a5adeadbeef
sd t1, 0(t0)
```

### Expected behavior

CVA6 documents PMP granularity `G=1`, so the attempted 4-byte TOR boundary should not create a valid PMP region boundary. Additionally, the RISC-V privileged specification requires the matching PMP entry to match all bytes of a memory operation.

Therefore:

- `tor_partial_match_aligned_ld_leak.elf` should exit with `tohost = 0`.
- `tor_partial_match_aligned_sd_side_effect.elf` should exit with `tohost = 0`.

### Observed behavior

CVA6 allows the S-mode operations to complete:

```text
tor_partial_match_aligned_ld_leak.elf
  *** FAILED *** (tohost = 94) after 490 cycles

tor_partial_match_aligned_sd_side_effect.elf
  *** FAILED *** (tohost = 94) after 532 cycles
```

Interpretation:

- `ld` returns the full 8-byte value `0x89abcdef01234567`, including the bytes
  that should not be accessible.
- `sd` writes the full 8-byte value `0xa5a55a5adeadbeef`, including bytes that
  should not be writable.

Test Cases: 

[test.zip](https://github.com/user-attachments/files/28734868/test.zip)

KnightGOKU:
CVA6 implements an 8-byte PMP granularity, but TOR matching still uses pmpaddr[0]. This can incorrectly allow an aligned 8-byte access across a TOR boundary. The change ignores that bit for both TOR boundaries and adds a regression test.
This PR fixes #3342.

### DIFF
--- a/core/pmp/src/pmp_entry.sv
+++ b/core/pmp/src/pmp_entry.sv
@@ -29,11 +29,16 @@ module pmp_entry #(
     output logic match_o
 );
   logic [CVA6Cfg.PLEN-1:0] conf_addr_n;
+  logic [CVA6Cfg.PLEN-3:0] tor_conf_addr;
+  logic [CVA6Cfg.PLEN-3:0] tor_conf_addr_prev;
   logic [$clog2(CVA6Cfg.PLEN)-1:0] trail_ones;
   logic [CVA6Cfg.PLEN-1:0] base;
   logic [CVA6Cfg.PLEN-1:0] mask;
   int unsigned size;
   assign conf_addr_n = {2'b11, ~conf_addr_i};
+  // CVA6 implements G=1. Bit 0 does not participate in either TOR boundary.
+  assign tor_conf_addr = {conf_addr_i[CVA6Cfg.PLEN-3:1], 1'b0};
+  assign tor_conf_addr_prev = {conf_addr_prev_i[CVA6Cfg.PLEN-3:1], 1'b0};
   lzc #(
       .WIDTH(CVA6Cfg.PLEN),
       .MODE (1'b0)
@@ -51,15 +56,15 @@ module pmp_entry #(
         size = '0;
         // check that the requested address is in between the two
         // configuration addresses
-        if (addr_i >= ({2'b0, conf_addr_prev_i} << 2) && addr_i < ({2'b0, conf_addr_i} << 2)) begin
+        if (addr_i >= ({2'b0, tor_conf_addr_prev} << 2) && addr_i < ({2'b0, tor_conf_addr} << 2)) begin
           match_o = 1'b1;
         end else match_o = 1'b0;
 
         // synthesis translate_off
         if (match_o == 0) begin
-          assert (addr_i >= ({2'b0, conf_addr_i} << 2) || addr_i < ({2'b0, conf_addr_prev_i} << 2));
+          assert (addr_i >= ({2'b0, tor_conf_addr} << 2) || addr_i < ({2'b0, tor_conf_addr_prev} << 2));
         end else begin
-          assert (addr_i < ({2'b0, conf_addr_i} << 2) && addr_i >= ({2'b0, conf_addr_prev_i} << 2));
+          assert (addr_i < ({2'b0, tor_conf_addr} << 2) && addr_i >= ({2'b0, tor_conf_addr_prev} << 2));
         end
         // synthesis translate_on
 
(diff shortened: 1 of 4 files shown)


codeadpool:
ran this PR's pmp_entry.sv against sail-riscv 0.12 model [65ddde80](https://github.com/riscv/sail-riscv/tree/65ddde80ee2b131bf46c20e6e748343c336c4071) and thought result might be useful for the review. Mitered CVA6's pmp/pmp_entry blocks against a systemverilog translation of sail's PMP decision function, with all relevant inputs symbolic.

Results: 
- S-mode and U-mode: CVA6 and the Sail agree exactly over all symbolic PMP states satisfying CVA6’s WARL constraints: reserved bits are zero, NA4 is excluded, and W=1, R=0 is excluded. Underlying pmpaddr[0] remains symbolic in every address mode.
- M-mode: every remaining mismatch has the pattern already reported in #3177: CVA6 skips a higher-priority unlocked entry and allows a lower-priority locked entry to determine the result. a separate classification assertion proves that any M-mode mismatch falls into that class. Within this equivalence scope, #3177 is the only remaining mismatch class

This result covers cv64a6_imafdc_sv39 (G=1, PLEN=56, eight PMP entries) and naturally aligned accesses fully contained within one eight-byte granule. PMP state is treated as a symbolic snapshot, so the proof does not cover CSR write sequencing orreadback behavior. It covers the pmp/pmp_entry blocks, not full-core integration.

We also checked [sail-riscv](https://github.com/riscv/sail-riscv/tree/27224ccb2290f022e46213c05b3e72e8a9ea635e)[ 0.13.1 (](https://github.com/riscv/sail-riscv/tree/27224ccb2290f022e46213c05b3e72e8a9ea635e)[27224ccb](https://github.com/riscv/sail-riscv/tree/27224ccb2290f022e46213c05b3e72e8a9ea635e)[)](https://github.com/riscv/sail-riscv/tree/27224ccb2290f022e46213c05b3e72e8a9ea635e): pmpReadAddrReg and pmpMatchAddr are byte-identical to 0.12, so the address-matching conclusion is unchanged.

happy to share the harness, generated SystemVerilog, run recipes, and logs if useful.

PR #3492, opened 2026-08-23
Mask reserved mstatus bits

flaviens:
Reserved `mstatus` fields retain writes.

Clear WPRI fields. Clear H fields when H is disabled.

Test: `make -C verif/tb/csr_unit mstatus_wpri_test`


JeanRochCoulon:
Good input @flaviens

PR #3501, opened 2026-08-24
Fix Zcmt table jump target LSB

### LINKED ISSUE #3441: [BUG] Zcmt table jump (cm.jt/cm.jalt) does not clear bit 0 of the target
### Code of Conduct

### CVA6 commit affected

88e810c

## Summary
For `cm.jt` / `cm.jalt`, CVA6 jumps to the raw value read from the jump vector table (JVT) **without clearing bit 0** of that value. Spike clears bit 0 before writing the PC correctly.

## Spike behavior
`cm.jt`/`cm.jalt` in Spike (`riscv/insns/cm_jalt.h`) ends with:

```cpp
if (xlen == 32)
    target = MMU.fetch_jump_table<int32_t>(base + (index << 2));
else
    target = MMU.fetch_jump_table<int64_t>(base + (index << 3));

if (index >= 32)  // cm.jalt
    WRITE_REG(1, npc);

set_pc(target & ~reg_t(1));   // bit 0 of the loaded value is cleared
```

This is the same LSB-clearing that Spike applies to every register-indirect jump (`jalr.h`, `c_jr.h`, `c_jalr.h` all do `& ~reg_t(1)`).

## CVA6 behavior
CVA6 clear bit 0 correctly for its register-indirect jumps: `JALR` masks the LSB in `branch_unit.sv`, and `C.JR`/`C.JALR` are expanded to `JALR` by the compressed decoder, so they go through the same masking. The table jump is the only register-indirect-style jump that misses it.

The reason is that CVA6 turns a table jump into an internal `JAL` (not `JALR`) and computes a PC-relative immediate from the raw table value, never masking bit 0:

`core/zcmt_decoder.sv`
```systemverilog
// TABLE_JUMP state
jump_address_o = $unsigned($signed(req_port_i.data_rdata) - $signed(pc_i)); // raw value, bit 0 kept
// instr_o rewritten to riscv::OpcodeJal
```

`core/branch_unit.sv` — target = `pc + imm = data_rdata`, and bit 0 is only cleared for `JALR`, so the table-jump (`JAL`) path keeps it:
```systemverilog
target_address = $unsigned($signed(jump_base) + $signed(fu_data_i.imm[CVA6Cfg.VLEN-1:0]));
// on a JALR we are supposed to reset the LSB to 0 (according to the specification)
if (fu_data_i.operation == ariane_pkg::JALR) target_address[0] = 1'b0;   // never taken for Zcmt
```

The resulting `target_address[0]` propagates through `frontend.sv` with no masking.

In short: because CVA6 rewrites the table jump as `JAL` instead of `JALR`, it bypasses the exact LSB-clearing logic it already applies (correctly) to `JALR`/`C.JR`/`C.JALR`.

124107157-KV:

## Why this PR is needed

Zcmt table jumps currently use the JVT-loaded value directly when
calculating the jump target.

For `cm.jt` and `cm.jalt`, bit 0 of the value loaded from the JVT must
be cleared before it is used as the architectural target. Since CVA6
expands these table jumps to an internal `JAL`, they do not pass through
the existing `JALR` target-LSB clearing in the branch unit.

As a result, an odd address stored in a JVT entry can cause CVA6 to
jump to the odd address instead of the required address with bit 0
cleared.

## Changes

- Clear bit 0 of the JVT-loaded target in `core/zcmt_decoder.sv`
  before converting the target to the PC-relative offset used by the
  internal `JAL` expansion.
- Add an RV32 directed regression for `cm.jt` with an odd JVT target.
- Add an RV32 directed regression for `cm.jalt` with an odd JVT target.
- Check that the `cm.jalt` target executes and returns through the
  expected link address.
- Register both directed tests in `verif/tests/testlist_issues.yaml`.
- Add both regressions to `verif/regress/issue-tests.sh`.

## Validation

The permanent directed regressions were checked both before and after
the RTL change.

Before the fix:

- `zcmt-jt-target-lsb-rv32`: FAIL
- `zcmt-jalt-target-lsb-rv32`: FAIL

After the fix:

- `zcmt-jt-target-lsb-rv32`: PASS
- `zcmt-jalt-target-lsb-rv32`: PASS
- Existing `zcmt-jvt-index-rv32` regression: PASS

Additional checks:

- `verible-verilog-format --verify core/zcmt_decoder.sv`: PASS
- `bash -n verif/regress/issue-tests.sh`: PASS
- `git diff --check`: clean
- Rebasing against the current `upstream/master` required no changes.
- Working tree is clean after the final regression runs.

## Scope / limitations

This change is limited to Zcmt table-jump target handling.

It does not change the normal `JAL`, `JALR`, or branch-unit target
handling. Bit 0 is cleared specifically on the target loaded from the
JVT before that target is converted to the PC-relative offset used by
the Zcmt internal `JAL` expansion.

No known limitation is introduced by this change.

Fixes #3441

### DIFF
--- a/core/zcmt_decoder.sv
+++ b/core/zcmt_decoder.sv
@@ -97,8 +97,9 @@ module zcmt_decoder #(
       end
       TABLE_JUMP: begin
         if (req_port_i.data_rvalid) begin
-          // save the PC relative Xlen table jump address 
-          jump_address_o = $unsigned($signed(req_port_i.data_rdata) - $signed(pc_i));
+          // Clear bit 0 of the JVT target before converting it to a PC-relative offset.
+          jump_address_o =
+              $unsigned($signed({req_port_i.data_rdata[CVA6Cfg.XLEN-1:1], 1'b0}) - $signed(pc_i));
           if (instr_i[9:2] < 32) begin  // jal pc_offset, x0 for no return stack
             instr_o = {
               20'h0, 5'h0, riscv::OpcodeJal
--- a/verif/regress/issue-tests.sh
+++ b/verif/regress/issue-tests.sh
@@ -51,4 +51,39 @@ if [ "$zcmt_status" -ne 0 ]; then
   return "$zcmt_status" 2>/dev/null || exit "$zcmt_status"
 fi
 
+
+# Check that cm.jt clears bit 0 of the loaded JVT target.
+python3 cva6.py \
+  --testlist=../tests/testlist_issues.yaml \
+  --test zcmt-jt-target-lsb-rv32 \
+  --iss_yaml cva6.yaml \
+  --target hwconfig \
+  --hwconfig_opts="cv32a60x *RVZCMT=1" \
+  --iss=veri-testharness \
+  --linker="../../config/gen_from_riscv_config/cv32a60x/linker/link.ld"
+
+zcmt_jt_lsb_status=$?
+if [ "$zcmt_jt_lsb_status" -ne 0 ]; then
+  echo "Error: Zcmt cm.jt target LSB regression failed"
+  cd ../..
+  return "$zcmt_jt_lsb_status" 2>/dev/null || exit "$zcmt_jt_lsb_status"
+fi
+
+# Check that cm.jalt clears bit 0 of the loaded JVT target.
+python3 cva6.py \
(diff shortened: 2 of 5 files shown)


JeanRochCoulon:
Good design and verification !

PR #3504, opened 2026-08-25
csr_regfile: mask epc[1] when C extension is disabled

### LINKED ISSUE #3316: [BUG] sepc/mepc/vsepc do not mask bit[1] when RVC is disabled (IALIGN=32 configurations)
### Code of Conduct

#### Bug Description

When the C extension is disabled (`CVA6ConfigCExtEn = 0`), the implementation supports only IALIGN=32. Per the RISC-V Privileged Specification, `sepc[1:0]` must always be zero in such configurations. However, the current implementation only forces `bit[0]` to zero on CSR writes and does not handle `bit[1]`
This affects `sepc`, `mepc`, `vsepc`

RISC-V Privileged Specification:
> The low bit of `mepc` (`mepc[0]`) is always zero. On implementations that support only IALIGN=32, the two low bits (`mepc[1:0]`) are always zero.
> The low bit of `sepc` (`sepc[0]`) is always zero. On implementations that support only IALIGN=32, the two low bits (`sepc[1:0]`) are always zero.

#### Buggy Code

**1. CSR write path** ([core/csr_regfile.sv](line 1358):
```systemverilog
riscv::CSR_SEPC:
if (CVA6Cfg.RVS) sepc_d = {csr_wdata[CVA6Cfg.XLEN-1:1], 1'b0};
```
Only `bit[0]` is forced to zero. `bit[1]` from `csr_wdata` is preserved regardless of `CVA6Cfg.RVC`.

**2. Trap write path** ([core/csr_regfile.sv](line 2069):
```systemverilog
sepc_d = {{CVA6Cfg.XLEN - CVA6Cfg.VLEN{pc_i[CVA6Cfg.VLEN-1]}}, pc_i};
```
`pc_i` is stored directly without masking any low bits (neither `bit[0]` nor `bit[1]`).

**3. CSR read path** ([core/csr_regfile.sv](line 511):
```systemverilog
riscv::CSR_SEPC:
if (CVA6Cfg.RVS) csr_rdata = sepc_q;
```
`sepc_q` is returned without any alignment mask. If `bit[1]` was stored as 1, it is read back as 1.

oChunCai:
When the C extension is compiled out (CVA6Cfg.RVC == 0) the core supports only IALIGN=32, so the two low bits of mepc/sepc/vsepc must always read as zero. The CSR write path only forced bit[0] to zero and kept bit[1] from csr_wdata, letting software store an *epc value whose bit[1] == 1. That value is read back verbatim and is also fed directly to MRET/SRET to set the PC, resuming at a 2-mod-4 address that is not a legal instruction address on a machine without compressed instructions.

Gate bit[1] with CVA6Cfg.RVC on the mepc/sepc/vsepc write paths so it stays writable only when RVC is enabled (IALIGN=16) and is forced to zero otherwise. Configurations with C keep bit-identical behaviour. The trap write path is unaffected because pc_i is always 4-byte aligned when IALIGN=32, so the register never stores bit[1] == 1 and the read path is correct transitively.

Priv. spec (mepc/sepc): "On implementations that support only IALIGN=32, the two low bits (mepc[1:0]) are always zero." vsepc is WARL and must hold the same set of values as sepc.

Fixes #3316

### DIFF
--- a/core/csr_regfile.sv
+++ b/core/csr_regfile.sv
@@ -1296,7 +1296,9 @@ module csr_regfile
         if (CVA6Cfg.RVH) vsscratch_d = csr_wdata;
         else update_access_exception = 1'b1;
         riscv::CSR_VSEPC:
-        if (CVA6Cfg.RVH) vsepc_d = {csr_wdata[CVA6Cfg.XLEN-1:1], 1'b0};
+        // bit[1] stays writable only with RVC; without RVC (IALIGN=32) vsepc[1:0] are always zero
+        if (CVA6Cfg.RVH)
+          vsepc_d = {csr_wdata[CVA6Cfg.XLEN-1:2], csr_wdata[1] & CVA6Cfg.RVC, 1'b0};
         else update_access_exception = 1'b1;
         riscv::CSR_VSCAUSE:
         if (CVA6Cfg.RVH) vscause_d = csr_wdata;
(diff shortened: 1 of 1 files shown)


JeanRochCoulon:
Thanks for your contribution @oChunCai !

PR #3507, opened 2026-08-26
fix: preload all ELF loadable segments

### LINKED ISSUE #3351: [BUG]  CVA6 MPRV=S load returns wrong value after PMP config update
### Code of Conduct

### Bug Description

# CVA6 MPRV=S load returns wrong value after PMP config update

## 1. Introduction 

A bare-metal RV64 test configures a PMP TOR region, performs an allowed MPRV=S load, updates `pmpcfg0` with `csrc`, and then performs another MPRV=S load from the same address. The final PMP permission still allows reads, so the second `lw` should be allowed and should return the low 32 bits of the target word, `0x11223344`.

On the local CVA6 test harness, the second load does not trap, but it returns an unexpected value. The self-checking test reaches its `allowed_access_bad_value` failure path and exits with `tohost=16`.

The same `testcase.elf` passes on Spike, Rocket `DefaultConfig`, and SmallBoomConfig. This makes the symptom look specific to CVA6 behavior around a PMP configuration update followed by an MPRV=S load.

## 2. Test case

- Source: `testcase.S`
- ELF: `testcase.elf`

The test performs the following sequence:

1. Starts in M-mode and disables paging.
2. Programs a TOR PMP region covering `target_inside`.
3. Initially configures the target region as RWX using `pmpcfg0 = 0x1f0f00`.
4. Sets `mstatus.MPRV=1` and `mstatus.MPP=S`.
5. Performs a priming `ld` from `target_inside`, which should be allowed.
6. Clears only the execute bit of the target PMP entry using `csrc pmpcfg0, 0x400`, leaving the final target permission as RW using `pmpcfg0 = 0x1f0b00`.
7. Sets `mstatus.MPRV=1` and `mstatus.MPP=S` again.
8. Performs `lw` from `target_inside`.
9. Expects no trap and expects the loaded value to be `0x11223344`.

The observed failure code is `tohost=16`, which is the test's internal code for `allowed_access_bad_value`.

Core assembly excerpt:

```asm
    # Program the PMP range.
    # pmpaddr0 is the lower TOR bound.
    # pmpaddr1 is the upper TOR bound.
    # pmpcfg0 = 0x1f0f00 means:
    #   pmp0: OFF
    #   pmp1: TOR, R=1, W=1, X=1  => initial RWX target region
    #   pmp2: allow remaining memory
    la      t0, pmp_region
    srli    t0, t0, 2
    csrw    pmpaddr0, t0
    la      t0, pmp_region_end
    srli    t0, t0, 2
    csrw    pmpaddr1, t0
    li      t0, -1
    csrw    pmpaddr2, t0
    li      t0, 0x1f0f00
    csrw    pmpcfg0, t0
    fence   rw, rw

    la      s0, target_inside

    # Prime with an allowed MPRV=S load.
    li      t0, 0x1800       # MSTATUS_MPP_MASK
    csrc    mstatus, t0
    li      t0, 0x0800       # MSTATUS_MPP_S
    csrs    mstatus, t0
    li      t0, 0x20000      # MSTATUS_MPRV
    csrs    mstatus, t0
    la      t5, target_inside
    ld      t6, 0(t5)
    li      t0, 0x20000      # MSTATUS_MPRV
    csrc    mstatus, t0

    # Clear only the X bit of pmp1. Final permission remains R=1, W=1.
    li      t0, 0x400
    csrc    pmpcfg0, t0
    fence   rw, rw

    # Probe with another MPRV=S load. This should be allowed.
    li      t0, 0x1800       # MSTATUS_MPP_MASK
    csrc    mstatus, t0
    li      t0, 0x0800       # MSTATUS_MPP_S
    csrs    mstatus, t0
    li      t0, 0x20000      # MSTATUS_MPRV
    csrs    mstatus, t0
    lw      s2, 0(s0)
```

The checker then requires no trap and verifies that the load returned `0x11223344`. If the value is different, it reports `tohost=16`:

```asm
expect_success:
    bnez    t1, fail_allowed_trapped

    la      t0, observed_reg
    ld      t1, 0(t0)
    li      t2, 0x11223344
    bne     t1, t2, fail_allowed_value
    li      a0, 0
    j       finish

fail_allowed_value:
    li      a0, 16
    j       finish
```

The target data is initialized as:

```asm
target_inside:
    .dword 0x5566778811223344
```

Therefore the expected result of `lw s2, 0(s0)` is `0x11223344`.

## 3. How to reproduce

(1) run the Spike reference. 
```sh
./riscv-isa-sim/install/bin/spike --isa=rv64imac_zicsr --priv=msu ./testcase.elf
echo $?
```

(2) run the failing CVA6 test harness. This is expected to fail with `tohost=16`:

```sh
timeout 120 ../../cores/cva6/work-ver/Variane_testharness ./testcase.elf
echo $?
```

```
This emulator compiled with JTAG Remote Bitbang client. To enable, use +jtag_rbb_enable=1.
Listening on port 34943
*** [rvf_tracer] INFO: Loading binary :
*** [rvf_tracer] INFO: tohost_addr: 0000000000000000
*** [rvf_tracer] WARNING: No valid address of 'tohost' (tohost == 0x0000000000000000), termination possible only by timeout or Ctrl-C!

*** FAILED *** (tohost = 16)
./testcase.elf *** FAILED *** (tohost = 16) after 35912 cycles
CPU time used: 3570.19 ms
Wall clock time passed: 3564.14 ms
16
```
TestCases: 

[testcase.zip](https://github.com/user-attachments/files/28982618/testcase.zip)

KnightGOKU:

  ## Why is this PR needed?

  The Verilator testbench only preloaded ELF segments starting at `0x80000000`. When an ELF contained additional loadable segments, data was not copied to the correct memory offset, causing valid loads to return incorrect
  values.

  This change loads every ELF loadable segment within the DRAM range using its actual address and size.

  ## Related issue

  Fixes #3351

### DIFF
--- a/corev_apu/tb/ariane_tb.cpp
+++ b/corev_apu/tb/ariane_tb.cpp
@@ -345,8 +345,15 @@ done_processing:
   size_t mem_size = 0xFFFFFF;
   while(get_section(&addr, &len))
   {
-    if (addr == 0x80000000)
-        read_section_void(addr, (void *) MEM , mem_size);
+    if (addr >= 0x80000000 &&
+        len >= 0 &&
+        static_cast<uint64_t>(addr) - 0x80000000ULL <= mem_size &&
+        static_cast<uint64_t>(len) <=
+            mem_size - (static_cast<uint64_t>(addr) - 0x80000000ULL))
+        read_section_void(addr,
+                          (void *) ((uint8_t *) MEM +
+                                    static_cast<uint64_t>(addr) - 0x80000000ULL),
+                          static_cast<uint64_t>(len));
     if (addr == 0x84000000)
         try {
           read_section_void(addr, (void *) MEM_USER , mem_size);


JeanRochCoulon:
@cainria  What do think about this PR, can you approve ?

corev_apu/tb/ariane_tb.cpp
```diff
@@ -345,8 +345,15 @@ int main(int argc, char **argv) {
...
-        read_section_void(addr, (void *) MEM , mem_size);
+    if (addr >= 0x80000000 &&
+        len >= 0 &&
+        static_cast<uint64_t>(addr) - 0x80000000ULL <= mem_size &&
+        static_cast<uint64_t>(len) <=
+            mem_size - (static_cast<uint64_t>(addr) - 0x80000000ULL))
+        read_section_void(addr,
+                          (void *) ((uint8_t *) MEM +
+                                    static_cast<uint64_t>(addr) - 0x80000000ULL),
+                          static_cast<uint64_t>(len));
```
cainria, line 356: Readability could be improved by introducing named variables such as `uint64_t offset = static_cast<uint64_t>(addr) - 0x80000000ULL` (appears 3 times in the +diff)

corev_apu/tb/ariane_tb.cpp
```diff
@@ -345,8 +345,15 @@ int main(int argc, char **argv) {
   size_t mem_size = 0xFFFFFF;
   while(get_section(&addr, &len))
   {
-    if (addr == 0x80000000)
-        read_section_void(addr, (void *) MEM , mem_size);
+    if (addr >= 0x80000000 &&
+        len >= 0 &&
+        static_cast<uint64_t>(addr) - 0x80000000ULL <= mem_size &&
+        static_cast<uint64_t>(len) <=
+            mem_size - (static_cast<uint64_t>(addr) - 0x80000000ULL))
```
cainria:
```
a) offset <= mem_size
b) len <= mem_size - offset
```

I think it would be easier to understand condition b) if it were worded as:

```
b) len + offset <= mem_size  // EDIT: my bad, this is not equivalent to b) above when there is an overflow
```

Which is, the end of the copied data is within the memory.

Then, condition a) seems redundant because `len` is unsigned so `len >= 0` (condition checked 2 lines before the cast), then `len + offset >= offset` (unless there is an overflow, which hopefully should not occur with 64-bit addresses?)

So condition b) `len + offset <= mem_size` with `len + offset >= offset` implies `offset <= len + offset <= mem_size` which is condtion a). // EDIT: yes, but not with my suggested rewording for b)

Consequently, condition a is useless because `(a => b) => (a && b) === b`. // EDIT: yes, but not with my suggested rewording for b)

The assumption here is that `len + offset` does not overflow. Is it guaranteed?
If the overflow is possible here, I think it should be easier to have an explicit check for it.

cainria (changes requested):

corev_apu/tb/ariane_tb.cpp
KnightGOKU: Agreed. I’ll introduce named variables for the section address, section size, and DRAM offset, then reuse them for the bounds check and destination pointer.

corev_apu/tb/ariane_tb.cpp
KnightGOKU: Thanks. get_section returns addr and len as signed long long values, so I’ll keep checking that they are non-negative before converting them to uint64_t. I’ll use offset <= mem_size and size <= mem_size - offset for the bounds check.

corev_apu/tb/ariane_tb.cpp
```diff
@@ -343,10 +343,25 @@ int main(int argc, char **argv) {
...
-        read_section_void(addr, (void *) MEM , mem_size);
+    if (addr >= 0 && len >= 0) {
+      const uint64_t section_addr = static_cast<uint64_t>(addr);
+      const uint64_t section_size = static_cast<uint64_t>(len);
+
+      if (section_addr >= dram_base) {
+        const uint64_t offset = section_addr - dram_base;
+
+        // Check the range without computing offset + section_size, which
+        // could overflow before it is compared with dram_size.
```
cainria: Did you verify that the overflow was possible as suggested in my last comment?

- `long long` should be a 64-bit signed number, whose range is -2⁶³−2⁶³-1
- it is positive (`if` condition line 350) so the range is 0−2⁶³-1
- so `section_addr + section_size` is in range 0−2⁶⁴-2 which does not overflow `uint64_t` whose range is 0−2⁶⁴-1
- `offset = section_addr - dram_base` is also positive because `section_addr` ≥ `dram_base` (`if` line 354) so the same argument holds for `offset + section_size`

corev_apu/tb/ariane_tb.cpp
cainria: ~~If we consider that `long long` is 64 bits (**which is true for all common data models** https://cppreference.net/cpp/language/types.html but is not guaranteed to be always true), the overflow is not possible and condition a) is useless.~~

~~If we consider that `long long` can be greater than 64 bits, then condition a) does not detect all overflows so the comment is misleading. We could have `offset + section_size` < `offset` ≤ `dram_size`, and `offset + section_size` < `offset` is the overflow here.
The wanted guarantee is `offset` ≤ `offset + section_size` ≤ `dram_size`, and you only check that `offset` ≤ `dram_size` and `offset + section_size` ≤ `dram_size` (even if it is equivalently worded as `section_size` ≤ `dram_size - offset`), which is not enough.~~
EDIT: not relevant since the equivalence for my suggestion of b is wrong

Note: the no-overflow guarantee could also be provided upstream, for instance by `get_section`, which sets `addr` and `len`.

Also, I guess the overflow would trigger a segfault in `read_section_void` anyway. So here we are maybe choosing between crashing the program in really rare cases or silently skipping a section, which may not be relevant either.

KnightGOKU: Thanks, that makes sense. I propose checking addr and len in their original long long type, then validating offset <= dram_size and len <= dram_size - offset before converting them. This avoids overflow while ensuring the section fits. Does this address your concern, or would you prefer the check in get_section?

cainria: My bad, I didn't notice that as `offset <= dram_size`, `dram_size - offset` does not overflow. qed.

Sorry!

LGTM

cainria (approved):

PR #3509, opened 2026-08-27
RTD: Update OS version to ubuntu-lts-latest

MikeOpenHWGroup:
Automatic re-generation of CVA6 documentation on ReadTheDocs is failing because of Ubuntu-20.04 is no longer supported.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -10,7 +10,7 @@ submodules:
   recursive: true
 
 build:
-  os: "ubuntu-lts-latest"
+  os: "ubuntu-26.04"
   tools:
     python: "3.9"
     nodejs: "20"
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -10,7 +10,7 @@ submodules:
   recursive: true
 
 build:
-  os: "ubuntu-20.04"
+  os: "ubuntu-lts-latest"
   tools:
     python: "3.9"
     nodejs: "20"


cainria (approved):

ASintzoff:
should we not use a specific version like `ubuntu-26.04`? Doing so will avoid any possible issue when `ubuntu-lts-latest`will be switching to `ubuntu-28.04`.

This reminds me of https://github.com/openhwgroup/cva6/pull/3443 by @cainria where Verible version is now explicitly specified

cainria:
I agree with @ASintzoff but my opinion on the OS version freeze is weaker than on the Verible version freeze

The issue with Verible was that changing the formatting impacts all files, so all contributors have to upgrade simultaneously. For the Read the Docs OS, I think the issues would be mainly in the list of packages to install in the runner, which do not impact contributors.

Apart from that, freezing is better for reproducibility, but `ubuntu-26.04` will eventually be deprecated in the long-term so we will need to change the version explicitly (preferably before the jobs fail next time :wink:)

.readthedocs.yaml
```diff
@@ -10,7 +10,7 @@ submodules:
   recursive: true
 
 build:
-  os: "ubuntu-20.04"
+  os: "ubuntu-lts-latest"
```
cainria: Following @ASintzoff suggestion:

```suggestion
  os: "ubuntu-26.04"
```

And for reference, here are the docs: https://docs.readthedocs.com/platform/stable/config-file/v2.html#build-os

JeanRochCoulon:
Thanks gentlemen for the discussion: @MikeOpenHWGroup for the fix, and @ASintzoff / @cainria for the fine tuning. I will apply the suggestion and merge.

Issue #3519, opened 2026-09-02
[BUG] <wrong instruction count in CSRs in 65x - master candidate>


AngelaGonzalezMarino:
### Code of Conduct

- [x] I have searched the existing bug issues.
- [x] I am a human engaging in an interpersonal interaction. During this interaction, my words are my own and are not generated. If relevant, I provide links to my sources.

### CVA6 commit affected

7878f034aff9db434ffd1565f5e8a163cf54acfc

### Bug Description

In master candidate, 65x benchmarks report wrong instruction count

udayaa-init:
When the instruction is cancelled (branch miss), the instruction is dropped. The commit stage gives an acknowledgment to the dropped instruction to the Scoreboard. This(received ack) causes the machine instruction counter to increment even when the instruction is cancelled.

cainria:
@udayaa-init Could you please elaborate?

The machine instruction counter should be in `csr_regfile`, not `scoreboard`, and the increment should be disabled for dropped instructions:

https://github.com/openhwgroup/cva6/blob/40ac92ccc09534fb0c6b139a613f65868af9f499/core/cva6.sv#L1191

cainria:
Ok I was looking at the bad revision.

https://github.com/openhwgroup/cva6/blob/7878f034aff9db434ffd1565f5e8a163cf54acfc/core/cva6_pipeline.sv#L1003

The issue seems to be introduced by https://github.com/openhwgroup/cva6/pull/2855

I guess that the added condition `CVA6Cfg.SpeculativeSb` is cast into either 2'b00 or 2'b01 so the bitwise `&` always ignores the MSB.

Can I open a PR on `master_candidate` @AngelaGonzalezMarino ?

AngelaGonzalezMarino:
yes please!! @cainria 

@udayaa-init can you confirm that this would fix the issue?

cainria:
It should be fixed on `master_candidate` now, @AngelaGonzalezMarino can you confirm please?

udayaa-init:
@cainria Yes, it is fixed. We can close the issue.

cainria:
Thanks!

PR #2855, opened 2025-03-21
To improve CC

JeanRochCoulon:
Improve CC by solving for loop in csr_regfile.sv and conditioning drop signal by SpeculativeSb in cva6_pipeline.sv

PR #3522, opened 2026-09-02
fix: do not count dropped instrs as committed

cainria:
This PR fixes bug #3519 introduced by #2855


cainria:
Cc @AngelaGonzalezMarino

JeanRochCoulon:
Good catch !

PR #3526, opened 2026-09-04
config_pkg: reject RVZCMP+RVD configs at check_cfg time

### LINKED ISSUE #3483: Requirements ISA-60 and ISA-150 are mutually exclusive
### Code of Conduct

### CVA6 commit affected

6cb2001 (head of master branch as of 2026-08-19)

### Bug Description

Requirement ISA‑60 states that the CVA6 can be configured to support the 'D' ISA and requirement ISA-150 states optional support for Zcmp.  [Section 28.1.9](https://docs.riscv.org/reference/isa/v20260120/unpriv/zc.html#Zcmp) of the Unpriv spec states that Zcmp is incompatible with Zcd (compressed double precision floating point loads and stores) which is part of 'D'.

It would be highly useful if there was a compile-time or run-time assertion to catch this. Something like
```systemverilog
initial begin
    assert (!(CVA6Cfg.RVZCMP && CVA6Cfg.RVD)) 
        else $fatal(1, "[CVA6 Config Error] Zcmp cannot be enabled alongside Double-Precision (D) floating point due to opcode collisions.");
end
```

Security-HC:
Fixes #3483. 
RVZCMP (Zcmp) and RVD (D) collide at the opcode level: Zcmp reuses the encodings of `c.fsdsp`/`c.fsd`, which are part of Zcd and therefore implied by D. A core implementing D must not implement Zcmp. `check_cfg()` had no build-time guard against enabling both.

This adds a `$fatal` in the existing `check_cfg()` sanity-check function, following the same pattern as the adjacent `RVZCMT`/`MmuPresent` check one line above. It sits inside the `// pragma translate_off` block, so it is elaboration-only and strips out at synthesis .

- Invalid config (`CVA6ConfigZcmpExtEn=1`, default `RVD=1`): elaboration now
  aborts with
  `%Error: config_pkg.sv:485: Assertion failed in check_cfg: [config] RVZCMP (Zcmp) and RVD (D) are mutually exclusive`.
- Default valid config `cv64a6_imafdc_sv39` (`RVD=1, RVZCMP=0`): elaborates
  and runs unchanged.

Patch drafted with LLM assistance; reviewed, spec-checked and Verilator-verified by me.

### DIFF
--- a/core/include/config_pkg.sv
+++ b/core/include/config_pkg.sv
@@ -486,10 +486,11 @@ package config_pkg;
     assert (!(Cfg.RVS && !Cfg.SoftwareInterruptEn));
     assert (!(Cfg.RVH && !Cfg.SoftwareInterruptEn));
     assert (!(Cfg.RVZCMT && ~Cfg.MmuPresent));
-    // Zcmp is incompatible with Zcd (compressed double-precision FP loads/stores),
-    // which is part of D, due to opcode collisions.
-    if (Cfg.RVZCMP && Cfg.RVD) begin
-      $fatal(1, "[config] RVZCMP (Zcmp) and RVD (D) are mutually exclusive");
+    // Zcmp is incompatible with Zcd (compressed double-precision FP loads/stores,
+    // automatically present whenever both C and D are implemented), due to
+    // opcode collisions.
+    if (Cfg.RVZCMP && Cfg.RVC && Cfg.RVD) begin
+      $fatal(1, "[config] RVZCMP (Zcmp) and RVC+RVD (Zcd = C+D) are mutually exclusive");
     end
     // pragma translate_on
   endfunction
--- a/core/include/config_pkg.sv
+++ b/core/include/config_pkg.sv
@@ -486,6 +486,11 @@ package config_pkg;
     assert (!(Cfg.RVS && !Cfg.SoftwareInterruptEn));
     assert (!(Cfg.RVH && !Cfg.SoftwareInterruptEn));
     assert (!(Cfg.RVZCMT && ~Cfg.MmuPresent));
+    // Zcmp is incompatible with Zcd (compressed double-precision FP loads/stores),
+    // which is part of D, due to opcode collisions.
+    if (Cfg.RVZCMP && Cfg.RVD) begin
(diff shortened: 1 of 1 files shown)


JeanRochCoulon:
Hello @Security-HC 
Why creating an assert between Zcmp and RVD,
According to me, it is more between Zcmp and Zcd, can you elaborate ?

Security-HC:
> Hello @Security-HC Why creating an assert between Zcmp and RVD, According to me, it is more between Zcmp and Zcd, can you elaborate ?

Hello,I think the check was broader than the actual collision.

I  check the regulations.The Zc spec states Zcmp is incompatible with Zcd, and Zcd is defined to be automatically present only when both C and D are implemented (Zcd = C && D), not D alone.  I've narrowed the check to `RVZCMP && RVC && RVD`.

Re-verified on a fresh Verilator build:
- RVZCMP=1, RVC=1 (default), RVD=1 (default): still fires `$fatal` (same as before).
- RVZCMP=1, RVC=0, RVD=1: no longer fires — an RVD-only, no-C config has no compressed opcode space at all, so there's no actual Zcd to collide with; the old RVD-only check was a false positive here.

JeanRochCoulon:
I am not aligned: Zcd can only be enabled when C and F are supported. But you can have C and F without supporting Zcd.

Security-HC:
> I am not aligned: Zcd can only be enabled when C and F are supported. But you can have C and F without supporting Zcd.

Hello @JeanRochCoulon,

I think there may be a mix-up between Zcf and Zcd — Zcf (single-precision,RV32-only) depends on C+F, but Zcd (double-precision, the one relevant to this assert) depends on C+D, not C+F.
[https://github.com/riscv/riscv-isa-manual/blob/777d5624dfdc85282af845ef4c88c3ad4cd8b827/src/unpriv/c-st-ext.adoc]#L12-L18url))

The ext:c[] extension depends upon the extlink:zca[] extension.
...
If the ext:d[] extension is present, the ext:c[] extension additionally
depends upon the extlink:zcd[] extension.
So C+D unconditionally implies Zcd — there's no valid RISC-V configuration with C+D present but Zcd absent, en, I think which is why CVA6Cfg has no separate Zcd bit to test.

JeanRochCoulon:
I agree !    Thanks for the discussion

PR #3542, opened 2026-09-08
csr_regfile: mask htval and mtval2 WARL writes

### LINKED ISSUE #3537: [BUG] csr_regfile: mtval2 and htval accept and read back arbitrary values, violating their WARL legal set
### Code of Conduct

### CVA6 commit affected

3496afa

### Bug Description

`core/csr_regfile.sv:1660-1662` (`CSR_MTVAL2: if (RVH) mtval2_d = csr_wdata;`) and `:1467-1469` (`CSR_HTVAL: htval_d = csr_wdata;`) store the full XLEN write value, and the read paths return the stored register unchanged.

Priv. §5.2.9 / §3.1.x: "htval is a WARL register that must be able to hold zero and may be capable of holding only an arbitrary subset of other 2-bit-shifted guest physical addresses"; mtval2 is defined the same way. With Sv39x4 (GPLEN = 41) a 2-bit-shifted GPA has at most 39 significant bits, so e.g. a write of all-ones must not read back as all-ones; CVA6 returns it verbatim on `cv64a6_imafdch_sv39`.

Low severity (software normally only reads these registers), but it makes the registers non-WARL. Fix: mask writes to `[GPLEN-3:0]` (or hold only the bits the hardware itself ever writes).

124107157-KV:

## Why

The `htval` and `mtval2` CSR software write paths currently store the complete XLEN-wide `csr_wdata` value.

Both CSRs are WARL registers and represent a guest physical address shifted right by two bits. For the RV64 Sv39x4 configuration, `GPLEN` is 41 bits, which means only 39 bits of the shifted guest physical address are implemented.

As a result, writing an arbitrary XLEN-wide value such as all ones is currently preserved and read back unchanged instead of being restricted to the implemented legal value range.

## What

This change:

- Restricts software writes to `htval` and `mtval2` to `csr_wdata[CVA6Cfg.GPLEN-3:0]`.
- Zero-extends the retained shifted guest-physical-address bits to XLEN.
- Keeps the existing hardware trap write paths unchanged.
- Adds a directed RV64 regression covering both `htval` and `mtval2`.
- Checks that values outside the implemented shifted-GPA width are discarded.
- Checks that both WARL registers can still hold zero.

For `cv64a6_imafdch_sv39`, `GPLEN = 41`, so an all-ones CSR write is legalized to:

`0x0000007fffffffff`

The same implementation is behavior-neutral for RV32, where `GPLEN = 34` and the shifted guest physical address occupies the full 32-bit CSR width.

## Verification

Tested with:

`cv64a6_imafdch_sv39`

Directed regression:

`htval-mtval2-warl-rv64`

Before the fix:

`*** FAILED *** (tohost = 1) after 972 cycles`

After the fix:

`*** SUCCESS *** (tohost = 0) after 977 cycles`

The modified RTL was formatted using `verible-verilog-format`.

`git diff --check` reports no whitespace errors.

Fixes #3537

### DIFF
--- a/core/csr_regfile.sv
+++ b/core/csr_regfile.sv
@@ -1536,7 +1536,7 @@ module csr_regfile
         end
         riscv::CSR_HTVAL: begin
           if (CVA6Cfg.RVH) begin
-            htval_d = csr_wdata;
+            htval_d = {{CVA6Cfg.XLEN - CVA6Cfg.GPLEN + 2{1'b0}}, csr_wdata[CVA6Cfg.GPLEN-3:0]};
           end else begin
             update_access_exception = 1'b1;
           end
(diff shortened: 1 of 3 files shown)


JeanRochCoulon:
LGTM

@oChunCai As issue author, do you approve the PR?

PR #3543, opened 2026-09-08
Gate Zcmp IDLE transitions on issue ack

### LINKED ISSUE #3535: [BUG] Zcmp macro_decoder leaves IDLE without issue_ack for single-register cm.push/cm.pop/cm.popret/cm.popretz, cm.mva01s/cm.mvsa01 and rlist=15: first micro-op is dropped when issue stalls
### Code of Conduct

### CVA6 commit affected

3496afa

### Configuration
RVC=1, RVZCMP=1 (RV32 and RV64 alike)

### Bug Description
core/macro_decoder.sv expands each Zcmp instruction into a sequence of micro-ops and
advances its FSM on `issue_ack_i`. In IDLE the generic transition is ack-gated:

    277:  state_d = issue_ack_i ? INIT : IDLE;

but several branches later in the same IDLE block overwrite `state_d` unconditionally,
without looking at `issue_ack_i`:

    314/322:  cm.mvsa01 / cm.mva01s              -> state_d = MOVE;
    339:      cm.push  with reg_numbers == 1     -> state_d = PUSH_ADDI;
    386/389:  cm.pop / cm.popret with 1 register -> state_d = PUSH_ADDI;
              cm.popretz with 1 register         -> state_d = POPRETZ_1;
    367/419:  rlist = 15 (12 registers)          -> state_d = PUSH_POP_INSTR_2;

If the issue stage does not accept the first micro-op in the cycle the macro is presented
(scoreboard full, LSU busy, operand not ready), the FSM has already moved on. The next
state emits only the *following* micro-op (`addi sp,sp,imm` for PUSH_ADDI, the second
`mv` for MOVE, ...), so the first store/load/move is never issued.

Example: `cm.pop {ra}, 16` (rlist=4, spimm=0) presented while issue stalls for one cycle
retires as `addi sp, sp, 16` only; `lw/ld ra, ...(sp)` is lost and `ra` keeps its old
value. For `cm.push {ra}, -16` the store of `ra` is lost while `sp` is still decremented.

All other rlist values go through INIT and are correctly ack-gated, so only the listed
encodings are affected.

### Expected behavior
Zc spec 1.0.0, §"PUSH/POP register instructions": cm.push stores every register of the
list and adjusts sp; cm.pop/cm.popret/cm.popretz load every register of the list;
cm.mva01s/cm.mvsa01 move both registers. The expansion must not depend on whether the
issue stage accepts the first micro-op in the first cycle.

### Fix
Gate the overriding transitions on `issue_ack_i`, i.e. only leave IDLE when the first
micro-op has been accepted (same as the generic `issue_ack_i ? INIT : IDLE`).

Related but distinct: #3408 (stale state across traps), #3465 (reserved rlist).

124107157-KV:

## Description

The Zcmp macro decoder contains special transitions from IDLE that can advance the FSM without waiting for `issue_ack_i`.

If the issue stage stalls while the first expanded micro-op is being presented, the decoder can move to the next state before the first micro-op is accepted. This can drop the first load, store, or move operation.

The affected cases include:

- `cm.mvsa01`
- `cm.mva01s`
- single-register `cm.push`
- single-register `cm.pop`
- single-register `cm.popret`
- single-register `cm.popretz`
- rlist=15 `cm.push`
- rlist=15 `cm.pop`

This change gates the affected state transitions with `issue_ack_i` so that the decoder remains in IDLE until the first expanded micro-op has been accepted.

## Changes

- Gate the `cm.mvsa01` transition to `MOVE` with `issue_ack_i`.
- Gate the `cm.mva01s` transition to `MOVE` with `issue_ack_i`.
- Gate single-register `cm.push` transition to `PUSH_ADDI`.
- Gate single-register `cm.pop` and `cm.popret` transition to `PUSH_ADDI`.
- Gate single-register `cm.popretz` transition to `POPRETZ_1`.
- Gate rlist=15 push/pop transition to `PUSH_POP_INSTR_2`.
- Add a focused macro decoder unit regression.
- Run the regression from `verif/regress/cv32a6_tests.sh`.

## Verification

A focused unit test was added in:

    verif/tb/macro_decoder_unit/

The test holds `issue_ack_i` low while the first expanded micro-op is presented and verifies that:

- the decoder remains in IDLE;
- the first micro-op remains stable;
- the FSM advances only after `issue_ack_i` is asserted.

The following cases are covered:

- `cm.mvsa01`
- `cm.mva01s`
- `cm.push {ra}, -16`
- `cm.pop {ra}, 16`
- `cm.popret {ra}, 16`
- `cm.popretz {ra}, 16`
- `cm.push {ra, s0-s11}, -64`
- `cm.pop {ra, s0-s11}, 64`

### Before the fix

The regression fails on the original RTL with:

    cm.mvsa01 left IDLE without issue_ack_i (state=4)

The baseline unit test returns a non-zero status.

### After the fix

The regression passes all affected cases:

    PASS: cm.mvsa01
    PASS: cm.mva01s
    PASS: cm.push {ra}, -16
    PASS: cm.pop {ra}, 16
    PASS: cm.popret {ra}, 16
    PASS: cm.popretz {ra}, 16
    PASS: cm.push {ra, s0-s11}, -64
    PASS: cm.pop {ra, s0-s11}, 64
    PASS: macro decoder holds every affected first micro-op until issue_ack_i

Final unit-test result:

    macro_decoder_unit_rc=0

Additional checks performed:

    verible-verilog-format
    git diff --check
    git diff --cached --check
    bash -n verif/regress/cv32a6_tests.sh

The branch was also checked against the latest `upstream/master`.

Fixes #3535

## Limitations

None known.

### DIFF
--- a/core/macro_decoder.sv
+++ b/core/macro_decoder.sv
@@ -321,7 +321,9 @@ module macro_decoder #(
             is_double_rd_macro_instr_o = 1;
             // addi xreg1, a0, 0
             instr_o_reg = {12'h0, 5'hA, 3'h0, xreg1, riscv::OpcodeOpImm};
-            state_d = MOVE;
+            if (issue_ack_i) begin
+              state_d = MOVE;
+            end
           end
 
           if (macro_instr_type == MVA01S) begin
@@ -329,7 +331,9 @@ module macro_decoder #(
             is_double_rd_macro_instr_o = 1;
             // addi a0, xreg1, 0
             instr_o_reg = {12'h0, xreg1, 3'h0, 5'hA, riscv::OpcodeOpImm};
-            state_d = MOVE;
+            if (issue_ack_i) begin
+              state_d = MOVE;
+            end
           end
 
           if (macro_instr_type == PUSH) begin
@@ -346,7 +350,9 @@ module macro_decoder #(
                   7'b1111111, 5'h1, 5'h2, 3'h2, 5'b11100, riscv::OpcodeStore
                 };  // sw store_reg, -4(sp)
               end
-              state_d = PUSH_ADDI;
+              if (issue_ack_i) begin
+                state_d = PUSH_ADDI;
+              end
             end
 
             if (reg_numbers == 4'b0010) begin
@@ -373,7 +379,7 @@ module macro_decoder #(
                 instr_o_reg = {7'b1111111, store_reg, 5'h2, 3'h2, 5'b11100, riscv::OpcodeStore};
               end
 
-              if (reg_numbers == 12) begin
+              if (reg_numbers == 12 && issue_ack_i) begin
                 state_d = PUSH_POP_INSTR_2;
               end
             end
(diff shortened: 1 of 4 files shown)


yanicasa:
It's surprising the bug didn't pop up sooner; any idea why it was hidden? But it looks good to me!

Issue #3408, opened 2026-07-23
[BUG] Cross-context Zcmp expansion-state poisoning occurs when the macro decoder retains stale expansion state after a trap, corrupting subsequent privileged execution.


zxz2004626:
### Code of Conduct

- [x] I have searched the existing bug issues.
- [x] I am a human engaging in an interpersonal interaction. During this interaction, my words are my own and are not generated. If relevant, I provide links to my sources.

### Bug Description

# Bug Description

When RVZCMP is enabled, CVA6 expands each Zcmp macro instruction into multiple internal micro-operations.

If an interrupt, exception, or debug entry flushes the pipeline while a Zcmp instruction is being expanded, the pipeline state is discarded, but the state held inside macro_decoder is not reset. After returning from the trap, the interrupted Zcmp instruction is fetched again, but expansion resumes from the stale internal state instead of restarting from its first micro-operation.

Meanwhile, This state is a shared hart-level execution resource and is not associated with a privilege level, process, or architectural context. An attacker capable of executing a chosen Zcmp instruction and inducing or aligning a trap during its expansion can therefore leave attacker-influenced state inside the decoder. The stale state may subsequently be consumed by a supervisor or machine-mode trap handler or any later Zcmp instruction executed before the stale state is cleared. This turns the issue from a local instruction-replay error into a potential cross-context state-poisoning vulnerability.

Affected version:
Commit: `a3dc2c5e7835`
Configuration: `RVC=1`, `RVZCMP=1`
The FSM implementation is shared by RV32 and RV64 configurations.

Relevant RTL:
[core/macro_decoder.sv](https://github.com/openhwgroup/cva6/blob/master/core/macro_decoder.sv)

# Explanation

The RISC-V `ZCmp` extension reduces code size by replacing common function prologue and epilogue sequences with compressed instructions such as `cm.push`, `cm.pop`, `cm.popret`, and `cm.popretz`. These instructions can save or restore `ra` and multiple callee-saved registers while adjusting the stack pointer, combining several loads, stores, and arithmetic operations into a compact encoding. We take `cm.push` as an example.

```asm
cm.push {ra, s0-s11}, -112
```

[...]

PR #3546, opened 2026-09-08
Uga fpetrot 20260908

fpetrot:
- [ x] I have searched for similar pull requests
  There is one PR (Gate Zcmp IDLE transitions on issue) that is fixing #3543  but I believe it solves neither the 64-bit issue nor simplifies logic, as this PR does. 

This PR simplifies the support of the zcmp insns by factoring the code and removing a lot of logic that was introduced by duplication. It also works for 64-bit processors, by using the XLEN and not 4 to compute offsets, so clearly the current state doesn't work with 64-bit machines. This PR also provides more  thorough tests (using xlen parameterization) to check that the insns behave as expected, for both 32 and 64 bit processors.

A side effect of this PR is that it fixes issue #3543

### DIFF
--- a/core/compressed_decoder.sv
+++ b/core/compressed_decoder.sv
@@ -858,7 +858,7 @@ module compressed_decoder #(
           end
 
           riscv::OpcodeC2Fsdsp: begin
-            if (CVA6Cfg.FpPresent) begin
+            if (CVA6Cfg.RVD && CVA6Cfg.RVC) begin
               // c.fsdsp -> fsd rs2, imm(x2)
               instr_o = {
                 3'b0,
--- a/core/include/config_pkg.sv
+++ b/core/include/config_pkg.sv
@@ -208,8 +208,8 @@ package config_pkg;
     // DcacheInvalidateOnFlush causes dcache to also be invalidated when flushed
     // tradeoff between coherence and efficiency, depending on remaining configuration:
 
-    // DcacheFlushOnFenceI is required for write-back caches - otherwise, 
-    // no way to reliably write instruction memory with store instructions, 
+    // DcacheFlushOnFenceI is required for write-back caches - otherwise,
+    // no way to reliably write instruction memory with store instructions,
     // as data and instruction cache are currently not coherent
     // DcacheFlushOnFence is required for write-back caches to ensure coherency
     // with other harts or DMA devices --> a fence forces all stores to commit to memory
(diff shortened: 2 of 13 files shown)


JeanRochCoulon:
@124107157-KV can you review this PR ?

124107157-KV:
> @124107157-KV can you review this PR ?

Thanks for the broader Zcmp refactor. The XLEN-based cleanup, rlist handling, and `issue_ack_i` gating are good improvements, and I understand this PR is intended to cover #3535 while superseding the narrower fix in #3543.

Before merging, I think the following issues should be addressed:

* `cm.popretz` currently skips `PUSH_ADDI`, so `sp` is not restored before the final `jalr`. Please restore the stack adjustment and extend the test to verify both final `sp` restoration and `a0 == 0`.
* `util.h` removes the local `static_assert` definition while `syscalls.c` still uses it.
* Since this PR is intended to replace #3543, please retain equivalent directed regression coverage for #3535. In particular, the test should explicitly hold `issue_ack_i` low on the first expanded micro-op and verify that the decoder remains in `IDLE` until acknowledgement. The current Tier-1 configurations have `RVZCMP` disabled, and the newly added RV32/RV64 Zcmp tests are not integrated into the normal regression flow.
* The refactored `INIT` state no longer qualifies its transition with `is_macro_instr_i`. Given the stale expansion-state problem tracked in #3408, please preserve this guard or verify why removing it is safe.
* Please also fix the current Verible formatting failure and clean up the Zcmp test stack alignment (`.align 16` on RISC-V does not mean 16-byte alignment).

Also, the PR description should reference #3535 as the issue being fixed; #3543 is the related PR.

With these points addressed, the broader refactor looks like a worthwhile improvement.

fpetrot:
Thanks for you feedback @124107157-KV, that led to even more simplifications.

I'll respin soon, with the following modifications:
- `cm.popretz` with proper stack restoration, sorry for having missed that in the first place.
- reverted `util.h`, I didn't get into why there is a duplication of what stands in the `common` directory, but this is infrastructure so I did not touch that.
- my understanding is that the `INIT` state does not need to check against `is_macro_instr_i`, because by looking at `compressed_decoder.sv` we see (line 875) that this signal is held until we allow a new instruction to enter the decoder by stopping stalling the fetch (`fetch_stall = 0`), which occurs only in IDLE (Mealy signal then), PUSH_ADD and POPRETZ_1 states in which we enter through INIT and that lead to IDLE.
- used verible to reformat the source code
- complexify and corrected the asm tests (hint: `sp` was set at the bottom of the stack and not its top, having the `cm.push` insns crush the static data, ....)
- the hard part for me, as I am new to the cva6 infrastructure, is to use the testbench file you propose for verification properly.

Cheers.

JeanRochCoulon:
To fix Verible, the command to be executed on the RTL is present at the bottom of CONTRIBUTING.md file

Issue #3545, opened 2026-09-08
[BUG] frontend drops the RAS call-push / return-pop when a call or return in the upper fetch slot follows a not-taken branch in the lower slot


Misbahud-Din:
### Code of Conduct

- [x] I have searched the existing bug issues.
- [x] I am a human engaging in an interpersonal interaction. During this interaction, my words are my own and are not generated. If relevant, I provide links to my sources.

### CVA6 commit affected

6348e9e

### Bug Description

In `frontend`, the prediction `always_comb` computes `ras_push`/`ras_pop`/`ras_update` as single scalars in a loop that runs the fetch slots high-index down to slot 0. The branch arm clears `ras_push`/`ras_pop` **unconditionally**, including when the branch is not taken. A not-taken branch falls through, so the upper slot behind it is on-path and consumed. Because slot 0 is processed last, its not-taken arm overwrites the push/pop the upper call/return just set.

`predict_address` is unaffected and the redirect this cycle is correct. Only the RAS update is lost, so a dropped push mispredicts the eventual `ret`, and a dropped pop corrupts the next return. Reachable on any `RVC` config (`INSTR_PER_FETCH = 2`, e.g. cv32a6_imac_sv32).

### Buggy Code

`core/frontend/frontend.sv`, prediction loop (slot 1 then slot 0):

```verilog
4'b1000: begin            // branch arm
  ras_pop  = 1'b0;        // cleared even when the branch is NOT taken
  ras_push = 1'b0;
  // ... compute taken_rvi_cf[i] / taken_rvc_cf[i] ...
  if (taken_rvi_cf[i] || taken_rvc_cf[i]) cf_type[i] = ariane_pkg::Branch;
end
// ...
if (is_call[i]) begin     // upper slot set ras_push here; slot 0's arm clobbers it
  ras_push   = instr_queue_consumed[i];
  ras_update = addr[i] + (rvc_call[i] ? 2 : 4);
end
```

### Reproduction

Both cases fetch two compressed instructions in one cycle: the lower slot a not-taken branch, the upper slot a consumed call/return.

[...]

PR #3547, opened 2026-09-08
Preserve RAS push/pop for an upper-slot call/return behind a not-taken branch

Misbahud-Din:
### Description

In `frontend`, the prediction `always_comb` computes `ras_push`/`ras_pop`/`ras_update` as single scalars in a loop that runs the fetch slots from the highest index down to slot 0, so the lowest slot's write wins. The branch arm (`4'b1000`) cleared `ras_push`/`ras_pop` unconditionally.

When the lower fetch slot is a not-taken conditional branch, it falls through, so the upper slot behind it is on-path and consumed. Because slot 0 is processed last, its not-taken branch arm overwrote the `ras_push`/`ras_pop` that the upper slot's call or return had set:

*   an upper-slot call never pushes its return address, so the eventual `ret` mispredicts.
*   an upper-slot return never pops, so the stale entry corrupts the next return.

`predict_address` is unaffected, so the redirect this cycle is correct. Only the RAS state update is lost.

### Fix

Move the two clears into the taken branch of the arm, so only a taken branch (which redirects and squashes the upper slot) clears the RAS ops. A not-taken branch keeps the upper slot's `ras_push`/`ras_pop`. The jump, jalr and return arms keep their unconditional clears, since those instructions always transfer control and their successor is off-path.

```verilog
if (taken_rvi_cf[i] || taken_rvc_cf[i]) begin
  cf_type[i] = ariane_pkg::Branch;
  ras_pop  = 1'b0;
  ras_push = 1'b0;
end
```

### Related Issue

#3545 

### Verification

Checked with JasperGold formal on config cv32a6_imac_sv32. With the fix both are proven unreachable, no counterexamples.

image

image

Prediction-quality only, no functional change: it changes when the RAS is written, not the fetched instruction stream.

### DIFF
--- a/.gitlab-ci/scripts/report_benchmark.py
+++ b/.gitlab-ci/scripts/report_benchmark.py
@@ -23,7 +23,7 @@ valid_cycles = {
     "dhrystone_single": 23866,
     "coremark_dual": 199928,
     "coremark_single": 290690,
-    "dhrystone_cv32a65x": 27329,
+    "dhrystone_cv32a65x": 27328,
     "dhrystone_cv32a60x": 36629,
 }
 
--- a/core/frontend/frontend.sv
+++ b/core/frontend/frontend.sv
@@ -279,7 +279,7 @@ module frontend
             cf_type[i] = ariane_pkg::Branch;
             // clear the RAS ops only for a taken branch. a not-taken branch falls
             // through, so it must keep the upper slot's ras_push/ras_pop
-            ras_pop  = 1'b0;
+            ras_pop = 1'b0;
             ras_push = 1'b0;
           end
         end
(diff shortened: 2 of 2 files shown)


JeanRochCoulon:
Great  @Misbahud-Din 
Curious to know the benefit on Coremark !  I will launch the Thales CI to get it

Misbahud-Din:
> Great @Misbahud-Din Curious to know the benefit on Coremark ! I will launch the Thales CI to get it

Thanks 🙂 I'd love to see the CoreMark numbers too, and cite them in my thesis if that's ok.

JeanRochCoulon:
Unfortunately I cannot see improvement on Coremark... Something is not working as expected. If you would like to run the coremark, the command is present in .gitlab-ci.yml

Misbahud-Din:
Thanks for checking. If it's safe to merge, please go ahead. I'll run the benchmarks after I finish the verification.

JeanRochCoulon:
If cva6 is not improved, no need to merge.

PR #3548, opened 2026-09-09
Fix duplicate cache store grant

### LINKED ISSUE #3502: [BUG] Double Grant for single request in cache_ctrl.sv
### Code of Conduct

### CVA6 commit affected

6cb2001

### Bug Description

I found a Bug in the cache controller, that produces two separate grant signals for the same write transaction. It was found by testing a new formal verification methodology. This bug can possibly lead to a deadlock if a read can occur between the two grants. This happens if the cache implementation is used in a processor that only holds the request-signals until a grant asserted. Then a possible scenario would be:

1. read request
2. read completes; same cycle write starts with immediate grant
3. processor start second read, because write was already granted
4. second grant of the write is asserted; processor can drop read signals and waits for read data; meanwhile cache waits for new request;

# Steps to reproduce
The counterexample requires four conditions. 
1. The cache controller of the selected port must complete a refill caused by a missed read request in the same cycle in which the new write request is issued. The preceding request has to be a read request, because this brings the controller into the subsequent `WAIT_CRITICAL_WORD` state.

2. The internal memory access must receive a grant in this state. It asserts the first grant for the write request and transitions into a state in which the write request is processed internally. In the following cycle, the cache accesses the metadata of the accepted write request.
 
3. This write request must miss in the cache, causing a refill to be initiated. This miss moves the controller into the `WAIT_REFILL_GNT` state. 
 
4. While the controller is in this state, either a bypass grant or a miss grant must be asserted. This directly causes the second grant for the same write request. Consequently, the write request is accepted twice although it should have been completed after the first grant.

## Observed behavior
Image
The counterexample shows a scenario in which an earlier read request completes while
a new write request is accepted in the same cycle. From the processor-side perspective,
the write request is completed by the grant. However, the cache continues processing
the same request internally and asserts a second grant two cycles later.

## Expected behavior
The grant signal in cycle t##-2 should be zero.

## Problem and possible solution
The underlying issue is that the grant behavior in the `WAIT_CRITICAL_WORD` state is inconsistent with the regular request acceptance path. In the other states in which a new request can be accepted like `IDLE`, a grant is only generated for a read request once the controller starts processing that request. Write requests are not completed through this early grant path. In the `WAIT_CRITICAL_WORD` state, however, the controller does not sufficiently distinguish between read and write requests and may assert a grant for both request types. A possible fix is therefore to align the grant behavior in the `WAIT_CRITICAL_WORD` state with the other request acceptance paths.

Current version of cache_ctrl.sv line 422-427 looks like this:
```verilog
// Wait until we have access on the memory array
if (gnt_i) begin
      state_d = WAIT_TAG;
      mem_req_d.bypass = 1'b0;
      req_port_o.data_gnt = 1'b1;
end
```

Proposed fix:
This is taken from the `IDLE` state, because it uses the condition '!req_port_i.data_we'. (line 171-175)
```verilog
if (gnt_i) begin
      state_d = WAIT_TAG;
      mem_req_d.bypass = 1'b0;
      // only for a read
      if (!req_port_i.data_we) req_port_o.data_gnt = 1'b1;
end
```

Feel free to contact me if there are questions regarding the issues. If you like I can also open a PR.

124107157-KV:

## Why this PR is needed

* `cache_ctrl` can generate two grants for the same write request.
* The problematic path occurs in the `WAIT_CRITICAL_WORD` state.
* A store can incorrectly receive an early `data_gnt`.
* The same store can later receive its normal grant through the miss/refill path.
* This can desynchronize the requester and cache and potentially cause a deadlock.

## What changed

* `WAIT_CRITICAL_WORD` now behaves consistently with the normal `IDLE` request-acceptance path.
* Early `data_gnt` is generated only when `!req_port_i.data_we`.
* Stores therefore wait for their normal final acceptance/grant.
* No changes were made to the legitimate later grant paths.

## Related issue

Fixes #3502

## Validation

* `verible-verilog-format`: passed
* `git diff --check` / `git show --check`: passed
* `make verilate target=cv64a6_imafdc_sv39`: passed
* `make verilate target=cv64a6_imafdc_sv39_wb`: passed
* WB configuration successfully elaborated `cache_ctrl.sv`

## Limitations

* The dedicated `tb_wb_dcache` testbench was not run locally.
* Questa/ModelSim tools `vlib`, `vlog`, and `vsim` are not installed in the local environment.

### DIFF
--- a/core/cache_subsystem/cache_ctrl.sv
+++ b/core/cache_subsystem/cache_ctrl.sv
@@ -423,7 +423,8 @@ module cache_ctrl
             if (gnt_i) begin
               state_d = WAIT_TAG;
               mem_req_d.bypass = 1'b0;
-              req_port_o.data_gnt = 1'b1;
+              // only for a read
+              if (!req_port_i.data_we) req_port_o.data_gnt = 1'b1;
             end
           end else begin
             state_d = IDLE;


JeanRochCoulon:
@cfuguet  Can you approve this PR ?

cfuguet:
Yes, I'll take a look into this

cfuguet (approved):
This looks good to me

cfuguet:
The only thing to note is that both legacy WT and WB cache subsystems will be deprecated in the following months... Only the HPDcache subsystem (which supports both write-through and write-back) will be officially supported.

The target of this PR is the legacy WB cache.

JeanRochCoulon:
In that case, I would prefer to not merge this PR. @124107157-KV, as César told, the WB cache is about to be removed from CVA6 repository. That is why I will not merge this PR. I will close it.
Thanks anyway for this PR and work.

PR #3550, opened 2026-09-10
Add valid bit to branch prediction structure

### LINKED ISSUE #3023: [BUG] Returns to address zero can be ignored

### Bug Description

Returns don't trigger `bp_valid` in the frontend when the RAS is empty, so fetch doesn't get redirected:

https://github.com/openhwgroup/cva6/blob/301d11ceb88c1169f75e9dea415e4bff4eb29888/core/frontend/frontend.sv#L306

However, returns mark the `cf_type` as `Return`, even if the RAS entry isn't valid:

https://github.com/openhwgroup/cva6/blob/301d11ceb88c1169f75e9dea415e4bff4eb29888/core/frontend/frontend.sv#L260-L264

 The RAS returns `predict_address` zero when it's empty.

The mispredict check in the branch unit checks if the predicted address differs from the target address:

https://github.com/openhwgroup/cva6/blob/301d11ceb88c1169f75e9dea415e4bff4eb29888/core/branch_unit.sv#L101-L103

The problem is that if the architecturally correct target address actually *is* zero then the instruction just retires as basically a nop.

Found by fuzzing against the Sail model using TestRIG: I didn't reproduce it on a more up-to date CVA6 using a more traditional test, but looking through the code the same problem seems to apply. It's obviously a fairly esoteric bug, but I think jumping to address zero should be legal. If people think it's important, I can try to find time to create a reproducer.

Three ideas for fixes:
- Add `|| branch_predict_i.predict_address == '0` to the mispredict check. I've confirmed this fixes the bug, but introduces some extra gates. It would also be very bad for performance if you were actually jumping to address zero repeatedly for some reason.
- Set the LSB of the RAS predict address if it's invalid: since target addresses will never have LSBs set, this guarantees the mispredict equality check will fail. I've confirmed this fixes the bug. We could actually put the valid bit in the RAS in the LSB of the returned address (and swap the sense) to save a bit of storage per entry.
- Only set `cf_type == Return` if the prediction is valid. I haven't tested this one. That means that the instruction will get detected as a mispredict.

teobiton:

I’ve added a `valid` bit to the prediction structure that is used throughout the pipeline.

This offers the following benefits:

-  Decoupling the control-flow instruction type from the validity of the prediction (code simplification)
-  Simplified handling of prediction validity (area savings; on an FPGA and in 32-bit mode, this represents 200 LUTs, or a 1% gain on the CV32A6)
-  Fixes #3023

Note that I have not verified the third point, but with this change, the bug should no longer occur.

I tested this change a few weeks ago, previous commit was [6cb2001](https://github.com/openhwgroup/cva6/tree/6cb200105fb9441d170e45786125a737fab98e91). Using the cv32a65x configuration, this modification improves the CoreMark score from 4.22 to 4.23 (only 14 cycles in reality) in simulation.

### DIFF
--- a/core/branch_unit.sv
+++ b/core/branch_unit.sv
@@ -100,7 +100,7 @@ module branch_unit #(
       end
       if (fu_data_i.operation == ariane_pkg::JALR
           // check if the address of the jump register is correct and that we actually predicted
-          && (branch_predict_i.cf == ariane_pkg::NoCF || target_address != branch_predict_i.predict_address)) begin
+          && (!branch_predict_i.valid || target_address != branch_predict_i.predict_address)) begin
         resolved_branch_o.is_mispredict = 1'b1;
         // update BTB only if this wasn't a return
         if (branch_predict_i.cf != ariane_pkg::Return)
--- a/core/cva6.sv
+++ b/core/cva6.sv
@@ -35,6 +35,7 @@ module cva6
     // this is the struct which we will inject into the pipeline to guide the various
     // units towards the correct branch decision and resolve
     localparam type branchpredict_sbe_t = struct packed {
+      logic                    valid;            // valid control flow prediction
       cf_t                     cf;               // type of control flow prediction
       logic [CVA6Cfg.VLEN-1:0] predict_address;  // target address at which to jump, or not
     },
--- a/core/frontend/frontend.sv
+++ b/core/frontend/frontend.sv
@@ -152,6 +152,7 @@ module frontend
   // Instruction FIFO
   logic [           CVA6Cfg.VLEN-1:0] predict_address;
   cf_t  [CVA6Cfg.INSTR_PER_FETCH-1:0] cf_type;
+  logic [CVA6Cfg.INSTR_PER_FETCH-1:0] cf_valid;
   logic [CVA6Cfg.INSTR_PER_FETCH-1:0] taken_rvi_cf;
   logic [CVA6Cfg.INSTR_PER_FETCH-1:0] taken_rvc_cf;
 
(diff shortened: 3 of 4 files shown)


JeanRochCoulon:
Hi @teobiton    Thanks for your PR.
This PR is very interesting if performance is improved. On CI side, it is not the case on Dhrystone and Coremark (your PR has no impact on cycle results).

Can't you discard mispredicted return addresses, this would really improve the performance. One way could be to push when JAL/JALR (speculative or not), pop when JAL/JALR mispredict or when RET.

JeanRochCoulon:
I do see where the code is simplified, but I do not see where you ignore the return address when it is zero

teobiton:
A return address equal to zero was always considered correctly predicted in the branch unit, because it is the default value of addresses stored in the RAS and the validity of the prediction was not propagated until the execution stage. Verifying that the RAS prediction is valid, no matter the address value, eliminates this corner case.
Doing this also avoids the repetitive checks `cf_type` , hence a reduced area.

> Can't you discard mispredicted return addresses, this would really improve the performance. One way could be to push when JAL/JALR (speculative or not), pop when JAL/JALR mispredict or when RET.

I see where you want to go with this, if I understand correctly you want some kind of RAS repairing based on speculative paths or function calls that end up being mispredicted, and IMO this requires much more changes to the frontend and to RAS management (and a separate PR).

JeanRochCoulon:
Ok, I will merge. 
NB: I saw a reduction of 400 gates synthetizing CVA6 on ASIC

JeanRochCoulon:
Merci !

PR #3552, opened 2026-09-10
Fix hgatp unsupported mode writes

### LINKED ISSUE #3495: [BUG] csr_regfile: a write to hgatp with an unsupported MODE is dropped entirely, so PPN and VMID keep the previous guest's values
### Code of Conduct

### CVA6 commit affected

6cb2001

## Bug description

`core/csr_regfile.sv:1529-1532` applies the satp idiom to hgatp verbatim:

```systemverilog
// only update if we actually support this mode
if (config_pkg::vm_mode_t'(hgatp.mode) == config_pkg::ModeOff ||
              config_pkg::vm_mode_t'(hgatp.mode) == CVA6Cfg.MODE_SV)
  hgatp_d = hgatp;
```

When MODE is neither Bare nor the supported Sv39x4 the assignment is skipped, so PPN and VMID
keep their previous values. That is the satp rule (supervisor.adoc: "if satp is written with
an unsupported MODE, the entire write has no effect; no fields in satp are modified"), and
hypervisor.adoc says the opposite for hgatp: "A write to hgatp with an unsupported MODE value
is not ignored as it is for satp. Instead, the fields of hgatp are WARL in the normal way, when
so indicated." Spike keeps the current MODE and takes the new PPN and VMID.

Effect: a hypervisor that writes hgatp with a MODE this core does not implement (Sv48x4, or a
reserved value) silently keeps the previous G-stage root and VMID, so the next guest runs on
the previous guest's second-stage translation. The same line is reached whatever the
privilege mode; the reproducer uses M-mode so only the CSR value is compared.

## Steps to reproduce

    git apply judged_hgatp-mode-warl_reproducer.patch
    make verilate target=cv64a6_imafdch_sv39
    cd hunt_repro/hgatp-mode-warl
    ./run_cosim.sh p_hgatp_mode.S        # experiment
    ./run_cosim.sh p_hgatp_mode_ctrl.S   # negative control

Each run compiles the program, runs it on `work-ver/Variane_testharness` and on
`tools/spike/bin/spike`, and diffs the commit traces: exit 0 = identical (`clean`), 1 =
divergence (`BUG_OBSERVED`), 2 = setup error. The program writes hgatp = {MODE=Sv39x4,
PPN=0x100}, reads it back, then writes {MODE=1 (reserved on RV64), PPN=0x200} and reads it
back. The control is the same program with MODE=Sv39x4 in the second write.

[judged_hgatp-mode-warl_fix.patch](https://github.com/user-attachments/files/31347520/judged_hgatp-mode-warl_fix.patch)
[judged_hgatp-mode-warl_reproducer.patch](https://github.com/user-attachments/files/31347521/judged_hgatp-mode-warl_reproducer.patch)

## Expected behavior

The second read-back is `0x8000000000000200`: MODE kept at Sv39x4, PPN taken. Both print
`clean`; the control already does.

## Observed behavior

Buggy design:

    BUG_OBSERVED: RTL and ISS diverge (p_hgatp_mode)
      RTL     59  pc=0x0000000080002028 insn=0x680029f3 rd=x19=0x8000000000000100
      ISS     59  pc=0x0000000080002028 insn=0x680029f3 rd=x19=0x8000000000000200
    clean: RTL and ISS retired identical instruction streams (p_hgatp_mode_ctrl)

The RTL still holds the first write; the whole second write was dropped.

Fixed design:

    clean: RTL and ISS retired identical instruction streams (p_hgatp_mode)
    clean: RTL and ISS retired identical instruction streams (p_hgatp_mode_ctrl)

## Suggested fix

`judged_hgatp-mode-warl_fix.patch`: keep the current MODE when the written one is
unsupported, and always take the rest of the write (comment elided):

```systemverilog
-  if (mode == ModeOff || mode == CVA6Cfg.MODE_SV)
-    hgatp_d = hgatp;
+  if (mode != ModeOff && mode != CVA6Cfg.MODE_SV)
+    hgatp.mode = hgatp_q.mode;
+  hgatp_d = hgatp;
```

This is the legalisation Spike applies. The satp write path keeps its own rule. Verified
only against the attached reproducer.

124107157-KV:

## Description

Fix the WARL handling of `hgatp.MODE` when software writes an unsupported MODE value.

The current `hgatp` write path only commits `hgatp_d` when the written MODE is either `Bare` (`ModeOff`) or the virtual-memory mode supported by the selected CVA6 configuration.

When software writes an unsupported MODE value, the assignment to `hgatp_d` is skipped completely. As a result, the entire write is discarded and the previous PPN and VMID values remain unchanged together with the previous MODE.

This behavior is incorrect for `hgatp`.

Unlike `satp`, an `hgatp` write containing an unsupported MODE value must not cause the entire CSR write to be ignored. The MODE field must be legalized according to its WARL behavior while the remaining writable fields are still allowed to take their legal values.

The fix keeps the currently valid `hgatp.MODE` when the written MODE is unsupported and then commits the rest of the legalized `hgatp` value. This allows the newly written PPN and VMID fields to update instead of retaining values from the previous guest.

The existing `satp` and `vsatp` behavior is not changed.

## Related issue

Fixes #3495

## Root cause

The existing `CSR_HGATP` write handling contains logic equivalent to:

    if (hgatp.mode == ModeOff || hgatp.mode == CVA6Cfg.MODE_SV)
      hgatp_d = hgatp;

This means that when the incoming MODE is unsupported, `hgatp_d` is never updated.

For example:

Initial legal value:

    MODE = Sv39x4
    PPN  = 0x100

Second write:

    MODE = 1
    PPN  = 0x200

MODE value 1 is unsupported for this configuration.

Before this fix, the second write is discarded entirely and the CSR remains equivalent to:

    MODE = Sv39x4
    PPN  = 0x100

The expected WARL behavior is:

    MODE = Sv39x4
    PPN  = 0x200

The legal MODE is retained while the legal PPN field from the new write is accepted.

## Changes

The `CSR_HGATP` write path in `core/csr_regfile.sv` is changed so that:

- supported MODE values continue to be written normally;
- an unsupported incoming MODE is replaced with the current legal `hgatp_q.mode`;
- the resulting legalized `hgatp` value is always committed to `hgatp_d`;
- existing PPN low-bit hardwiring remains unchanged;
- existing VMID masking remains unchanged;
- `satp` and `vsatp` handling remain unchanged.

A directed regression test is also added to reproduce the bug and verify the corrected WARL behavior.

## Regression test

Added:

    verif/tests/custom/hgatp_warl/hgatp_mode_warl.S

Testlist:

    verif/tests/testlist_hgatp_warl.yaml

Test name:

    rv64h-p-hgatp-mode-warl

The regression performs the following sequence:

1. Write a legal `hgatp` value with MODE set to Sv39x4 and PPN set to `0x100`.

2. Read back `hgatp` and verify that the legal write was accepted.

3. Write `hgatp` again with unsupported MODE value `1` and PPN set to `0x200`.

4. Read back `hgatp`.

5. Verify that MODE remains Sv39x4 while PPN changes to `0x200`.

The second check specifically detects the original bug where the complete second write was discarded.

## Verification

Target configuration:

    cv64a6_imafdch_sv39

The Verilator model was built with:

    make verilate target=cv64a6_imafdch_sv39

The post-change Verilator build completed successfully with exit status 0.

### Before the fix

The directed regression reproduced the bug:

    hgatp_mode_warl.o *** FAILED *** (tohost = 2)

The simulation terminated after approximately 2928 cycles.

The failure occurs at test number 2, which is the unsupported-MODE write/read-back check.

This confirms that the original RTL retains the previous PPN after the unsupported MODE write instead of accepting the new PPN.

### After the fix

After rebuilding the Verilator model with the RTL change, the exact same regression passed:

    hgatp_mode_warl.o *** SUCCESS *** (tohost = 0)

The post-fix RTL run returned:

    RTL_RC=0

This demonstrates the expected transition:

    Before fix: FAILED, tohost = 2
    After fix:  SUCCESS, tohost = 0

The directed test was also compiled and run against Spike as the reference model to confirm the expected WARL behavior.

## Files changed

    core/csr_regfile.sv
    verif/tests/custom/hgatp_warl/hgatp_mode_warl.S
    verif/tests/testlist_hgatp_warl.yaml

## Scope

This change is intentionally limited to the `hgatp` unsupported-MODE WARL handling.

It does not modify the behavior of:

- `satp`;
- `vsatp`;
- unrelated CSR handling;
- MMU translation logic;
- page-table walking logic.

## Limitations

Local directed RTL verification was performed with the `cv64a6_imafdch_sv39` configuration.

Broader configuration and regression coverage is expected to be exercised by the CVA6 continuous-integration flow.

### DIFF
--- a/core/csr_regfile.sv
+++ b/core/csr_regfile.sv
@@ -1579,10 +1579,11 @@ module csr_regfile
               hgatp[1:0] = 2'b0;
               // only make VMID_LEN - 1 bit stick, that way software can figure out how many VMID bits are supported
               hgatp.vmid = hgatp.vmid & {{(CVA6Cfg.VMIDW - CVA6Cfg.VMID_WIDTH) {1'b0}}, {CVA6Cfg.VMID_WIDTH{1'b1}}};
-              // only update if we actually support this mode
-              if (config_pkg::vm_mode_t'(hgatp.mode) == config_pkg::ModeOff ||
-                            config_pkg::vm_mode_t'(hgatp.mode) == CVA6Cfg.MODE_SV)
-                hgatp_d = hgatp;
+              // Preserve the current mode if the written mode is unsupported.
+              if (config_pkg::vm_mode_t'(hgatp.mode) != config_pkg::ModeOff &&
+                  config_pkg::vm_mode_t'(hgatp.mode) != CVA6Cfg.MODE_SV)
+                hgatp.mode = hgatp_q.mode;
+              hgatp_d = hgatp;
             end
             // changing the mode can have side-effects on address translation (e.g.: other instructions), re-fetch
             // the next instruction by executing a flush
--- a/verif/tests/custom/hgatp_warl/hgatp_mode_warl.S
+++ b/verif/tests/custom/hgatp_warl/hgatp_mode_warl.S
@@ -1,0 +1,55 @@
+#include "riscv_test.h"
+#include "test_macros.h"
+
+.option arch, +h
+
+#define CSR_HGATP               0x680
+#define HGATP_MODE_SV39X4       8
+#define HGATP_MODE_UNSUPPORTED  1
+
+RVTEST_RV64M
+RVTEST_CODE_BEGIN
+
+  li TESTNUM, 1
+
+  # Install an initial legal value:
+  # MODE = Sv39x4
+  # PPN  = 0x100
+  li   t0, HGATP_MODE_SV39X4
+  slli t0, t0, 60
+  ori  t0, t0, 0x100
+
+  csrw CSR_HGATP, t0
+  csrr t1, CSR_HGATP
+
+  bne  t1, t0, fail
+
+  li TESTNUM, 2
+
+  # Write an unsupported MODE with a different PPN.
+  # MODE must be legalized, but PPN must still update.
+  li   t2, HGATP_MODE_UNSUPPORTED
+  slli t2, t2, 60
+  ori  t2, t2, 0x200
+
+  csrw CSR_HGATP, t2
+  csrr t3, CSR_HGATP
+
+  # Expected:
+  # MODE remains Sv39x4
+  # PPN changes to 0x200
+  li   t4, HGATP_MODE_SV39X4
+  slli t4, t4, 60
+  ori  t4, t4, 0x200
+
+  bne  t3, t4, fail
+
+  j pass
+
+  TEST_PASSFAIL
+
+RVTEST_CODE_END
+
+RVTEST_DATA_BEGIN
+  TEST_DATA
+RVTEST_DATA_END
--- a/verif/tests/testlist_hgatp_warl.yaml
+++ b/verif/tests/testlist_hgatp_warl.yaml
@@ -1,0 +1,8 @@
+# Regression for unsupported hgatp.MODE WARL handling.
+
+testlist:
+  - test: rv64h-p-hgatp-mode-warl
+    iterations: 1
+    path_var: TESTS_PATH
+    gcc_opts: "-static -mcmodel=medany -fvisibility=hidden -nostdlib -nostartfiles -I<path_var>/riscv-tests/isa/macros/scalar/ -I<path_var>/riscv-tests/env/p/ -I<path_var>/riscv-tests/riscv-target/spike/"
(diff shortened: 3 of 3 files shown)


JeanRochCoulon:
@ASintzoff can you approve ?

PR #3562, opened 2026-09-15
decoder: reject reserved same-format FP conversions

### LINKED ISSUE #3494: [BUG] decoder: reserved same-format FP conversion encodings are accepted (decode gap; breaks ISS co-simulation)
### Code of Conduct

### CVA6 commit affected

6cb2001

## Bug description

`core/decoder.sv:1503-1519` decodes the OP-FP float-to-float conversion case
(`funct5 = 5'b01000`) by checking each format field on its own: the source through
`unique case (instr.rftype.rs2[22:20])`, whose arms only ask whether that format's
extension is compiled in (`3'b000: if (~CVA6Cfg.RVF) illegal_instr = 1'b1;`), and the
destination separately at `decoder.sv:1565-1572` (`unique case (instr.rftype.fmt)`).
Neither check ever compares the two fields.

Only conversions between *different* formats exist. d-st-ext.adoc: "the
double-precision to single-precision and single-precision to double-precision
conversion instructions, FCVT.S.D and FCVT.D.S ... The _rs2_ field encodes the datatype
of the source, and the _fmt_ field encodes the datatype of the destination." With F and
D that is the whole defined set, so both same-format code points are reserved:
`fmt=00, rs2=00000` ("fcvt.s.s") and `fmt=01, rs2=00001` ("fcvt.d.d"). Both retire here as
ordinary FCVT_F2F; the ISS raises illegal-instruction on both. Every other undefined point
of this case is rejected - non-zero `rs2[24:23]`, a format whose extension is off, a bad
rounding mode. Only source-equals-destination survives: that legality is a relation
between two fields, not a property of either.

rv32.adoc:156 makes decoding a reserved instruction UNSPECIFIED, so this is a decode gap,
not a wrong-result bug: it mismatches Spike co-simulation, consumes two encoding points
a future extension may claim, and the two cases disagree (canonical NaN vs bit copy).

## Steps to reproduce

`git apply judged_fcvt-same-format_reproducer.patch`, build the model
(`make verilate target=cv64a6_imafdch_sv39`), then:

    cd hunt_repro/fcvt-same-format
    ./run_cosim.sh p_fcvt_same_fmt.S     # experiment
    ./run_cosim.sh p_fcvt_control.S      # negative control

Each run compiles the program, runs it on `work-ver/Variane_testharness` and on
`tools/spike/bin/spike`, and diffs the commit traces: exit 0 = identical (`clean`), 1 =
divergence (`BUG_OBSERVED`), 2 = setup error. The control is the same program with its
two reserved words replaced by defined conversions.

[judged_fcvt-same-format_fix.patch](https://github.com/user-attachments/files/31347425/judged_fcvt-same-format_fix.patch)
[judged_fcvt-same-format_reproducer.patch](https://github.com/user-attachments/files/31347426/judged_fcvt-same-format_reproducer.patch)

## Expected behavior

Both print `clean`. The ISA leaves reserved decoding UNSPECIFIED, so no trap is mandated; 
the expectation is consistency - with the decoder's own handling of every neighbouring undefined 
point in this case (all rejected), and with the ISS, which traps on both words.

## Observed behavior

Buggy design:

    BUG_OBSERVED: RTL and ISS diverge (p_fcvt_same_fmt)
      RTL     60  pc=0x000000008000202c insn=0x40000153 rd=f2=0xffffffff7fc00000
      ISS     60  pc=0x0000000080002040 insn=0x34102373 rd=x6=0x000000008000202c
    clean: RTL and ISS retired identical instruction streams (p_fcvt_control)

At `0x8000202c` the RTL retires `0x40000153` and writes `f2` while the ISS traps into the
handler (`csrr t1, mepc` reads back `0x8000202c`). `0x421001d3` also retires, `f3=1.0`.

Fixed design:

    clean: RTL and ISS retired identical instruction streams (p_fcvt_same_fmt)
    clean: RTL and ISS retired identical instruction streams (p_fcvt_control)

## Suggested fix

`judged_fcvt-same-format_fix.patch`, appended to the source-format check (comment elided):

    +  if (!CVA6Cfg.XF16ALT && instr.rftype.rs2[22:20] == {1'b0, instr.rftype.fmt})
    +    illegal_instr = 1'b1;

It mirrors the per-field rejections already in that case. Configs with Xf16alt encode the
alt-half destination as fmt=10 with rm=101 and need the same check written for that encoding;
they are deliberately left unchanged here. Verified only against the attached reproducer.

## Relationship to known issues
Same class as #3463 (HLV/HLVX/HSV and CBO fixed-zero fields) and #3306 (zip/unzip on RV64): a reserved encoding accepted by the decoder; different instructions and different decoder lines, and neither fix touches the FCVT case.

124107157-KV:

## Why is this PR needed?

The scalar `FCVT_F2F` decode path validates the source and destination
floating-point formats independently, but it does not reject conversions
where the source and destination formats are the same.

As a result, reserved same-format encodings such as the equivalent of
`fcvt.s.s` and `fcvt.d.d` are accepted by CVA6 instead of being rejected
by the decoder.

Spike treats these encodings as illegal instructions, so executing them
causes CVA6 and Spike to diverge during ISS co-simulation.

## What does this PR change?

Add an additional legality check in the scalar `FCVT_F2F` decoder path
that rejects an instruction when its source floating-point format matches
its destination format.

A directed regression test, `fcvt-same-format-rv64`, is also added. It
checks that:

- legal `fcvt.s.d` remains accepted;
- legal `fcvt.d.s` remains accepted;
- reserved same-format S-to-S encoding raises an illegal-instruction
  exception;
- reserved same-format D-to-D encoding raises an illegal-instruction
  exception;
- reserved FP16-to-FP16 encoding raises an illegal-instruction exception;
- reserved FP16ALT-to-FP16ALT encoding raises an illegal-instruction exception;
- FP16-to-FP16ALT and FP16ALT-to-FP16 cross-format conversions remain accepted.

The test also verifies the exception cause and the address of each
faulting instruction.

## Verification

Tested with:

`cv64a6_imafdc_sv39`

Before the RTL fix:

- `fcvt-same-format-rv64`
  - CVA6: failed with `tohost = 1`
  - Spike: raised illegal-instruction exceptions for both reserved
    same-format encodings

After the RTL fix:

- `fcvt-same-format-rv64`
  - PASS
  - 136 instructions matched between CVA6 and Spike

Existing floating-point conversion regressions:

- `rv64uf-p-fcvt`
  - PASS
  - 76 instructions matched

- `rv64ud-p-fcvt`
  - PASS
  - 104 instructions matched

XF16/XF16ALT-specific verification:

- FP16 -> FP16ALT: accepted
- FP16ALT -> FP16: accepted
- FP16 -> FP16: illegal instruction
- FP16ALT -> FP16ALT: illegal instruction
- test completed with `tohost = 0`

Final local regression report:

- 6 PASSED
- 0 FAILED

Additional checks:

- `verible-verilog-format --inplace core/decoder.sv`
- `git diff --check`

Fixes #3494


IhsaneTahir:
Hi @124107157-KV,

I understand that XF16ALT has a special encoding that differs from the other formats. But why isn't there a second arm that raises an illegal_instr when a same-format conversion in XF16ALT happens?

124107157-KV:
> Hi @124107157-KV,
> 
> I understand that XF16ALT has a special encoding that differs from the other formats. But why isn't there a second arm that raises an illegal_instr when a same-format conversion in XF16ALT happens?

Good point. I followed the issue's suggested patch, which intentionally left XF16ALT unchanged. Since XF16ALT uses `fmt=10/rm=101` for the destination and `rs2[22:20]=110` for the source, I'll add the equivalent ALT-to-ALT rejection and a regression while keeping FP16↔FP16ALT conversions legal.

core/decoder.sv
```diff
@@ -1531,6 +1531,11 @@ module decoder
                   3'b011:  if (~CVA6Cfg.XF8) illegal_instr = 1'b1;
                   default: illegal_instr = 1'b1;
                 endcase
+                if (CVA6Cfg.XF16ALT && instr.rftype.fmt == 2'b10 && instr.rftype.rm == 3'b101) begin
+                  if (instr.rftype.rs2[22:20] == 3'b110) illegal_instr = 1'b1;
+                end else if (instr.rftype.rs2[22:20] == {1'b0, instr.rftype.fmt}) begin
```
IhsaneTahir: It's redundant to include bit 22 of rs2 field in this check and compare it against 1'b0. The default case in the unique case statement, which checks source format encodings, already covers it.

124107157-KV: Updated in the latest push. The regular comparison now uses rs2[21:20], while the full rs2[22:20] field is only used for the XF16ALT-specific handling. I also re-ran the directed and existing FCVT regressions successfully.

IhsaneTahir: Thank you, this looks good to me.

PR #3564, opened 2026-09-16
Fix RV64 Zcmp push/pop offsets

### LINKED ISSUE #3440: [BUG] RV64 Zcmp PUSH/POP expansion uses 4-byte offset increments, causing overlapping accesses and incorrect SP adjustment
### Code of Conduct

### CVA6 commit affected

88e810c

## Affected
- File: `core/macro_decoder.sv`
- Config: `RVZCMP = 1` and `IS_XLEN64 = 1`

## Summary
The RV64 expansion of `cm.push/cm.pop/cm.popret/cm.popretz` reuses the RV32 element size (4 bytes) instead of 8 bytes. The per-register offset still steps by −4 (`12'hFFC`), and the only RV64-specific adaptations are (a) forcing bit 2 of the SD/LD immediate to 0 and (b) subtracting an extra 4 on the final SP adjustment. As a result, RV64 `cm.*` produce wrong memory addresses and a wrong final stack pointer. RV32 is unaffected.

## Root cause
Generally `bytes = (XLEN==32) ? 4 : 8`, addresses step by `bytes`, and the frame is adjusted by exactly `stack_adj`.

In `macro_decoder.sv`:
1. The offset decrement is hardcoded to `12'hFFC` (−4) and shared by both XLENs, e.g. lines 462, 481, 503, 540, 552, 565, 663, 673, 690, 698. Correct for RV32 (word = 4B), wrong for RV64 (dword = 8B, should be −8).
2. The RV64 SD/LD encodings force immediate bit 2 to 0 (stores: the `offset_d[4:3], 1'b0, offset_d[1:0]` split; loads: `offset_d[11:3], 1'b0, offset_d[1:0]`, e.g. lines 515/534/546/558/671). This masks the −4 step to look 8-byte aligned instead of fixing the step.
3. The final SP adjustment subtracts an extra 4 only on RV64 (lines 606/609), whereas the RV32 branch uses the plain value (lines 627/630).

In short: the RV64 path was bolted onto RV32 logic without changing the element size from 4 to 8 bytes.

## Observable effects (RV64)
1. Stack pointer off by 4:
   - `cm.push`: `sp -= (stack_adj + 4)` instead of `sp -= stack_adj`
   - `cm.pop/cm.popret/cm.popretz`: `sp += (stack_adj - 4)` instead of `+= stack_adj`

2. Saved/restored registers do not occupy distinct 8-byte slots. Because the offset steps by 4 while accesses are 8-byte SD/LD (with bit 2 masked), consecutive registers collapse onto the same slot.

   Example `cm.push {ra, s0-s1}, -32` (RV64) expands to:
   ```
   sd s1, -8(sp)
   sd s0, -8(sp)     # overwrites s1
   sd ra,-16(sp)
   addi sp, sp, -36  # should be -32
   ```
 correct expansion:
   ```
   sd s1, -8(sp)
   sd s0,-16(sp)
   sd ra,-24(sp)
   addi sp, sp, -32
   ```

124107157-KV:

## Why is this PR needed?

The RV64 Zcmp PUSH/POP macro expansion currently uses the RV32 register-slot size when calculating stack offsets.

On RV32, registers occupy 4-byte stack slots, so decrementing the generated offset by 4 bytes is correct. On RV64, however, the generated `SD` and `LD` operations operate on 8-byte values and therefore the offset must advance in 8-byte steps.

The existing implementation continues to decrement offsets by 4 bytes for both XLEN values and compensates for RV64 by forcing immediate bit 2 of generated `SD`/`LD` instructions to zero. This causes consecutive RV64 register accesses to alias the same stack location.

For example:

    cm.push {ra, s0-s1}, -32

should expand on RV64 as:

    sd   s1, -8(sp)
    sd   s0, -16(sp)
    sd   ra, -24(sp)
    addi sp, sp, -32

Before this change, the expansion effectively produces overlapping stack accesses:

    sd   s1, -8(sp)
    sd   s0, -8(sp)
    sd   ra, -16(sp)
    addi sp, sp, -36

The same offset problem affects `cm.pop`, `cm.popret`, and `cm.popretz`.

The RV64 path also applies an additional 4-byte correction to the final stack-pointer update. As a result, PUSH adjusts SP by four bytes too much and POP-family instructions adjust SP by four bytes too little.

## What does this PR change?

This PR makes the Zcmp register-slot offset dependent on XLEN:

- RV32 uses a 4-byte register-slot step.
- RV64 uses an 8-byte register-slot step.

The new XLEN-dependent offset is used throughout the PUSH/POP macro expansion state machine.

The change also:

- fixes the initial POP-family register offset on RV64;
- uses the correct offset progression for all subsequent expanded register accesses;
- removes the RV64-specific immediate bit-2 masking from generated `SD` and `LD` instructions;
- removes the extra 4-byte correction from the final RV64 stack-pointer adjustment;
- keeps the existing RV32 behavior unchanged;
- fixes the special maximum-rlist path handled through `PUSH_POP_INSTR_2`.

## Verification

A focused RV64 macro-decoder regression was added.

Before applying the RTL fix, the regression reproduced the issue with the following incorrect behavior:

- `cm.push {ra,s0-s1}, -32`
  - `s1`: `-8(sp)`
  - `s0`: `-8(sp)` instead of `-16(sp)`
  - `ra`: `-16(sp)` instead of `-24(sp)`
  - SP adjustment: `-36` instead of `-32`

- `cm.pop {ra,s0-s1}, 32`
  - `s1`: `24(sp)`
  - `s0`: `24(sp)` instead of `16(sp)`
  - `ra`: `16(sp)` instead of `8(sp)`
  - SP adjustment: `28` instead of `32`

- the maximum-rlist PUSH and POP paths also repeated adjacent stack offsets instead of progressing by 8 bytes.

After the fix, the RV64 regression verifies:

- `cm.push`
  - offsets `-8`, `-16`, and `-24`
  - final SP adjustment `-32`

- `cm.pop`
  - offsets `24`, `16`, and `8`
  - final SP adjustment `32`

- `cm.popret`
  - correct RV64 load offsets
  - correct SP adjustment
  - correct final return micro-op

- `cm.popretz`
  - correct RV64 load offsets
  - correct zeroing of `a0`
  - correct SP adjustment
  - correct final return micro-op

- maximum-rlist PUSH/POP offset progression through `PUSH_POP_INSTR_2`.

The existing RV32 macro-decoder regression continues to pass.

The following checks were run successfully:

    make -C verif/tb/macro_decoder_unit
    make -C util/sanitize xlen
    make verilate target=cv64a6_imafdc_sv39

The full RV64 Verilator build completes successfully.

## Related issue

Fixes #3440

## Limitations

The new regression tests the macro decoder directly rather than executing the compressed instructions as a full architectural software test.


fpetrot:
Thanks for your efforts, nevertheless I believe more profound changes are needed to clarify the intent of this macro-decoder, and that it can be made while halving the number of lines (you can have a new look at PR#3546 that you were kind enough to review) : the stack adjustment can be computed as a single line instead of 95 !

Note that I have updated since the code of that PR along the lines you suggested, see  ([https://github.com/fpetrot/cva6/blob/tmp/master/core/macro_decoder.sv](https://github.com/fpetrot/cva6/blob/tmp/master/core/macro_decoder.sv)), but I did not resubmit it because I face the same issue as the current implementation does, which is that stack instructions fail as soon as there are quite a few other dependent instructions that enter the pipeline.
I believe there is a sync issue as new insns enter although fetch_stall_o is 0, but spent useless hours trying to fix that, so I hope someone more knowledgeable could fix it for 32-bit, and then I could move on to 64 (see issue  #3582).

124107157-KV:
> Thanks for your efforts, nevertheless I believe more profound changes are needed to clarify the intent of this macro-decoder, and that it can be made while halving the number of lines (you can have a new look at PR#3546 that you were kind enough to review) : the stack adjustment can be computed as a single line instead of 95 !
> 
> Note that I have updated since the code of that PR along the lines you suggested, see (https://github.com/fpetrot/cva6/blob/tmp/master/core/macro_decoder.sv), but I did not resubmit it because I face the same issue as the current implementation does, which is that stack instructions fail as soon as there are quite a few other dependent instructions that enter the pipeline. I believe there is a sync issue as new insns enter although fetch_stall_o is 0, but spent useless hours trying to fix that, so I hope someone more knowledgeable could fix it for 32-bit, and then I could move on to 64 (see issue #3582).

Thanks for the feedback. I simplified the stack-adjustment logic as suggested, replacing the large RV32/RV64 lookup with a common XLEN-based calculation.

I kept the existing FSM/state sequencing unchanged because the broader refactor is still affected by the synchronization issue tracked in #3582, so I wanted to keep this PR focused on the RV64 offset bug in #3440 without introducing unrelated control-flow changes.

After the simplification, the RV32 and RV64 macro-decoder unit tests pass, the XLEN sanitizer passes, and full Verilator builds for `cv32a6_imac_sv32` and `cv64a6_imafdc_sv39` both complete successfully.

fpetrot:
Thanks. I understand your approach, but I believe it would be safer to have a working 32 bit version, and then move it to 64. As you can see in my patch, we only need to compute funct3 and xlen_bytes combinatorially, and then there are no more dependencies on 32 vs 64 within the finite state machine as the emitted instructions only need these 2 quantities (my current implementation also passes the unit tests, but a processor including this extension would fail, so I want to raise a warning here).

Let me summarize my understanding:
- when issue_ack_i == 1 we can advance from state to state, otherwise we stay in the same state
- when we set fetch_stall_o = 1 (in a combinatorial manner, depending on issue_ack_i), we should not have changes in the instruction we currently process, and when set fetch_stall_o = 0 when going back to IDLE, the instr_i should change on the rising edge of the clock, so be new when we enter IDLE.
That's what I have been doing, but it does not seem to be the expected behavior ...

124107157-KV:
> Thanks. I understand your approach, but I believe it would be safer to have a working 32 bit version, and then move it to 64. As you can see in my patch, we only need to compute funct3 and xlen_bytes combinatorially, and then there are no more dependencies on 32 vs 64 within the finite state machine as the emitted instructions only need these 2 quantities (my current implementation also passes the unit tests, but a processor including this extension would fail, so I want to raise a warning here).
> 
> Let me summarize my understanding:
> 
> * when issue_ack_i == 1 we can advance from state to state, otherwise we stay in the same state
> * when we set fetch_stall_o = 1 (in a combinatorial manner, depending on issue_ack_i), we should not have changes in the instruction we currently process, and when set fetch_stall_o = 0 when going back to IDLE, the instr_i should change on the rising edge of the clock, so be new when we enter IDLE.
>   That's what I have been doing, but it does not seem to be the expected behavior ...

Thanks for the clarification. I updated the macro decoder so that the XLEN-dependent quantities are computed once using `XLEN_BYTES` and `MEM_FUNCT3`, and the FSM no longer contains separate RV32/RV64 instruction-generation paths. The existing FSM and `issue_ack_i` sequencing are otherwise unchanged.

I also checked the synchronization concern at processor level using the dependent `cm.push`/`cm.popret` sequence from #3582. Both RV32 and RV64 Zcmp-enabled processor tests complete successfully with `tohost = 0`. The macro-decoder unit tests, XLEN sanitizer, and RV32/RV64 Verilator builds also pass.

Thanks again for pointing this out.

fpetrot:
Thanks @124107157-KV. Sorry to be such a pain, but could you check that 
[https://github.com/fpetrot/cva6/blob/master/verif/tests/custom/Zcmp/cm_popretz_test.S](https://github.com/fpetrot/cva6/blob/master/verif/tests/custom/Zcmp/cm_popretz_test.S)
and
[https://github.com/fpetrot/cva6/blob/master/verif/tests/custom/Zcmp/cm_popret_test.S](https://github.com/fpetrot/cva6/blob/master/verif/tests/custom/Zcmp/cm_popret_test.S)
behave as expected by `spike`?

124107157-KV:
Thanks @fpetrot. I checked both `cm_popret_test.S` and `cm_popretz_test.S`.

Both RV32 and RV64 versions pass with Spike.

At CVA6 processor level, the tests still expose the Zcmp synchronization issue. I also ran the same RV32 binaries on the pre-PR base commit `81245a47f`, and they show the same non-terminating behavior there as on this PR head.

So this appears to be the separate synchronization problem tracked in #3582 rather than something introduced by the RV64 offset changes in #3564.

I restored the branch to the focused #3440 changes only. The macro-decoder regressions, RV64 offset tests, XLEN sanitizer, and `git diff --check` all pass.

